Issue #5, opened 2026-09-07
[QSL#5] [LC04] Coordinate executable IR expansion and qualified backend parity


kreneskyp:
## Required implementation language and containment

Owner directive (2026-09-07): first-party production and possible qualification paths must be **Rust**, including libraries, CLI/tools, generators, validators, canonicalization, result adapters, test assertions and fixture/CI audits. "Glue", prototypes, scripts and tests are not exemptions; a Rust wrapper around non-Rust semantic logic is not remediation.

- Every new TypeScript choice needs explicit owner approval for its exact scope. Any other executable language must be brought to the owner's attention; do not silently introduce or expand it.
- Quoin's existing implementation is explicitly retained for now: **contain its spread, do not schedule a wholesale Quoin rewrite**. Keep new reusable semantics in Rust behind versioned, structured interfaces. Existing Node/TypeScript is not blanket permission for new components or expanded language scope.
- Inventory existing JavaScript/Python/shell and executable generated/inline CI logic; record a Rust remediation owner or an explicit owner disposition. Data schemas and foreign-language fixture samples must be distinguished from executable checks. External tools/hosts/bindings need a scoped disposition, not an invented exemption.
- Reuse shared assurance contracts and Quoin evidence ownership; do not recreate local evidence frameworks. Classify each helper as an existing shared capability, a missing reusable capability, or domain-specific logic before porting.
- Complete /specify and /spec-review for the changed requirements/interfaces, resolve findings, then implement and verify. Preserve ongoing work and actual evidence provenance. Do not claim a language port or qualification complete from a policy update.
- Filament repositories/tickets are excluded from this remediation. Interface consumption does not authorize producer changes.

Before closure: record executable-path inventory, approved dispositions, Rust implementation/test/CI parity and failure-path evidence for the reviewed scope. Existing Quoin retention is the only accommodation recorded here; this block grants no new TypeScript approval.

---

## Required spec cycle — owner directive

**Every change requires the spec cycle before implementation: `/specify` → `/spec-review` → resolve findings and repeat as needed → implement and verify against the reviewed specification.** Architecture notes, ticket acceptance checklists, an ADR draft, passing code tests, or `quire validate` alone do not satisfy this gate.

[...]

PR #13, opened 2026-09-09
Derive native Boolean projections through the existing IR binder

kreneskyp:
Native packages previously retained only unlowered clauses. This adds `lowering::lower`, deriving complete Boolean executable projections through the existing IR constructors and strict binder. It preserves authored identities, source coordinates, direct model declarations and native/IR observation correspondence. Unsupported clauses and exhausted budgets return no projection; the original native artifact retains the full model authority.

The qualification compiles the actual pinned code generator's output. All eight Boolean assignments match native execution and independent truth equations; native implication activation counts also match their independent expectations. Both production IR `690bde7` and the backend's pinned IR `04eb6f8` accept the emitted bytes with matching bound identity.

**Delivery scope: proof-of-concept engineering.** Generated activation qualification remains open: Rust 1.98.1 produces LLVM export 3.1.0, while the pinned codegen coverage reader accepts 3.0.1. The required activation test and FR-009-AC-5 remain incomplete. This assurance work is deferred from engineering delivery under the owner's direction; no full backend qualification or production-readiness claim is made. C owns the [existing reader handoff](https://github.com/agent-ix/quire-research/issues/54#issuecomment-5605460660).

Validation and review:

- Rust implementation/tests at `d58ca7a`: 266 ordinary tests plus three compile-fail doctests passed; six targeted tests, formatting and strict Clippy passed. Three private fixture audits passed at the implementation baseline. The ignored activation lane fails explicitly when invoked.
- Independent SR-114/115 closes the hard-limit and diagnostic findings. Operator substitution and individual off-by-one limit mutations are detected. Two low maintenance findings remain in Plan-008.
- QUOIN specification reviews SR-116–123 complete the selected PR-time review. SR-124 records that full Plan-008 acceptance fails because Task-020 retains deferred assurance work; it does not block this engineering delivery. The advisor could not detect the installed Quire version, so verification-method choices are explicitly author judgement.
- Quire validation passes for all 244 selected documents. TM-006 has 3/3 test cases backed by trace attributes; complete activation is not inferred from trace presence. No production or test code changed after `d58ca7a`.

CI stays `workflow_dispatch` only; no hosted run was requested. First-party implementation is Rust under AGPL-3.0-only, with dependency and generated-source grants retained.

Refs #5. Full numeric/object lowering, compiled ConfigVersion, FS05 acceptance and Quire integration remain open engineering/acceptance work.


kreneskyp:
## Independent code review at `813069f` — CONDITIONAL

Artifact: `reviews/26-09-09-lc04-boolean-lowering-independent-review.md` (SR-114, pushed, quire-validated).

**The backend qualification is real.** I doubted it first — `tests/native_backend.rs` finishes in **1.28 s**, which looks far too fast for "compile generated Rust, run it 8 times, run `cargo llvm-cov report` 8 times". So I mutated the lowering:

| # | Mutant | Result |
|---|---|---|
| M1 | `BinaryOp::And` lowered as `ShortCircuitOr` | **FAIL** — `generated_truth_matches_all_assignments…` at `native_backend.rs:282` |
| M2 | `LoweringLimits::default()` → `usize::MAX` on all three axes | **all 266 pass** |

M1 settles it: the loop genuinely compiles, executes and compares the generated program. The runtime is honest — the generated crate is three files with no dependencies. Worth saying so, because the next reader will have the same doubt.

### Gates at `813069f` (Rust 1.98.1, `-j 1`, `--test-threads=1`)

`fmt` ✓ · `clippy --all-targets --all-features -D warnings` ✓ · `cargo test --locked` **266 passed / 1 ignored** ✓ · `cargo-llvm-cov 0.9.0` present ✓ · `docs/dependency-licenses.json` 140 entries vs `Cargo.lock` 140 packages ✓

### One medium

**FND-001 — the three hard ceilings are documented, clamped, and unbound.** FR-009 states them normatively ("10,000 visited native nodes, depth 64 and 16 MiB serialized output"), and `LoweringLimits` repeats them. Replacing all three with `usize::MAX` leaves **every one of the 266 tests green**.

`exact_limits_and_fresh_retries` is a good test — zero/one-below/exact on all three axes, fresh retry after each exhaustion, `usize::MAX` clamped. But every assertion is *relative* to `LoweringLimits::default()`; the `above` case asserts `lower(above).usage() == projection.usage()`, which holds for any default. Two lines close it:

```rust
let hard = LoweringLimits::default();
assert_eq!((hard.nodes, hard.depth, hard.bytes), (10_000, 64, 16_777_216));
```

Note `depth: 64` mirrors `checking.rs:68` and `linking.rs:44`, so a package deep enough to trip lowering's depth ceiling is refused two stages earlier — defence-in-depth, and worth one sentence in FR-009 so nobody later deletes it as dead. `nodes` and `bytes` are lowering-only and genuinely unguarded.

### Three lows

**FND-002** — the CI job gains `cargo install cargo-llvm-cov --locked` (a from-source tool build) plus the generated-crate build and 8 coverage reports, while `timeout-minutes: 10` is unchanged. Dispatch-only so nothing is gated on it, but the first person to dispatch will hit a timeout unrelated to their change.

**FND-003** — `docs/dependency-licenses.json` is referenced by no test, tool or CI step. It is accurate today (I counted: 140 vs 140, including both revision-qualified git sources) and this PR updates it correctly; nothing keeps it that way.

**FND-004** — `native_backend.rs:313` pins a downstream crate's verbatim prose diagnostic next to the durable `code == "unsupported_profile"` assertion. Pinning the refusal is right — "update the qualification when the backend changes" is a deliberate tripwire and the honest alternative to waiving the gate. Asserting the message *contains* `3.0.1` and `3.1.0` keeps the tripwire without breaking on an upstream rewording.

### Checked and found sound

So you don't re-derive these: **read deduplication cannot lose an observation** — `or_insert_with` would under-report only if two reads of one name in one clause carried different observations, and the only construct that produces one is `pre`, a builtin, which falls to `expression()`'s `_` arm and refuses the whole projection. **The hand-written `wire.rs` is checked** — `lower()` re-parses its own bytes with `from_json_bytes`, `format` comes from `ir::EXECUTABLE_PROJECTION_FORMAT` not a literal, and M1 proves operand semantics are caught downstream. **Byte-limit classification is right** (`is_io()` → `ResourceExhausted`, asserted by the `bytes - 1` case). **AC-4 is tested from both sides.** **The dependency split is honest** — all three new deps are dev-only, and `consumer.digest() == projection.bound().digest()` verifies rather than asserts the two-IR claim.

### One clarification for the PR body

"Draft: generated activation acceptance is blocked" reads as though the backend qualification is unverified. It's narrower, and the narrower statement is stronger. Today's running lane compares all eight assignments of the compiled generated program against native evaluation **and** against independently written truth equations, and compares native implication activation counts against independent `expected_entries`. Only the *generated* activation counts — the ones needing the LLVM export — are unreachable.

That separates "we cannot yet measure generated activation" from "we don't know whether the generated code is correct". Your evidence supports the first only, and TM-006 (`⛔ Activation pending`), Task-020 and the README all already say so correctly.

kreneskyp:
Addressed SR-114 in `d58ca7a`:

- **FND-001 fixed:** the test now asserts the normative defaults `(10_000, 64, 16_777_216)` independently. Repeating M2 (all defaults changed to `usize::MAX`) fails this assertion. After restoring production, all six targeted tests pass. FR-009 now clarifies that both node and depth ceilings also apply upstream in the native checker.
- **FND-004 fixed:** retain `unsupported_profile` and both format versions, without pinning the downstream diagnostic's prose.
- **FND-002 deferred:** retain the ten-minute hosted cost cap while testing stays local. Qualify timing/caching before enabling hosted execution; no cloud timing claim or dispatch.
- **FND-003 deferred:** the reviewed 140-entry inventory is accurate, including both revision-qualified IR sources. Automated maintenance remains an A-owned follow-up before distribution; this patch does not add a new audit framework.

Formatting and strict all-target/all-feature Clippy also pass. The original independent review and its baseline/verdict are preserved.

The PR body now states the evidence precisely: all eight generated truth results and native activation counts pass; only generated activation counts are blocked by C's LLVM 3.1.0 reader qualification. PR remains draft, and the required activation/spec/gap acceptance gates remain open.

kreneskyp:
## Independent re-review at `d58ca7a` — CONDITIONAL (both actioned findings closed)

Artifact: `reviews/26-09-09-lc04-hard-limit-remediation-rereview.md` (SR-115, pushed, quire-validated).

**The hard-limit fix is stronger than the one I asked for** — the ceilings are bound *per axis*, not just in aggregate:

| # | Mutant | at `813069f` | at `d58ca7a` |
|---|---|---|---|
| M2 | All three defaults → `usize::MAX` | all 266 pass | **FAIL** `native_lowering.rs:225` |
| M2a | `nodes` 10,000 → **10,001** only | not probed | **FAIL**, same assertion |
| M2b | `depth` 64 → **65** only | not probed | **FAIL**, same assertion |
| M2c | `bytes` 16 MiB → **+1** only | not probed | **FAIL**, same assertion |

The single-axis mutants are the ones that matter — a tuple assertion can easily end up with one axis carrying the whole check. All three constants are individually bound, which is what FR-009's three separate numbers require.

**FND-004: I checked the new assertion is live, not vacuously true.** Inverting it temporarily prints the message the test actually receives:

```
expected cargo-llvm-cov 0.9.0 / llvm.coverage.json.export 3.0.1;
found 0.9.0 / llvm.coverage.json.export 3.1.0
```

Both substrings are genuinely present, so `contains("3.0.1") && contains("3.1.0")` keeps the version tripwire and survives an upstream rewording. `assert_eq!(report["state"], "unsupported")` and the `unsupported_profile` code assertion are unchanged and still precede it, so when codegen accepts export 3.1.0 the `state` assertion fires first — the deliberate tripwire is intact.

### Gates at `d58ca7a`

`fmt` ✓ · `clippy --all-targets --all-features -D warnings` ✓ · full suite green, 0 failures, activation lane still explicitly ignored ✓

### The FR-009 sentence is the right kind of clarification

> The native checker also caps source nodes at 10,000 and depth at 64; lowering retains these ceilings as an additional guard over already checked packages.

That records what `checking.rs:68` and `linking.rs:44` already implement, and it's what stops a later reader deleting the depth bound as dead.

### Verdict

CONDITIONAL, and the two open lows are only the ones you deferred — FND-002 (CI job budget) and FND-003 (ungated license inventory) from SR-114, carried forward unchanged. Nothing new. Both are your call and neither is affected by this change.

**One observation:** the deferral is recorded only in the PR body, alongside the SR-114 cross-reference. Both are accurate — but a PR body isn't part of the record after merge, and `plan/Plan-008-native-lowering/plan.md` already carries the LC04 accounting and would keep them where the next author looks.

### Re-confirmed unchanged

TM-006 still marks FR-009-AC-5 and TC-094 `⛔ Activation pending`, Task-020 is still `in_progress` with unchecked subtasks, and the activation lane's `#[ignore]` still names its cause. Nothing was quietly reclassified while the findings were being closed — which is the thing most worth checking in a remediation pass.

Issue #6, opened 2026-09-07
[QSL#6] [LC05] Qualify Quire source integration and composed ecosystem workflows


kreneskyp:
## Required implementation language and containment

Owner directive (2026-09-07): first-party production and possible qualification paths must be **Rust**, including libraries, CLI/tools, generators, validators, canonicalization, result adapters, test assertions and fixture/CI audits. "Glue", prototypes, scripts and tests are not exemptions; a Rust wrapper around non-Rust semantic logic is not remediation.

- Every new TypeScript choice needs explicit owner approval for its exact scope. Any other executable language must be brought to the owner's attention; do not silently introduce or expand it.
- Quoin's existing implementation is explicitly retained for now: **contain its spread, do not schedule a wholesale Quoin rewrite**. Keep new reusable semantics in Rust behind versioned, structured interfaces. Existing Node/TypeScript is not blanket permission for new components or expanded language scope.
- Inventory existing JavaScript/Python/shell and executable generated/inline CI logic; record a Rust remediation owner or an explicit owner disposition. Data schemas and foreign-language fixture samples must be distinguished from executable checks. External tools/hosts/bindings need a scoped disposition, not an invented exemption.
- Reuse shared assurance contracts and Quoin evidence ownership; do not recreate local evidence frameworks. Classify each helper as an existing shared capability, a missing reusable capability, or domain-specific logic before porting.
- Complete /specify and /spec-review for the changed requirements/interfaces, resolve findings, then implement and verify. Preserve ongoing work and actual evidence provenance. Do not claim a language port or qualification complete from a policy update.
- Filament repositories/tickets are excluded from this remediation. Interface consumption does not authorize producer changes.

Before closure: record executable-path inventory, approved dispositions, Rust implementation/test/CI parity and failure-path evidence for the reviewed scope. Existing Quoin retention is the only accommodation recorded here; this block grants no new TypeScript approval.

---

## Required spec cycle — owner directive

**Every change requires the spec cycle before implementation: `/specify` → `/spec-review` → resolve findings and repeat as needed → implement and verify against the reviewed specification.** Architecture notes, ticket acceptance checklists, an ADR draft, passing code tests, or `quire validate` alone do not satisfy this gate.

[...]

PR #14, opened 2026-09-09
LC05: compile mapped native clauses with original source correspondence

kreneskyp:
Mapped callers previously had to coordinate parse/link/check/package themselves and retain original document correspondence separately. `mapped::compile` now accepts a verified source map and one authored clause binding, runs the existing compiler stages, and returns the native package with its map. Failures retain their typed stage cause and exact original regions; package paths do not acquire invented source spans.

The mapped parent workflow exercises healthy, violating and dangling-reference inputs through the real compiler and runtime. Wrong language, malformed/foreign/multiple clauses, resolution/type failures, stage exhaustion and fresh retries are covered. This advances #6; C's actual Quire producer adoption and the remaining standalone workflow stay open.

Validation: 269 tests and three compile-fail doctests pass; four pre-existing ignored tests retain their separate lanes. Formatting, strict Clippy, rustdoc and Quire validation pass. QUOIN all-set spec reviews SR-125–132, author code/Rust review SR-133 PASS, scoped gap report SR-134 CONDITIONAL for broader LC05 work. Reviews occurred at PR readiness as directed. All checks ran locally and serially; CI remains manual-dispatch only.


kreneskyp:
## MERGEABLE

Reviewed against `main` on `agent-a/lc05-native-workflow` (428e19a). Gates run locally in an isolated worktree with a private `CARGO_TARGET_DIR`; no fabricated results.

### Gates

```
    Finished `dev` profile ... in 18.80s
    272 passed (269 tests + 3 compile-fail doctests); 0 failed; 4 ignored
```

`[lints.clippy] all = "deny"` and `unsafe_code = "forbid"` are intact; no new `#[allow]` outside the two `#[allow(dead_code)]` on shared test fixture modules, which the repo already uses.

### Findings

**1. `src/mapped.rs:140,148` — `language: &str` is a stringly-typed domain value on a public API (medium).**
`compile()` takes the language tag as a bare `&str` and admits it with `if language != "ix:native"`. A caller that passes `"ix-native"`, `"IX:native"` or `"ix:native "` gets `Code::UnknownLanguage`, indistinguishable from a genuinely unsupported language, and the admissible set lives at a single literal that the next language tag will turn into an `||` chain. Fix: a `LanguageTag` newtype (or closed enum) with `TryFrom<&str>`/`FromStr`, so admission happens once at construction and `compile` cannot be called with an unvalidated tag.

**2. `src/mapped.rs:30` (`CompileCause`), `src/mapped.rs:17` (`CompileLimits`) — public enum/struct with no `#[non_exhaustive]` (medium).**
Adding a compiler stage (a lowering cause, a new limits field) is a breaking change for every downstream `match`/struct literal. This applies to the whole stack (`ExecutionOutcome` #15, `InputReadCause`/`InputReadStage` #16, `ModelSourceCause` #17, `RunCause` #18) — raising it once here. The repo has no existing precedent either way, so this is a portable-Rust default, not a repo-idiom violation.

**3. `src/mapped.rs:87` — `original_spans() -> Result<Option<Vec<LocatedSpan>>, Box<Diagnostic>>` (low).**
Three outcomes (no native diagnostic / mapped regions / refusal to guess) encoded as a nested `Result<Option<Vec<_>>>`. Callers must distinguish `Ok(None)` from `Ok(Some(vec![]))` — the test at `tests/mapped.rs:335` already has to special-case `phase.is_none()` to know which to expect. A small three-variant enum (`NoLocation` / `Regions(Vec<_>)`) states the invariant.

**4. `src/mapped.rs:53` — inconsistent field visibility on `CompileError` (low).**
`binding` and `mapping` are private with accessors; `cause` is `pub`. Pick one — the accessors exist precisely so the failed request cannot be mutated, and `pub cause` makes the type partly destructurable anyway.

### Gap analysis

- **Requirements → tests:** all five FR-022 ACs bind. `TC-095` ↦ `tests/mapped.rs:120` (AC-1, AC-5), `TC-096` ↦ `:234` (AC-2, AC-3) and `:318` (AC-4). All new tests use the bare `use ix_trace_rs::trace;` import, so the markers are readable by the coverage stage (the path-qualified form binds nothing) — checked, none are path-qualified.
- **Matrix rows:** `spec/native-workflow/tests.md` (TM-007) rows for FR-022-AC-1..5 all resolve to a tagged test. No unbacked rows.
- **Matrix registration (low):** TM-007 is not referenced from `spec/spec.md` — `spec.md` names only TM-003 and TM-004, and the root matrix TM-001 carries zero FR-022..FR-027 rows. TM-005 and TM-006 are equally unreferenced, so this is a pre-existing repo pattern rather than a regression from this PR, but this stack adds six more FRs to an index nothing points at. Worth one follow-up ticket for the whole set, not a block on this PR.
- **Code → requirements:** every new item in `src/mapped.rs` is reachable from FR-022's stated behavior. No orphan public surface.
- **Tests:** no tautologies found. `mapped_parent_workflow_...` reconstructs the body from the mapped original regions and compares byte-for-byte (`tests/mapped.rs:186`), which fails if the mapping is wrong; the refusal test asserts the mapped slice equals the diagnostic's body slice rather than asserting `is_err()`.

No stubs, `todo!()`, `dbg!`, panic on untrusted input, `as` casts, or unbounded loops in the added source.

kreneskyp:
### Follow-up: broader design/architecture pass (verdict unchanged — MERGEABLE)

Re-read `src/mapped.rs` and `tests/mapped.rs` against a wider smell checklist than the first pass. Nothing here changes the verdict; four additions, ranked.

**5. `tests/mapped.rs:57-113` — `mapping()` is a hand-rolled Markdown-fence/CRLF/indentation emitter (medium).**
45 lines of `push_str("  ")`, `strip_suffix('\n')`, `split_inclusive('\n')` and manual `Segment` bookkeeping, building the fenced original *and* its byte correspondence in the same loop. This is the producer side of the very correspondence under test, written ad hoc: if the fixture's segment arithmetic is wrong in the same direction as `SourceMap::verify`, the round-trip assertions at `:186` still pass. It is also a second implementation of what C's Quire extractor will emit, kept in sync by hand. Fix direction: build the fixture from a declarative table of (line, indent, terminator) and derive both sides, or generate it from the same layout rules `Layout` already names.

**6. `tests/mapped.rs:120-231` — one test asserting ten unrelated things (medium).**
`mapped_parent_workflow_preserves_authorship_and_runtime_results` covers original/body digest retention, clause binding identity, discontiguous span reconstruction, an out-of-range span refusal, *and* a three-case runtime evaluation loop (healthy / violating / dangling) — 110 lines carrying both AC-1 and AC-5. A failure anywhere in the first half prevents the runtime half from running at all, so AC-5 silently stops being exercised whenever AC-1 regresses. Split at least the mapping assertions from the runtime loop.

**7. `src/mapped.rs:138-146` — six-parameter `compile()` (low).**
`(mapping, language, binding, formal_identity, models, limits)`. The types are distinct enough that a call-site swap won't compile, so this is altitude rather than a correctness risk — but `language` and `formal_identity` are both properties *of the mapping*, and bundling them into the verified `SourceMap` (or a `MappedRequest`) would make an unverified combination unrepresentable instead of merely wrong.

**8. `src/mapped.rs:147-186` — the `compile_native` closure mixes admission with orchestration (low).**
One 40-line closure does language admission, parse, clause-arity check, clause-name check, binding construction, link, check and package. The three refusal paths each open-code the same `crate::diagnostic::error(body, code, phase, 0, 0, msg)` shape with two magic `0` span arguments. Extract the admission checks (`language`, arity, name) from the stage pipeline; give the "no meaningful span" case a named constructor rather than `0, 0`.

PR #15, opened 2026-09-09
LC05: execute native packages and retain request-bound results

kreneskyp:
Callers previously had to keep a temporary validated context alive and coordinate separate runtime stages; failed validation discarded the offered request. `runtime::execute` now validates and evaluates one native package and returns an immutable report retaining the package, all offered input bytes and the exact authored selection for every returned outcome.

Failed validation has no evaluation observations. Completed execution carries its actual Boolean; stopped execution retains its typed cause, measured work, implication-event prefix and cost-model version. The existing separate APIs retain their signatures and behavior. This is an in-process native API under #6; standalone file intake and actual Quire extraction adoption remain open.

Stacked on #14; review and land #14 first. This PR's implementation is 9c3e5ff. Four new integration tests cover aggregate truth, failed/foreign requests, stage exhaustion, cancellation between stages, fresh retries, and operation captures/frame refusals.

Validation: full suite passed 273 tests and three compile-fail doctests, with four existing ignored tests. After the review's exhaustive-match cleanup, all four targeted tests and strict Clippy passed again. Formatting, rustdoc, minimal build, fixture audits and Quire validation passed. QUOIN all-set reviews SR-135–142 and actual code/Rust review SR-143 are complete; SR-144 retains broader LC05 gaps. Checks ran locally; hosted CI remains dispatch-only and was not run.


kreneskyp:
## MERGEABLE

Reviewed against `agent-a/lc05-native-workflow` (#14) on `agent-a/lc05-native-execution`. Gates run locally on the head branch.

### Gates

```
    276 passed (273 tests + 3 compile-fail doctests); 0 failed; 4 ignored
```

### Findings

**1. `src/runtime/execution.rs:35` — `cost_model: &'static str` is a wire-visible version carried as a bare string (medium).**
This value reaches consumers verbatim (`#18` emits it as the `cost_model` field of `native-run-result/1`, and `tests/runtime_execution.rs:81` asserts the literal `"native-ref-cost/1-draft"`). As `&'static str` it has no catalog, no `FromStr`, and no compile-time relationship to the evaluator that produced it: a consumer can only string-compare, and a version bump is invisible to the type system. Fix: `CostModelVersion(&'static str)` newtype with `as_str()`/`Display`, defined next to the evaluator that owns it.

**2. `src/runtime/execution.rs:23` — `ExecutionOutcome` is public with no `#[non_exhaustive]` (medium).**
A third outcome (a pre-validation refusal, a lowering result) breaks every downstream `match`. Same class as finding 2 on #14.

**3. `src/runtime/execution.rs:83` — `ExecutionReport::truth()`'s `EvaluationOutcome::Refused` arm has no test (medium).**
`tests/runtime_execution.rs` exercises `Completed`, `Incomplete(ResourceExhausted)`, `Incomplete(Cancelled)` and `ValidationFailed`. The two `Refused` occurrences in the file (`:146`, `:323`) are `ValidationStatus::Refused`, not `EvaluationOutcome::Refused`. So the "defensive invariant refusal" branch of the new outcome — the one the doc comment at `:29` singles out — ships untested, and FR-023-AC-1's "concrete truth, incompleteness or a defensive invariant refusal" is only two-thirds bound. Either construct an evaluation that refuses, or state in FR-023 that the arm is unreachable-by-construction and assert that.

**4. `tests/runtime_execution.rs:255-262` — poll-count assertion is coupled to validation's internals (low).**
`assert_eq!(combined_polls, validation_polls + 1)` where `validation_polls` was measured by a *separate* `validate()` call earlier in the test. It does verify that one poll closure is forwarded across both stages, but it also fails whenever validation's polling frequency changes with no behavioral regression. A state flag ("evaluation polled at least once after validation completed") tests the requirement without pinning the count.

**5. `src/runtime/validation.rs:185-188` — `validate()` now allocates a `Box<FailedValidation>` on every failure only to discard it (low).**
`validate_retaining(...).map_err(|failure| failure.report)` boxes the retained request, then drops the box and the two moved fields. The pre-existing public API pays an allocation on its error path for a caller that never asked for the retention. `validate_retaining` could return the parts unboxed and let `execute` do the boxing.

### Gap analysis

- **Requirements → tests:** FR-023-AC-1..4 all have tagged tests. `TC-097` ↦ `tests/runtime_execution.rs:41` (AC-1) and `:280` (AC-4); `TC-098` ↦ `:95` (AC-2, AC-3) and `:171` (AC-3). Bare `use ix_trace_rs::trace;` import, so the markers bind.
- **Matrix rows:** the four TM-007 rows added for FR-023 all resolve. No unbacked rows.
- **Code → requirements:** `EvaluationReport::into_parts` and `ValidatedContext::into_request` are `pub(in crate::runtime)` plumbing for FR-023 — correctly scoped, no new public surface without an owning requirement. Exception: the `Refused` arm noted in finding 3.
- **Tests:** not mock-shaped. `execute()` results are cross-checked against separately-run `validate()`/`evaluate()` calls (`:143-152`, `:226-229`), so the aggregation is measured against the real stages rather than asserted against a double.

No `unwrap`/`expect`/panic on caller input in the added source, no `as` casts, no unbounded loops. `execute` forwards one `impl FnMut() -> bool` across both stages with no lock or await held.

kreneskyp:
### Follow-up: broader design/architecture pass (verdict unchanged — MERGEABLE)

Re-read against a wider smell checklist. Three additions, ranked; verdict unchanged.

**6. `src/runtime/execution.rs:26-38` — the `Evaluated` variant duplicates `EvaluationReport`'s field list by hand (medium).**
`EvaluationReport` already owns `(outcome, usage, events)` plus `cost_model()`. `into_parts()` (`src/runtime/evaluation/api.rs:117`) unpacks it into a 3-tuple, and the variant then re-spreads that tuple into four inline fields. Three declarations of the same shape must now be edited together: add a counter to `EvaluationUsage` and nothing breaks, but add an *observation* to `EvaluationReport` and you must remember to thread it through `into_parts` and the variant. Fix: have the variant hold one owned struct (`OwnedEvaluationReport` / `EvaluationParts`) so the shape is declared once.

**7. `tests/runtime_execution.rs:95-165` — the failure table is a stringly-typed tag dispatched by an `if` chain (medium).**
The loop is `for case in ["field","foreign","unavailable","work","cancel","details"]` followed by five `if case == "..."` / `matches!(case, "field" | "details")` branches that mutate the draft, the selection, the input and the limits in turn, plus a sixth `matches!(case, "work"|"cancel")` that decides whether a retry runs. A typo in any literal silently disables that case's setup while the loop still iterates and the assertions still pass against an unmodified fixture — a case that stops testing anything looks identical to one that passes. Fix: an enum (or a struct of closures) per case, so the compiler enforces that every case is handled everywhere it is branched on.

**8. `src/runtime/execution.rs:105-144` — both arms of `execute` construct the same five-field `ExecutionReport` (low).**
The `Ok` and `Err` arms differ only in how `input`/`selection`/`validation_usage` are obtained and which `outcome` is set. Computing `(input, selection, validation_usage, outcome)` in the match and constructing the report once afterwards removes the duplicated literal — cheap, and it makes a future sixth field impossible to add to only one arm.

Nothing new on layering: `execute` stays inside `runtime`, touches no filesystem, env, or stdout, and holds no lock or borrow across the poll.

PR #16, opened 2026-09-09
LC05: read verified native snapshot and invocation bytes

kreneskyp:
Native runtime inputs could be constructed in Rust but could not be read back from their serialized bytes. `Snapshot::read_verified` and `Invocation::read_verified` now verify an exact selected reference, decode the closed native-state-input/1 envelope/body, and repeat the existing structural constructor checks before returning an artifact.

Serde owns JSON parsing. Version/kind selection precedes typed body decoding; stale selections, malformed fields and exhausted budgets retain their stage and typed cause. Successful reads preserve external bytes and digests, including whitespace and object field order. Model-aware validation still happens during execution. The existing package object adapter is shared with runtime intake; the tagged Absent variant explicitly rejects extra fields.

Stacked on #15 (after #14). Advances #6 without closing it: standalone model intake/command and actual Quire extraction adoption remain open. Five new integration tests cover all value variants, both artifact roles, closed/duplicate/missing fields, limits/retries and real execution of reread inputs.

Validation: full suite passed 278 tests and three compile-fail doctests, with four existing ignored tests. After sharing the object adapter, all 18 package-reader and five runtime-reader tests plus strict Clippy passed. Formatting, rustdoc, minimal build, fixture audits and Quire validation passed. QUOIN all-set reviews SR-145–152 and actual code/Rust review SR-153 are complete; SR-154 retains broader LC05 gaps. Checks were local; hosted CI remains dispatch-only and was not run. No new dependency version or license change.


kreneskyp:
## CHANGES REQUESTED

Reviewed against `agent-a/lc05-native-execution` (#15) on `agent-a/lc05-runtime-intake`. Gates run locally on the head branch. Nothing here is a correctness failure I can demonstrate; the request is for the two structural items (1 and 2), which are cheap now and expensive after consumers exist.

### Gates

```
    281 passed (278 tests + 3 compile-fail doctests); 0 failed; 4 ignored
```

Confirmed the new decoders do **not** inherit the crate's `unbounded_depth` escape hatch: `disable_recursion_limit()` is called only at `src/package/intake.rs:405`, and `reading.rs` uses plain `serde_json::from_slice`/`from_str`, so serde's 128-level nesting guard still applies to untrusted runtime input. Worth a comment at the call site so a future "share the intake decoder" refactor doesn't silently remove it.

### Findings

**1. `src/runtime/reading.rs:77-125` — failure `stage` and `code` are carried on mutable locals in a side-channel, not on the error (high).**
`read()` is a 50-line closure with eight early returns, and the `stage`/`code` reported to the caller are whatever the last assignment left in two captured `let mut` bindings (`:77`, `:78`) read *after* the closure returns (`:125`). There is no compile-time link between a return and the code it reports. Concrete failure: `:96` sets `code = Code::UnknownWire` *inside* the version branch; the kind check at `:101` deliberately relies on the assignment at `:93` still holding. Add a check between `:101` and `:106` that needs its own code and forget the assignment, and it reports the previous check's code with the new check's message — no warning, no test failure, and the only symptom is a consumer branching on the wrong `Code`. Equally, hoisting `:96` one line up makes every later failure report `UnknownWire`. Fix: carry `(stage, code)` on the `InputReadCause` variants (or return a `(stage, code, cause)` triple from each arm) so the compiler enforces the pairing, and drop the `FnMut`-as-try-block.

**2. `src/runtime/reading.rs:30` — `InputReadCause::Selection(&'static str)` discriminates five distinct refusals by message text (high).**
Byte-limit exceeded, digest mismatch, unadmitted version, wrong artifact kind and stale identity all become one variant separated only by a human-readable literal — in a module whose own doc comment at `:26` says "without recovering fields from a formatted message". The tests prove the gap: `tests/runtime_reading.rs:113` can only assert `matches!(failure.cause, InputReadCause::Selection(_))` and has to fall back on `failure.code` for the real distinction, which cannot separate byte-limit from digest-mismatch (both would need `code` alone) or kind-mismatch from a generic invalid input. Fix: one variant per refusal, or a `SelectionRefusal` enum with stable codes.

**3. `src/runtime/input.rs` — 40 hand-written `deserialize_with` attributes where the type should carry its own decoder (high).**
Nine are `deserialize_with = "deserialize_object"`, and the file adds six free functions (`deserialize_identity`, `deserialize_requirement`, `deserialize_symbol`, `deserialize_digest`, `deserialize_required_option`, `deserialize_absent`) that are attached by name at each use site. Every one is copy-pasted, differing by one token, and every new field of an existing type must remember the right incantation. A field that silently *omits* `deserialize_with = "deserialize_object"` still compiles and still passes `deny_unknown_fields` — it just quietly accepts a positional JSON array for that one record, which is exactly the hole `from_object` exists to close. The abstraction is already in the tree: `Object<T>` at `src/serde_object.rs:27`. Wrapping the field *types* (or newtyping `ir::SymbolName`/`ByteDigest` at the wire boundary with their own `Deserialize`) makes the correct behavior the default instead of a per-field checklist item.

**4. `src/runtime/reading.rs:121-122` — construction result is patched after the constructor returned (medium).**
`artifact.bytes = bytes.to_vec(); artifact.digest = digest;` reaches into a value `Artifact::new` just produced and overwrites two of its fields. The behavior is intended and documented, but the invariant "bytes and digest correspond to the draft" now holds for constructor output and not for reader output, enforced by nothing. If `Artifact::new` ever gains a check tying bytes to the draft, this path silently violates it. Fix: a constructor that takes the external bytes (`Artifact::from_external(identity, draft, bytes, digest, limits)`).

**5. `src/runtime/input.rs:438` — `deserialize_digest` reconstructs a formatted string to reuse `FromStr` (medium).**
`format!("sha256:{hex}").parse()` — the wire carries bare hex, the decoder re-adds the prefix so it can round-trip through the display spelling. Allocation and a string round-trip at a wire boundary to make up for a missing constructor. Pair it with the existing `serialize_digest` (`:431`), which does the mirror-image `trim_start_matches("sha256:")`: two functions maintaining one encoding by string surgery at both ends. Fix: `ByteDigest::from_hex(&str)` / `fn hex(&self) -> &str`.

**6. `src/runtime/reading.rs:62-72` — `Envelope { version: String, kind: String }` compared against magic literals (medium).**
`"native-state-input/1"` at `:95` and `T::KIND` at `:101` are the wire contract, spelled as free-floating string literals with no catalog and no single owner (`model_source::FORMAT` in #17 and `"native-run/1"` in #18 repeat the pattern with three more literals). There is no schema for `native-state-input/1` either — `schemas/` contains only `native-linked-package-1.schema.json`. A declared wire format with neither a type nor a schema drifts silently. Fix: a `WireFormat` newtype/const module, and a schema alongside the package one.

**7. `src/runtime/reading.rs:16` — `InputReadStage` has no `Display`/`as_str`, unlike every sibling enum (medium).**
`Code::as_str` and `Phase::as_str` exist; `InputReadStage` does not, so #18 open-codes the four strings in a `match` at `src/command/output.rs:93-97`. Two lists of stage names that must be kept in sync by hand, in different crates' worth of concern (core vs. presentation). Add the accessor here, where the enum is introduced.

**8. `src/runtime/input.rs:70-160` — `ValueNode`'s ten variants each re-declare their per-field decoders (low).**
`Enum`, `Record`, `Reference`, `Object` all repeat `deserialize_with = "deserialize_object"` on `declaration`/`identity`, and `Absent` needs a bespoke `deserialize_absent` (`:493`) to close serde's internally-tagged unit-variant hole. The `Absent` workaround is correct and well-commented — the smell is that it is the tenth instance of the same manual guard rather than a property of the enum. Same fix as finding 3.

More than eight candidates exist; the remainder are all in the same two categories (repeated serde attribute plumbing, and magic wire literals) and are covered by findings 3 and 6.

### Gap analysis

- **Requirements → tests:** FR-024-AC-1..4 all bind. `TC-099` ↦ `tests/runtime_reading.rs:68` (AC-1, AC-3) and `:279` (AC-1, AC-4); `TC-100` ↦ `:88` (AC-2), `:128` (AC-2, AC-3) and `:212` (AC-3, AC-4). Bare `use ix_trace_rs::trace;`, so the markers bind.
- **Matrix rows:** the four TM-007 rows added for FR-024 all resolve to a tagged test. No unbacked rows.
- **Code → requirements:** `InputReadStage::Construction` and the byte/digest retention behavior are named in FR-024's Behavior section. `src/serde_object.rs` is new shared plumbing with no owning requirement of its own — acceptable as an extraction of existing FR-020 code (the doc header says FR-020/024), but the `Object<T>` wrapper it introduces is genuinely new surface; if it is the intended fix for finding 3, FR-020 or FR-024 should say so.
- **Tests:** behavioral, not mock-shaped. `reads_every_value_variant_...` re-reads the constructor's own bytes *and* a re-pretty-printed variant and asserts draft equality (`:79`), so a decoder that dropped a field fails. No tautologies found. The one test-structure note: `full_snapshot()` (`:31`) is a 35-line fixture that every test in the file depends on — a change to it moves all five tests at once.
- **Wire contract:** `deny_unknown_fields` is present on every new externally-supplied record, duplicate-key and positional-array refusals are tested (`:128`), and the `Absent` unit-variant hole is explicitly closed and tested. That part is done properly.

kreneskyp:
Required findings 1/2 are corrected at 0016103 (branch head e7d240b): each selection refusal has a typed cause with observed context, and one exhaustive cause match determines stage/code. Envelope/body JSON errors retain their original diagnostics separately. Tests now assert the distinct variants and payloads.

Also corrected findings 4/7: the artifact constructor owns external-byte adoption, and InputReadStage owns its stable wire spelling. Reader/writer share the input-format constant and the default Serde recursion guard is documented.

Remaining follow-ups are explicit in SR-153: type-owned decoding (3/8), bare-hex digest conversion (5), and the broader format catalog/schema (6). This does not claim those comments are resolved. Cross-cutting error inventory is #27; matrix status validation is #28; typed CLI dispatch is #29. All three are on private board 20.

Local correction checks passed: 278 tests + 3 compile-fail doctests (4 existing ignored), strict all-targets/all-features Clippy, fmt, minimal build, rustdoc with warnings denied, and scoped Quire validation. QUOIN all-review and actual code/rust review artifacts were updated in place at PR readiness. No hosted CI ran.

PR #17, opened 2026-09-09
LC05: expose the source-aware rule-model frontend

kreneskyp:
Standalone callers could not construct native models from source without copying the fixture frontend. This promotes that existing source-aware Rust implementation to `model_source::read`, with explicit `native-rule-model/1` selection, retained typed errors and caller-lowered limits. `ModelDraft::admit` uses the existing native model checks. Test helpers now call the library.

The fixed preexisting model artifact remains byte-identical. Five new public tests exercise source forms/loci, refusals, incomplete limits and real state/operation execution. Full local suite: 283 tests and three compile-fail doctests passed; four existing assurance tests remain ignored. Formatting, strict Clippy, rustdoc, cached minimal build, fixture audits and Quire validation passed. No hosted CI was dispatched.

Stacked on #16; review after #14 → #15 → #16. Refs #6. FR-025 / Task-024; QUOIN all spec reviews SR-155–162, author code/Rust review SR-163 PASS, scoped gap review SR-164 CONDITIONAL because the standalone command and broader LC05 integration remain open. Advisor version detection is unavailable; evidence methods are explicitly recorded as author judgment. This is PR-ready engineering, not independent assurance or completion of LC05.


kreneskyp:
## CHANGES REQUESTED

Reviewed against `agent-a/lc05-runtime-intake` (#16) on `agent-a/lc05-model-source`. Gates run locally on the head branch.

### Gates

```
    286 passed (283 tests + 3 compile-fail doctests); 0 failed; 4 ignored
```

### Findings

**1. `src/located_json.rs:44-56` — original-span recovery is pointer-address arithmetic against the source buffer (high).**
```rust
let start = selected.as_ptr().addr()
    .checked_sub(original.as_ptr().addr())
    .ok_or(Error::ForeignOccurrence)?;
```
This subtracts two allocation addresses to guess a byte offset, then validates it with `original.get(start..end) != Some(selected)`. It is not UB — no pointer is dereferenced — and the content check keeps it from producing a *wrong* span. But it is a fragile substitute for a locating deserializer, and it is now the sole provenance mechanism for every IR declaration source span the model frontend emits. Failure modes: (a) any `RawValue` that is not a borrow into this exact buffer — an owned `Box<RawValue>`, a value routed through `serde_json::Value`, a future serde_json that copies on escape sequences — degrades to `ForeignOccurrence`, i.e. the whole model refuses with `Code::InvalidSourceMap` and no indication that the *decoder*, not the document, changed; (b) `checked_sub` rejects any allocation placed below the buffer but silently admits one placed above it whose bytes happen to match at the computed offset. The contract "the `raw` you pass must be borrowed from `source`" is enforced by arithmetic and stated nowhere in the doc comment. Fix direction: a span-tracking deserializer (or record offsets during one tokenizing pass) so the position is produced by the parse rather than reverse-engineered from it; failing that, document the borrow contract on `decode` as a hard precondition and add a test that an owned `RawValue` is refused.

**2. `src/model_source.rs:181,192,195` — the doc claims typed causes; the type discriminates by message string (high).**
`/// Original frontend or admission failure; no message parsing determines its kind.` sits directly above `Invalid(&'static str)` and `Resource(&'static str)`. Those two variants carry **seven** distinct conditions separated only by their literal: declaration/field/variant/operation entry ceilings (`:388`, `:405`, `:429` and the loop in `decode_model`), type-depth (`:479`), duplicate scalar identity (`:467`) and unknown scalar identity (`:490`). `ModelSourceError::code()` then folds all of them onto two `Code`s, so a caller that needs "you declared the same scalar twice" versus "you referenced a scalar that doesn't exist" has literally nothing to match on but the string — the thing the doc says it avoids. The doc comment describes an intention the code does not implement. Fix: a variant (or a stable sub-code) per condition.

**3. `src/model_source.rs:467,490` — `ModelSourceCause::Invalid` is never exercised, and no acceptance criterion owns it (high).**
`tests/model_source.rs` covers `Decode` (`:135`), `Formal` and `Admission` (`:187`) and `Resource` (`:202`). Neither "duplicate scalar identity" nor "unknown scalar identity" has a test, and FR-025's ACs never name them — AC-3 lists "profile, decode, formal and admission failures". So this is code with **no owning requirement and no covering test**: a model declaring `scalars: [{"name":"X"...},{"name":"X"...}]`, or a field typed `{"kind":"scalar","name":"NoSuchScalar"}`, exercises a path nothing in the matrix asserts. Either add an AC + test or state why the IR constructors already make it unreachable.

**4. `src/model_source.rs` is a 728-line module doing three unrelated jobs (medium).**
Wire structs (`:23-175`), decode-with-budget (`:370-450`), and IR lowering (`:455-700`). The three concerns are interleaved by dependency rather than separated, `ScalarTable` is a mutable accumulator threaded through six `lower_*` functions, and `DecodedModel`/`RuleModel` are six-field parallel structures whose fields must stay in one-to-one correspondence by hand — add a declaration kind and you edit both structs plus `decode_model` plus `lower_model` with nothing checking you did all four. Split into `model_source/wire.rs`, `model_source/decode.rs`, `model_source/lower.rs`.

**5. `src/model_source.rs:370-450` — the entry budget is hand-threaded through five separate `checked_sub` blocks (medium).**
`remaining` is decremented in four near-identical loops that differ only by the `&'static str` in the `ok_or` (`:380`, `:388`, `:405`, `:429`). Copy-paste differing by one token, and the accounting order is load-bearing: the operation-entry loop at `:424` runs *after* `decode_items` has already decoded every operation, so a request with 10,000 operations is fully decoded and allocated before the ceiling is consulted. Fix: a `Budget { remaining }` newtype with `take(n) -> Result<()>` charged *before* the work, so the ceiling is enforced by construction rather than by five call sites remembering to subtract.

**6. `src/model_source.rs:171-179` — `bounded()` silently clamps caller limits down (medium).**
A caller asking for `source_bytes: 10_000_000` gets 1 MiB and then a `ResourceExhausted` it did not request, with no way to learn the effective ceiling. The doc says "independently clamped", so it is intentional, but a public API that quietly substitutes different limits than the caller passed produces confusing incompleteness reports. Either refuse an over-large request explicitly, or expose the effective limits on the draft/error.

**7. `src/model_source.rs:206-233` — `ModelSourceError::code()` classifies partly by scanning a foreign diagnostic vector (medium).**
The guard arm `Formal(errors) if errors.iter().any(|e| matches!(e.code, SemanticInputTooLarge | CanonicalizationResourceExhausted))` decides "incomplete vs. refused" by searching IR's diagnostic list for two specific upstream codes. That couples this crate's exit classification to an enumeration owned by `quire-contract-ir`: a third resource code added upstream is silently reclassified as `InvalidModelBinding` (refused) instead of `ResourceExhausted` (incomplete), which flips the CLI exit code from 3 to 1 in #18. Ask IR for a `is_resource_exhaustion()` predicate rather than mirroring its catalog here.

**8. `src/located_json.rs:33` — `pub fn read` applies no byte limit (low).**
`located_json` is `pub mod` (`src/lib.rs:21`), so `read` is public API that deserializes an entire `FormalSource` with no ceiling. `model_source::read` checks `source_bytes` before calling it (`:322`), but nothing makes that ordering mandatory for other callers. Either take the limit as a parameter or make the module `pub(crate)`.

### Gap analysis

- **Requirements → tests:** FR-025-AC-1..5 all bind. `TC-101` ↦ `tests/model_source.rs:42` (AC-1), `:66` (AC-2), `:236` (AC-5); `TC-102` ↦ `:135` (AC-3), `:187` (AC-4). Bare `use ix_trace_rs::trace;`, markers bind.
- **Matrix rows:** the five TM-007 rows added for FR-025 all resolve. No unbacked rows.
- **Code → requirements:** finding 3 is the reverse gap — `ModelSourceCause::Invalid`'s two conditions have no owning AC. `ModelDraft::declared_license` is retained metadata explicitly named in the FR, fine. `located_json` is new public surface introduced as "promoting the existing fixture frontend"; FR-025 describes the model frontend but not a general public JSON-provenance utility, so `pub mod located_json` is public API ahead of its requirement (see finding 8).
- **Tests:** genuinely behavioral. `public_frontend_matches_the_preexisting_frozen_model_artifact` pins the artifact bytes against the frozen `minimal.package.json` and a literal digest, so a lowering change cannot pass silently; `source_forms_and_repeated_names_keep_exact_original_occurrences` slices the original document at each recorded span and re-parses it, which fails if provenance is wrong. No tautologies, no mocks, no `#[ignore]` added.
- **Test-structure note (low):** `profile_decode_formal_and_admission_failures_retain_the_original_source` (`:135`) asserts five decode refusals, a formal failure and an admission failure in one 55-line body carrying a single AC; the later halves stop running once an earlier assertion fails.

kreneskyp:
Line-number corrections to the review above (content unchanged): finding 6's `bounded()` pre-check is `src/model_source.rs:305`, not `:322`; finding 7's `code()` match is `:232-252` with the guard arm at `:242`, not `:206-233`; finding 5's four `checked_sub` sites are `:377`, `:388`, `:405`, `:429` with the post-hoc operation loop at `:419`; finding 5's depth guard is `:478`.

kreneskyp:
Corrections are implemented at fbdf687:

- Findings 1/8: `located_json::decode` documents the exact-source borrow contract and rejects an owned `RawValue` with identical bytes; public reads require a byte ceiling. Serde still owns JSON grammar.
- Findings 2/3: duplicate and unknown scalars have separate typed causes with names. Byte, entry-category and type-depth failures retain effective budget context. FR-025 and the tests explicitly own these refusals.
- Findings 4/5/6: wire shapes, bounded decoding and IR lowering are separate modules. One budget helper charges each group before value decoding, including operation parameters/frame entries. Drafts and errors retain effective frontend limits.

The seven-category budget test places malformed values and a malformed later operation behind each exhausted budget. It asserts the specific resource cause, then admits one more entry and requires the JSON error. Existing frozen model bytes/digest remain unchanged.

Finding 7 remains an upstream follow-up under #27: the pinned IR has no resource-classification predicate. Its existing classification remains in place; no producer repository was changed.

Local checks passed: full suite (286 tests + 3 compile-fail doctests, 4 existing ignored), strict all-targets/all-features Clippy, fmt, minimal build, rustdoc with warnings denied and scoped Quire validation. QUOIN all-review and code/rust review records were updated in place. Coverage remains trace binding, not validated Tested status (#28). No hosted CI ran. The branch includes #16's correction through a merge commit, preserving reviewed history.

PR #18, opened 2026-09-09
LC05: run native workflows from selected local files

kreneskyp:
Authors can now run a native workflow from selected files with `quire-spec run <request-file>`. The command uses the public model frontend, real parse/link/check/package stages and verified snapshot/invocation readers, then emits actual native execution results. Closed JSON, explicit identities/digests and bounded single-open reads keep intake separate from semantics. Existing parse/format commands retain their behavior.

The Rust fixture generator provides healthy, violating, recorded-operation and frame-refused examples. All four generated cases ran successfully from outside the repository, with expected exit codes 0/1/0/1. Four binary tests cover identities, counters/events, stale inputs, request errors, I/O, file/count/aggregate ceilings and stopped/retried runtime work.

Validation: full local suite passed 287 tests and three compile-fail doctests; four existing assurance tests remain ignored. Formatting, strict all-targets/all-features Clippy, rustdoc, cached minimal build, fixture audits and Quire validation passed. No hosted CI ran. Instructions: `docs/native-standalone.md`.

Stacked on #17; review in order #14 → #15 → #16 → #17 → this PR. Refs #6. FR-026 / Task-025; QUOIN all spec reviews SR-165–172, author code/Rust SR-173 PASS and scoped gap SR-174 CONDITIONAL. Advisor CLI-version detection remains unavailable; method choices are explicitly author judgment. Real Quire extraction, portable evidence and deferred backend assurance remain separately owned; this PR does not close LC05 or claim production readiness.


kreneskyp:
## CHANGES REQUESTED

Reviewed against `agent-a/lc05-model-source` (#17) on `agent-a/lc05-standalone-run`. Gates run locally on the head branch.

### Gates

```
    290 passed (287 tests + 3 compile-fail doctests); 0 failed; 4 ignored
```

The four generated fixtures do run end-to-end as claimed. Line numbers below are against `agent-a/lc05-standalone-run`.

### Findings

**1. `src/command/output.rs:106-180` — `native-run-result/1` exists as an untyped `Value` mutated by string key, with no type and no schema (high).**
`report()` builds a `json!({...})` literal and then patches it with `value["stage"] = ...`, `value["status"] = ...`, `value["truth"] = ...`, `value["diagnostics"] = ...` across two match arms and three nested ones. The declared output format is therefore defined by the union of assignments across a 75-line function; nothing in the type system says which keys are present for which status, `schemas/` contains only `native-linked-package-1.schema.json`, and the only consumer contract is `tests/standalone.rs` asserting individual keys. Concrete failure: `truth` is set only in the `Completed` arm and the tests assert `result.get("truth").is_none()` elsewhere — a stray assignment in a new arm produces a result claiming a Boolean for a refused run, and nothing but a hand-written test catches it. Fix: a `#[derive(Serialize)]` result type where illegal states are unrepresentable (status as an enum carrying its own payload), and a schema next to the package one.

**2. `src/command/output.rs:59-104` — the emitted `code` field mixes two namespaces, four of which are documented nowhere (high).**
Five arms emit `Code::as_str()` values (`unknown_wire`, `resource_exhausted`, …); four emit ad-hoc literals invented here — `"io-error"` (`:63`), `"invalid-request"` (`:68`), `"invalid-digest"` (`:77`), `"invalid-identifier"` (`:78`). I grepped `docs/native-error-codes.md`: none of the four appear. So the command's own error surface is half a stable catalog and half string literals, and a consumer branching on `code` has no single list to switch over. `tests/standalone.rs:198,213` already pin two of them as a wire contract. Fix: either extend `Code` with the four intake conditions, or give `RunCause` its own documented code enum with an `as_str()`, and add both to `docs/native-error-codes.md`.

**3. `src/command.rs:98-141` — `exit_code()` is four copy-pasted `if incomplete { 3 } else { 1 }` arms, one of which open-codes the predicate (high).**
`Native`, `Model` and `Input` each call `e.is_incomplete()`; `Package` open-codes `e.code == Code::ResourceExhausted` because `PackageError` never got the accessor the other three have. Four blocks differing by one token, plus one that reimplements it. A fifth error source added without noticing the pattern gets the wrong exit code silently — and the exit code is the CLI's primary contract (0/1/2/3). Fix: one `is_incomplete()` on every cause type (add it to `PackageError`) and a single `if self.cause.is_incomplete() { 3 } else { ... }`.

**4. `src/main.rs:72-96` — command dispatch is a string-comparison chain split across two functions with two different arities (high).**
`main()` matches `if let [command, path]` then `if command == "run"`; `run()` matches `if let [command, identity, revision, path]` then validates UTF-8 for three operands *before* checking `matches!(command, "parse" | "format")` (`:41-59`), then branches again on `if command == "format"` (`:62`). The usage string is a literal in one place (`:28`) and must be edited whenever either arity list changes — #19 adds a third command and does exactly that. Concrete failure: `quire-spec run req.json extra` has arity 3, matches neither pattern, and reports the generic usage error rather than "run takes one operand"; `quire-spec format x` reports `command must be parse or format` only after passing the UTF-8 checks for operands it does not have. Fix: parse argv once into a `Command` enum (name + operands) and dispatch on that.

**5. `src/command.rs:169-175` — request-supplied file paths are joined to the request directory with no containment (medium).**
`self.directory.join(path)` where `path` is `wire::SourceFile::file` / `FileSelection::file`, straight from the JSON. `Path::join` with an absolute operand discards the base entirely, and `..` walks out. FR-026 explicitly says "absolute paths remain valid local file operands", so this is *specified* rather than accidental — but the spec does not mention `..`, there is no test for either, and the error path leaks the resolved path and OS errno (`src/command/output.rs:63`), giving a file-existence oracle. Since #19 turns these requests into artifacts meant to be exchanged, please re-confirm the decision explicitly and either add a containment mode or add a test pinning the permissive behavior as intended.

**6. `src/command.rs` — filesystem, argv-shaped exit codes and presentation JSON now live inside the library crate (medium).**
`pub mod command` puts `File::open`, `Path::join`, a `pub exit_code: u8` and `pub value: serde_json::Value` into `quire-spec-language`, while the core modules it calls (`model_source`, `checking`, `package`, `runtime`) are carefully free of I/O. `RunResult { exit_code, value }` is a CLI concern with no meaning to a library consumer. It is testable this way (and `tests/standalone.rs` drives the real binary anyway, so that is not the reason), but the boundary has moved. Consider `command` as a separate module tree that returns a typed result, with exit-code mapping in `main.rs`.

**7. `src/command.rs:58` + `tests/standalone.rs:238` — an intake ceiling is identified by its message string, and a test pins that string as the wire contract (medium).**
`RunCause::Limit(&'static str)` carries `"file bytes"` and `"selected files"`; `output::error` emits it as `details.limit`, and the test asserts `result["details"]["limit"] == "selected files"`. Renaming a human-readable message is now a breaking wire change. Same class as #16 finding 2 and #17 finding 2 — three modules in this stack now discriminate failures by literal. Fix: a `LimitKind` enum with stable names.

**8. `src/command/output.rs:166` — `EvaluationOutcome::Refused` is emitted but never tested, here or in #15 (medium).**
`tests/standalone.rs` exercises `completed` (both truths), `incomplete` (three limit variants) and `refused` *via validation failure* — never a refused evaluation. Combined with #15 finding 3 (`ExecutionReport::truth()`'s `Refused` arm untested), the whole `EvaluationOutcome::Refused` path from evaluator to CLI JSON to exit code 1 ships unexercised. FR-026-AC-3 names "actual stage/code", which covers it in principle; the test does not.

More candidates exist below this bar — chiefly `run_bytes` at `:213` being a 60-line straight-line orchestration whose stages are only separable by reading it, and the `FILES - 1` magic adjustment at `:219`.

### Gap analysis

- **Requirements → tests:** FR-026-AC-1..5 all bind. `TC-103` ↦ `tests/standalone.rs:41` (AC-1, AC-5) and `:85` (AC-2); `TC-104` ↦ `:117` (AC-3) and `:202` (AC-4). Bare `use ix_trace_rs::trace;`, markers bind.
- **Matrix rows:** the five TM-007 rows added for FR-026 all resolve to a tagged test.
- **Partial AC binding (low):** FR-026-AC-5 asserts three things — relative-path independence, "existing parse/format behavior remains", and "all execution remains local Rust". The tagged test (`:41`) exercises only the first (via `current_dir(directory.parent())` with an absolute request path — good). Parse/format regression cover exists in `tests/cli.rs`, but that file carries no FR-026 marker, so the matrix row overstates what the tagged test verifies. The third clause is a policy statement, not a testable criterion; it belongs in the NFR/policy section rather than an AC with `Verification: Test`.
- **Code → requirements:** every new module maps to FR-026. The four undocumented error codes in finding 2 are output surface with no owning catalog entry.
- **Tests:** genuinely end-to-end — they invoke `CARGO_BIN_EXE_quire-spec` as a subprocess and assert on real stdout/stderr/exit codes, with digests recomputed from the files on disk. No mocks, no tautologies. Two structural notes: `bounded_intake_and_runtime_stops_allow_fresh_default_execution` (`:202`) is a 60-line test that mutates and restores `program.native` and `snapshot-0.json` in place across eight separate limit assertions — a failure partway through leaves the fixture in a modified state and the remaining assertions meaningless; and `invoke()` (`:11`) decides which stream to parse by `!result.stdout.is_empty()`, so a run that wrote to *both* streams would be read from stdout and the stderr diagnostics silently ignored.

kreneskyp:
The four high findings are corrected at d7437e2:

- Typed Serialize payloads define every validation/evaluation status; only Completed carries truth. The JSON result wrapper is immutable. A native result envelope schema validates real CLI output, with negative controls for missing fields and truth on refusal.
- The four existing command code spellings are now in Code and its catalog; output-failure is also explicit. RunCause/Code are non-exhaustive. LimitKind distinguishes the two intake ceilings. One code/cause classifier handles incompleteness, and PackageError now exposes the accessor; a cancelled-package adapter control pins exit 3.
- One typed argv parser runs before I/O. Command-specific arity tests pass. Absolute and parent-relative file operands are explicitly specified and execute successfully from a separate working directory. Stream assertions now reject output on both streams.

The existing command module remains the public filesystem adapter; argv and stream handling live in the binary, and core compiler/runtime semantics remain separate. API changes: RunResult.value is an immutable NativeResult (inspect with as_value()); RunError.value() returns Result so output errors propagate.

Finding 8 remains an explicit later-assurance gap: there is still no evaluator-to-CLI EvaluationOutcome::Refused case. Validation refusal is covered and is not claimed as a substitute.

Full suite passed at 1971d36: 294 tests + 3 compile-fail doctests, 4 existing ignored. After the final package accessor/stream assertion/immutable wrapper, 14 focused tests, strict all-targets/all-features Clippy, fmt and warnings-denied rustdoc passed. Minimal build passed before those final additions. QUOIN all-review, actual code/rust review and scoped Quire validation are recorded in place. No hosted CI ran.

The branch includes #16/#17 corrections through merge commits. #19 and later command extensions still need adaptation; #27/#29 track that cross-cutting work, and #28 tracks the matrix status defect.

PR #19, opened 2026-09-09
LC05: export exact native packages from source-only requests

kreneskyp:
`quire-spec compile <request-file>` now exports the exact existing native linked package artifact from a source-only `native-compile/1` request. It shares run's bounded file intake and static compiler path, then writes the complete package bytes without a wrapper or extra newline. No runtime files are needed. Static failures leave stdout empty and preserve the existing command error codes/provenance.

Two real binary tests cover state and recorded-operation exports, byte identity, actual `NativePackage::read_verified` intake and request/static refusals. The Rust example generator now writes `compile.json`; usage is in `docs/native-standalone.md`.

Validation: full local suite passed 289 tests plus three compile-fail doctests, with four existing assurance tests ignored. After consolidating duplicate fixture imports, strict all-targets/all-features Clippy and all nine command tests passed. Formatting, rustdoc, cached minimal build, both fixture audits, example generation/export and Quire validation passed. No hosted CI ran.

Stacked on #18; review after #14 → #15 → #16 → #17 → #18. Refs #6. FR-027 / Task-026; QUOIN all spec reviews SR-175–182, author code/Rust SR-183 PASS and scoped gap SR-184 CONDITIONAL. Broader Quire/consumer adoption and deferred assurance remain open. Advisor version detection is unavailable; method choices are explicitly author judgment. Stdout I/O errors can leave a partial prefix, so artifact consumers still verify the selected digest.


kreneskyp:
## CHANGES REQUESTED

Reviewed against `agent-a/lc05-standalone-run` (#18) on `agent-a/lc05-compile-command`. Gates run locally on the head branch.

### Gates

```
    292 passed (289 tests + 3 compile-fail doctests); 0 failed; 4 ignored
```

The `with_request`/`request`/`intake` extraction is a genuine improvement over #18's inline `run_bytes` — the findings below are about where it stopped.

### Findings

**1. `src/command.rs:182-201` — the static compiler is now a method on the file-intake type; name and body disagree (high).**
`Intake` is documented and named as bounded file reading (`directory`, `remaining`), and it has gained `Intake::package()`, which runs parse → link → check → package and returns a `NativePackage`. Compilation has no relationship to byte budgeting beyond needing one file. The consequence is already visible: `compile()` at `:237` reads as `intake.models(...)` then `intake.package(...)`, so the whole static pipeline is hidden behind a type whose job is I/O, and a reader looking for "where does compilation happen" finds it under a name that says it does not. Fix: keep `Intake` to `file`/`source`, and put `models`/`package` on a `Compiler` (or free functions taking `&mut Intake`).

**2. `src/main.rs:84-110` — the dispatch chain grew a third string comparison, and the success half of the duplicated block was not extracted (high).**
`main()` is now `if let [command, path] { if command == "compile" {...} if command == "run" {...} }`, falling through to `run()` which checks `matches!(command, "parse" | "format")` and then `if command == "format"`. Four commands, four string comparisons, two arities, three functions, and a usage literal at `:28` edited by hand for the third time. `command_error` was correctly extracted for the failure half, but the two success blocks remain near-identical `match write ... BrokenPipe ... ExitCode` bodies differing only by `write_all(&bytes)` vs `writeln!("{}", result.value)`. Concrete failure: `quire-spec compile a b` (arity 3) matches neither pattern and reports the parse/format usage line. Fix as in #18 finding 4 — parse argv once into a `Command` enum; then adding `compile` is a variant, not three edits.

**3. `src/command.rs:237-245,268-276` — the wire format is selected by passing a magic literal into a generic function (high).**
`request(bytes, "native-compile/1")` and `request(bytes, "native-run/1")`. There is now no single place listing the formats this binary admits: `"native-run/1"`, `"native-compile/1"`, `"native-state-input/1"` (#16 `reading.rs:95`), `"native-rule-model/1"` (#17 `model_source::FORMAT`), `"native-run-result/1"` (#18 `output.rs`), `"native-linked-package/1"`. Only one of the six has a schema in `schemas/`, and only one is a named constant. A typo in either literal here compiles and produces `RunCause::Format` — i.e. "unsupported format" — for a request that was actually well-formed. Fix: one `wire_format` module of consts (or a `WireFormat` enum) that both the reader and the schema directory are derived from.

**4. `src/command.rs:279-287` — `intake(directory, counts: &[usize])` loses which count overflowed (medium).**
The caller passes an anonymous slice — `&[models.len()]` for compile, `&[models.len(), snapshots.len(), invocations.len()]` for run — and every overflow reports the same `Limit("selected files")`. A request rejected for 64 models is indistinguishable from one rejected for 64 snapshots, in a diagnostic whose entire job is telling the author what to fix. The parallel-slice shape also means the compile path silently omits the two counts it does not have rather than declaring them zero. Fix: pass the named counts (a small struct), and name the exceeded category in the error.

**5. `src/main.rs:87-95` — `compile` writes raw package bytes to stdout with no framing, so a partial write is indistinguishable from a short package (medium).**
`write_all(&bytes)` on success; a mid-write `BrokenPipe` returns `ExitCode::SUCCESS` after an arbitrary prefix has already been emitted. The PR body acknowledges this ("consumers still verify the selected digest"), and the design intent — exact bytes, no wrapper, no trailing newline — is right for the artifact. But exit 0 on a truncated artifact is the wrong signal regardless of what the consumer checks: a shell pipeline `quire-spec compile r.json > pkg.json` that hits `ENOSPC` mid-write reports success. `BrokenPipe` specifically may be worth keeping quiet; every other write error, and any short write, should not return `SUCCESS`.

**6. `tests/compile_command.rs:96-110` — the failure table dispatches on a string tag with `_ => unreachable!()` (medium).**
`for (change, expected, code) in [("format", 1, "unknown_wire"), ("runtime", 2, "invalid-request"), ...]` followed by `match change { "format" => ..., "runtime" => ..., "stale" => ..., "syntax" => ..., _ => unreachable!() }`. The tag, the mutation and the two expected values are kept in correspondence by position and by literal. A renamed tag falls into `unreachable!()` — which at least panics — but a tag whose mutation arm is deleted silently tests an unmodified fixture against the expectation of a mutated one. Same smell as #15's test table; an enum per case makes the compiler check it. (`assert_eq!(failure["code"], code)` at `:122` also pins `"invalid-request"`, one of the four undocumented codes from #18 finding 2.)

**7. `tests/compile_command.rs:44-53` — the "no runtime files needed" property is tested by deleting files by name (medium).**
The test removes `snapshot-0.json`, `snapshot-1.json`, `invocation-0.json`, `request.json` (each guarded by `if path.exists()`) and then asserts compile succeeds. That does demonstrate the property, but it is coupled to the fixture generator's current filenames: add `snapshot-2.json` to `standalone_setup::write` and the test keeps passing while no longer proving what it claims. A tempdir containing only the files `compile.json` references would test the requirement directly rather than the fixture's file list.

**8. `src/command.rs:237` — `compile` returns `Vec<u8>` while `run` returns `RunResult` (low).**
Two sibling public entry points with unrelated success types, both erroring into `RunError` — whose `exit_code()` and `value()` are shaped for `run`'s JSON-on-stderr convention and are simply reused. `RunError`/`RunCause` are also still named for `run` now that they serve both. Minor, but the naming will be wrong again with the fourth command.

### Gap analysis

- **Requirements → tests:** FR-027-AC-1..3 all bind. `TC-105` ↦ `tests/compile_command.rs:29` (AC-1, AC-2) and `:84` (AC-2, AC-3). Bare `use ix_trace_rs::trace;`, markers bind.
- **Matrix rows:** the three TM-007 rows added for FR-027 all resolve to a tagged test. No unbacked rows in TM-007 across the whole stack (FR-022..FR-027, 26 rows, all backed).
- **Code → requirements:** `Intake::models`/`Intake::package` are refactoring extractions of FR-026 code, correctly re-headed `FR-026/027`. `wire::CompileRequest` maps to FR-027's inputs. No orphan surface introduced by this PR.
- **Test-visibility change:** `tests/support/runtime_setup.rs` and `standalone_setup.rs` widened `pub(super)` → `pub(crate)` to share fixtures across binaries. Correct fix for the duplication, and it removes a `#[path]`-include copy rather than adding one.
- **Tests:** real end-to-end. `source_only_compilation_emits_exact_bytes_accepted_by_the_existing_reader` asserts byte-for-byte digest identity against the in-process `NativePackage::new` result *and* round-trips the CLI's stdout through `NativePackage::read_verified` — a genuine consumer check, not an assertion about a double. `assert!(!output.stdout.ends_with(b"\n"))` pins the no-trailing-newline contract. No tautologies.
- **Still open across the stack (not blocking this PR):** TM-007 is not referenced from `spec/spec.md` and the root matrix TM-001 carries no FR-022..FR-027 rows (same as TM-005/TM-006 — pre-existing); five of the six declared wire formats have no schema in `schemas/`; and the four ad-hoc command error codes are absent from `docs/native-error-codes.md`.

kreneskyp:
Corrections pushed at 2883eb6 (implementation f0cfe7b, adopting corrected #18).

- Findings 1/2: compilation now lives in `command::compilation`; bounded intake owns file/source reads. One typed command parser and output dispatcher handle run/compile/syntax, including command-specific arity errors before file I/O.
- Finding 3: one `WireFormat` catalog owns the six selectors; each request type owns its admitted format, with reader aliases and result serialization sharing the catalog. Complete JSON schema coverage remains follow-up work; this does not claim all six schemas are delivered.
- Findings 4/6/7: named file counts preserve category/requested/remaining/maximum; adverse compile cases use an exhaustive enum; the source-only test builds a fresh directory containing only request-selected sources.
- Finding 5: the real binary redirected to `/dev/full` exits 2. The ENOSPC-success claim did not reproduce; the documented quiet BrokenPipe policy remains. Exact artifact bytes and verified consumer reread still pass.
- Finding 8: shared `RunError` naming remains a compatibility follow-up under #27. Catalog/matrix status reconciliation remains tracked by #28; command consolidation is tracked by #29.

Local validation: 300 tests + 3 compile-fail doctests passed, 4 existing assurance tests ignored; all 17 focused command tests passed. Formatting, strict all-targets/all-features Clippy, cached minimal build, warnings-denied rustdoc and scoped Quire validation passed. FR-027 has 3/3 traced criteria and TM-007 has 11/11 backed cases; these counts do not engine-verify matrix status labels. Existing QUOIN all-set, code/Rust and gap review artifacts were updated in place. No hosted CI ran.

PR #20, opened 2026-09-09
LC05: verify and execute selected native package files

kreneskyp:
The local run command can now select a compiled native package by file and exact digest. It verifies the artifact against the supplied source, authored bindings and admitted models before execution. Reader failure preserves its original details and never falls back to compilation; omitting selection retains the existing behavior.

Stacked on #19; review this delta after its base. Implements FR-028 / Task-027 under #6. Four new binary tests cover actual state/operation outcomes, alternate-layout byte identity, forged/stale artifacts and exhausted-budget retries.

Validation: 293 tests plus three doctests pass; four existing assurance tests remain ignored. Strict Clippy, formatting, rustdoc, cached minimal build and Rust fixture audits pass. Generated selected-package examples return the expected 0/1 outcomes. All eight QUOIN specification reviews are recorded in SR-185–192, code/Rust review SR-193 passes, and scoped gap review SR-194 retains broader integration/assurance work. Quire validation passes with existing registry warnings. No hosted CI was dispatched.


kreneskyp:
**MERGEABLE** — gates green, FR-028/TC-106 binds for real, no stubs. Findings are design/traceability debt, none blocking.

## Gates (run on `8db2292`, worktree `wt-b`, `CARGO_TARGET_DIR` isolated)

```
$ cargo fmt --check                                        # clean
    Finished `dev` profile ... in 28.45s                   # no warnings
$ cargo test
   totals: 296 passed; 0 failed; 4 ignored
   tests/standalone.rs: 8 passed (4 new TC-106 tests included)
```
4 ignored = 3 `QUIRE_STATE_CORE` private-packet tests + 1 LC04 activation gate; both are named lanes.

## Findings

1. **Wire error codes are half a stable catalog and half ad-hoc string literals** — `src/command/output.rs:63,68,77,78` emit `"io-error"`, `"invalid-request"`, `"invalid-digest"`, `"invalid-identifier"` as bare `&'static str`, while sibling arms use `Code::X.as_str()` from the `src/diagnostic.rs:50` catalog. Same for the stage column (`"file"`, `"request"`, `"envelope"`, `"intake"`, `"package"`, `"input"`). Failure scenario: a consumer written against `Code` cannot enumerate the command's error space; renaming a literal is invisible to the compiler and silently breaks a downstream matcher. Fix direction: one `CommandCode`/`Stage` enum with `as_str()`, or fold these into `Code`.

2. **`RunCause::SelectedPackage` and `RunCause::Package` collapse to the same wire `stage: "package"`** — `src/command/output.rs:85-95`. A consumer distinguishes source-compilation package failure from selected-artifact failure only by sniffing for `details.file`. That is presence-of-field typing on a wire contract. Give the selected path its own stage token.

3. **`RunCause` (`src/command.rs:41`) and `Code` (`src/diagnostic.rs:50`) are `pub` with no `#[non_exhaustive]`** — this PR adds a variant to `RunCause`, and #23/#25 add three more. Every added variant is a breaking change for any downstream `match`. Add `#[non_exhaustive]` now, before the crate has external consumers.

4. **Request file operands are joined to the request directory with no traversal or absolute-path guard** — `src/command.rs:243` (`self.directory.join(path)`), reachable from `package.file`, `program.source.file`, `models[].source.file`, `snapshots[].file`. `"file": "/etc/passwd"` makes `Path::join` discard the base entirely and read that file; `"../../x"` escapes. SR-191 records "the command is not a sandbox", so this is the documented posture, not a regression — but it is undocumented in FR-026/FR-028 themselves and unasserted by any test. Either state the non-guarantee in the FR or reject non-relative/parent-escaping operands.

5. **`intake()` computes a file-count budget it then throws away** — `src/command.rs:404-414`: `remaining_files` is `checked_sub`'d down and dropped; only `remaining` (bytes) reaches `Intake`. It happens to be correct because all counts are known up front, but the local reads as a live budget and is not one. Either store it in `Intake` or make it a plain `fn check_file_count(counts) -> Result<()>` so the shape matches the intent.

6. **`selected_package` rebuilds `CheckBindings` that `Intake::package` already knows how to build** — `src/command.rs:181-190` duplicates the source-read + `clauses.iter().map(wire::Binding::bind).collect()` block from `src/command.rs:236-241`. Two copies that must stay in sync when the binding shape changes. Extract a `fn bindings(&mut self, program: &wire::Program) -> Result<CheckBindings>`.

7. **`output::error`'s `SelectedPackage` arm inlines two nested match closures inside a `json!`** — `src/command/output.rs:97-108`, with fully-qualified `crate::package::PackagePathSegment::…` paths inside the macro. Presentation logic for `PackagePathSegment`/`PackageCause` belongs in named `fn`s next to `runtime_path`/`diagnostic`, which is the pattern the rest of the file already uses.

8. **Test smell: `selected_package_failures_preserve_reader_details_without_compilation_fallback` (`tests/standalone.rs:132-212`) is one test asserting five mutation families plus five malformed-request shapes**, dispatched by `match mutation { "stale" => …, "forged" => … }` on a `&str` tag with an `unreachable!()` arm. Same string-tag-dispatch shape in `tests/lower_command.rs` later in the stack. A `#[derive]`d enum or five `#[test]`s would name the failing case in the harness output instead of only in the assertion message.

No further findings; the substantive behaviour (no compilation fallback after a reader refusal, fresh budgets per request, alternate-layout digest identity) is genuinely exercised through the real binary.

## Gap analysis

- FR-028-AC-1..4 → TC-106 → four `#[trace("TC-106", "FR-028-AC-N")]` tests in `tests/standalone.rs:61,92,132,214`. `quire coverage --scope .` binds all four; no unbacked FR-028 row.
- Repo-wide (pre-existing, not introduced here, but it makes this PR's new `✅ Tested` marks decorative): `quire coverage` reports `status-column-matches-nothing` for **all seven** matrix documents including `spec/native-workflow/tests.md` — the `functional-coverage` declaration reads a `Status` column, the tables authored a `Coverage Status` column, so *status classification is skipped entirely* and a complete-but-unbacked row can never be detected. Either rename the column or fix `traceability.status.column`.
- Coverage rollup at this head: 289/297 backed. The 8 unbacked are pre-existing (`TC-010` Manual, `FR-011-AC-1..4`, `FR-017-AC-2`). 20 `untracked_symbols`: `NFR-007-M-2..5` tags in `src/package/encoding/tests.rs` and `tests/package_construction_cases/limits.rs` grep fine and bind to nothing — pre-existing.

kreneskyp:
Corrections adopt the fixed #19 parent (implementation 4f15f0f).

- All command codes now come from `Code`; stages are a typed enum. Selected-reader failures use `selected_package`, separately from source-compilation `package` failures. `Code` and `RunCause` retain the parent's `#[non_exhaustive]` markers.
- File-count preflight is a separate check with named groups, including packages. Compilation and verified package reading reuse one binding constructor without changing their prior stage order.
- Selected-package error details use typed serialization and named path/cause helpers. Adverse test cases use an exhaustive enum; malformed request shapes have a separate test.
- FR-028 explicitly inherits the local absolute/parent-relative path contract. The real binary tests those operands and checks the retained selected package digest.

Local validation: 305 tests + 3 compile-fail doctests passed, 4 existing assurance tests ignored. Formatting, strict all-targets/all-features Clippy, cached minimal build, warnings-denied rustdoc and scoped Quire validation passed. FR-028 has 4/4 traced criteria and TM-007 has 12/12 backed cases. Matrix status reconciliation remains tracked by #28; those counts do not verify authored Tested labels. Existing QUOIN all-set, code/Rust and gap reviews were updated in place. No hosted CI ran.

PR #21, opened 2026-09-09
LC05: export existing Boolean projections from standalone source requests

kreneskyp:
Authors can now run `quire-spec lower <native-compile/1 request-file>` to export the existing Boolean executable projection. The command shares compilation's bounded source-only intake, invokes the existing lowering/strict binder and writes exact executable-projection/v1 bytes. A later unsupported clause refuses the whole export with original native package, source and authored clause details.

Stacked on #20; review this delta after its base. Implements FR-029 / Task-028 under #6. The target remains boolean-oracle/v1; this does not widen numeric/object backend support.

Three new real-binary tests compare library bytes, read output with both pinned IR consumers, generate the complete existing backend oracle population, and check later-clause refusals, intake limits and retries. A runnable Boolean example is included.

Local validation: 296 tests and three doctests pass; four existing assurance tests remain ignored. Strict Clippy, formatting, rustdoc, cached build, Rust fixture audits and Quire validation pass. QUOIN all-set reviews SR-195–202, code/Rust review SR-203 PASS and scoped gap review SR-204 are recorded. Broader Quire adoption and activation assurance remain open. Hosted CI remains dispatch-only; none ran.


kreneskyp:
**MERGEABLE** — gates green; the export is real (`lowering::lower` → exact library bytes, verified against both pinned IR readers and the codegen oracle population). Findings are design debt.

## Gates (run on `4f2c20f`)

```
$ cargo fmt --check                                        # clean
    Finished `dev` profile ... in 2.49s                    # no warnings
$ cargo test
   totals: 299 passed; 0 failed; 4 ignored
   tests/lower_command.rs: 3 passed
```

## Findings

1. **Three more wire error codes invented as bare literals** — `src/command/output.rs:96-102`: `"unsupported_projection"`, `"projection_binding"`, `"invalid_projection_correspondence"` are `&'static str`s, while the fourth arm of the same match uses `Code::ResourceExhausted.as_str()`. So one `LoweringCode` maps to a catalogued code and three map to strings that exist nowhere else in the crate. This is the #20 finding compounding: the command's error vocabulary is now split across `Code`, `LoweringCode`, and free-floating literals. Give `LoweringCode` a `code(self) -> Code` (or a `CommandCode`) so the mapping is total and checkable.

2. **`RunCause::Lowering` clones the entire `FormalSource` into the error value** — `src/command.rs:307` (`package.checked().bindings().source.clone()`). The error object then carries up to `REQUEST_BYTES` (1 MiB) of program text purely so `output.rs:110` can call `program.source().locate(value)`. The clone is there to satisfy the borrow checker, not to express ownership. Fix direction: resolve the span to a `LocatedSpan` at construction and store *that*, so the error stays small and the source stays borrowed.

3. **Command dispatch is now string equality tested twice for the same argument** — `src/main.rs:85-91`: `if command == "compile" || command == "lower"` followed immediately by `let export = if command == "compile" { … } else { … }`. Two string comparisons of one operand to decide one branch. This is the seed of the arg-parsing sprawl that #25 grows into a four-site dispatch (see that review). A single `match command.to_str() { Some("compile") => …, Some("lower") => …, … }` fixes it here cheaply.

4. **The usage string is a fourth, unenforced copy of the command grammar** — `src/main.rs:28` lists `<run|compile|lower>` as free text. Nothing ties it to the `if` chain below it; adding `--target` in #25 required editing the same string by hand again. It should be generated from, or at minimum adjacent to, the dispatch table.

5. **`span` is silently `null` when the locate fails** — `src/command/output.rs:110`: `error.source.and_then(|value| program.source().locate(value)).map(span)`. Two distinct conditions ("the failure had no source span" and "the span could not be located in the program") both serialize as `null`, and neither is asserted. If a coordinate bug ever makes `locate` return `None`, the wire looks exactly like a spanless failure and the test at `tests/lower_command.rs:114-120` would not notice (it only reads `span` on a case that has one).

6. **Test smell: `projection_export_reuses_source_request_refusals_and_intake_limits` (`tests/lower_command.rs:181-216`) uses the same `match mutation { "format" => …, "runtime" => …, _ => unreachable!() }` string-tag dispatch as #20's standalone test.** Third occurrence of the shape in this stack. It's a fixture *table* wanting an enum or a struct, not a string switch with an unreachable arm.

7. **`Case::Boolean(bool)` is a boolean flag parameter selecting behaviour** — `tests/support/standalone_setup.rs:24`. `Case::Boolean(true)` vs `Case::Boolean(false)` reads at the call site as "boolean case, true" with no clue that the payload is the *state value* being bound. Same for `Case::Operation(bool)`, where the flag means "should the frame be violated". Name the payload (`Boolean { flag: bool }`) or split the variants.

8. **`compile_with` takes a closure to vary one line** — `src/command.rs:322-336`. It works, but the two call sites differ only in "return the package bytes" vs "lower then return the projection bytes", and the closure is `FnOnce(&NativePackage) -> Result<Vec<u8>>`. This is the right seam; flagging only because #25 then threads a `ProjectionTarget` through the same path and the parameter list starts growing. Watch it.

## Gap analysis

- FR-029-AC-1..3 → TC-107 → `#[trace]` at `tests/lower_command.rs:44,133,180`; `quire coverage` binds all three.
- The tests are genuinely behavioural, not mock-shaped: exported bytes are compared to `lower(&native, …)` library output, read by **both** pinned IR revisions (`ir::BoundPackage` and `quire-contract-ir-backend`), and driven through `codegen::generate_bound_oracles` to a real two-clause population with `syn`-parsed Rust output. That is the opposite of asserting on a stub.
- Same repo-wide caveat as #20: the `Coverage Status` / `Status` column mismatch means the new `✅ Tested` marks in `spec/native-workflow/tests.md` are never classified by the engine.

kreneskyp:
Corrections adopt the fixed #20 parent (implementation d1fcf16).

- `LoweringCode::code()` maps every category to the native `Code` catalog. The three previously ad-hoc spellings are catalogued without renaming them, and incomplete/exit classification uses the shared typed path.
- Lowering failures now retain boxed source metadata and coordinates resolved during failure construction. The old `FormalSource` clone shared an Arc rather than copying 1 MiB of text, but retained that text; the new error retains no program text.
- Absent and invalid coordinates have distinct typed variants and serialized `span_status`; invalid coordinates retain `unmapped_span`. The actual unsupported-clause binary test still checks the exact located source slice. Separate adapter tests cover absence/invalidity and all four lowering codes/exits.
- `lower` is a variant of the shared parsed command, with command-specific arity and shared output handling. Adverse request tests use an exhaustive enum; shared Boolean/operation fixture payloads are now named fields. The existing `compile_with` export seam is retained.

Local validation: full suite at b910d64 passed 311 tests + 3 compile-fail doctests, with 4 existing assurance tests ignored. After boxing metadata at d1fcf16, all 21 library/lower tests passed, as did strict all-targets/all-features Clippy, formatting, cached minimal build, warnings-denied rustdoc and scoped Quire validation. Export bytes still pass both pinned IR readers and actual codegen/syn checks. FR-029 has 3/3 traced criteria; TM-007 has 13/13 backed cases. Matrix status reconciliation remains #28. Existing QUOIN all-set, code/Rust and gap reviews were updated in place. No hosted CI ran.

PR #22, opened 2026-09-09
feat: consume actual Quire extraction in the native compiler

kreneskyp:
Actual Quire Markdown extraction can now feed the native compiler and runtime through the optional `quire-extraction` feature. The consumer calls pinned Quire `extract_clauses`, verifies the selected body against the original document, and reuses the mapped compiler. It retains authored identity, original locations and Quire's unchanged availability, diagnostics and complete clause population on success or refusal.

The join handles Quire's final-CRLF behavior and byte-based closing-fence columns explicitly. Original input is capped before extraction; default native builds exclude the optional producer. Its dependency grants are inventoried. C retains producer changes and existing-repository CLI/wire adoption.

Validation: full local suite 300 tests plus 3 compile-fail doctests; 4 existing assurance tests ignored. Following the final coordinate correction, the 4 integration tests, strict all-targets/all-features Clippy and formatting pass again. Rustdoc, minimal build and Rust fixture audits also passed. QUOIN all-set specification reviews SR-205–212, code/Rust PASS SR-213 and scoped gap CONDITIONAL SR-214 are validated. Quire traces FR-030 5/5, FR-011 4/4 and TM-007 14/14. Broader assurance remains open. Hosted CI remains dispatch-only and was not run.

Stacked on #21; review this PR's incremental diff. Delivers Task-029 under #6, which remains open for broader LC05 work. Implementation `bf8c170`; review `908a89b`.


kreneskyp:
**CHANGES REQUESTED** — the extraction is **real, not a stub** (verified below), and the join code is the best-written module in the stack. But this PR silently deletes the minimal-feature lint and test lanes from CI, and adds four tracking tags that bind to nothing.

## Gates (run on `bf8c170`)

```
    Checking quire-rs v0.46.0 (git ...8b8020e6)
    Finished `dev` profile ... in 15.69s                     # no warnings
$ cargo test --all-features
   totals: 303 passed; 0 failed; 4 ignored
$ cargo test --all-features --test quire_source
   4 passed; 0 failed
```
I also ran the two lanes this PR removes from CI — both still pass locally:
```
$ cargo clippy --all-targets --no-default-features -- -D warnings   # clean
$ cargo test --no-default-features                                  # 33 green suites
```

**Stub check (your explicit ask): this is not a stub.** `src/quire_source.rs:319` calls `quire_rs::semantic::extract_clauses` on the real pinned crate (`rev = 8b8020e6`, quire-rs 0.46.0, present in `Cargo.lock` and compiled by clippy above). `tests/quire_source.rs:37` builds its `SemanticContext` through Quire's own `read_semantic_block` validator rather than an invented decoder, and asserts `program.extraction() == extract_clauses(...)` byte-for-byte. There is no second Markdown parser and no `starts_with`/`split` fence scanning anywhere in the diff — the fence geometry is validated against `Source::line_start` byte indices with `checked_add` throughout (`src/quire_source.rs:203-234`). That part is right.

## Findings

1. **CI's minimal-feature lint and test lanes were replaced, not added to** — `.github/workflows/ci.yml:19-22`. Before: `cargo clippy --locked --all-targets --no-default-features` and `cargo test --locked --no-default-features`. After: both switched to `--all-features`. What remains for the default build is `cargo build --locked --no-default-features` — a *build*, which runs no tests and lints no test targets. Failure scenario: FR-030-AC-5 and FR-031 both promise "minimal builds retain their existing behavior", and after this PR no CI job can observe that. A `#[cfg(feature = "quire-extraction")]` mistake that breaks the default build's tests ships green. Fix: keep both lanes and add the all-features ones — this is four extra lines, not a trade.

2. **Four tracking tags bind to nothing** — `tests/quire_source.rs:82,198,254` add `IT-003-SC-01`, `-SC-02`, `-SC-03`, `-SC-04`. `quire coverage --scope .` reports all four under `untracked_symbols`. The ids *exist* as prose in `spec/integration/IT-003-opaque-extraction-adapter.md:35-41`, but no matrix declares an integration-scenario reference table that mints them, so the tags grep fine and verify nothing. (`IT-004` has the same problem in `tests/fixture_audit.rs` — pre-existing, three more.) Either add the integration-scenario table to `spec/tests.md` or drop the tags; a tag that resolves to no target is worse than no tag, because it reads as coverage.

3. **The Quire version pin is a pair of magic string literals compared with `!=`** — `src/quire_source.rs:136`: `context.module.contract_version != "1.0.0" || context.module.semantic_core != "0.1.0"`. #23 then *constructs* a module with those same two literals in `src/command/extraction.rs:52-53`, so the pin now lives in two files with nothing tying them together. Failure scenario: bump the consumer to contract 1.1.0 in `quire_source.rs`, forget `extraction.rs`, and every extracted-run request refuses with `UnknownProfile` at runtime instead of failing to compile. Fix: `const QUIRE_CONTRACT: &str` / `const QUIRE_SEMANTIC_CORE: &str` in `quire_source`, used by both.

4. **`body_map` is a 140-line function doing five unrelated jobs** — `src/quire_source.rs:156-297`: availability gate, clause lookup, locus validation, fence-coordinate arithmetic, CR-stripping, `Source::read`, segment construction, and `SourceMap::verify`. Nesting reaches four levels inside the `if original.slice(...) != Some(extracted_text) || !matches!(...)` condition at :235-250. The comments at :221 and :251 are load-bearing precisely because the code cannot say it itself. Split into `locate_fence`, `verify_body_bytes`, `build_map`.

5. **`Error` exposes an upstream type (`ClausesOutcome`) through the public API** — `src/quire_source.rs:64,82` and `ExtractedPackage::extraction` at :109. `quire_rs`'s type is now part of this crate's public surface whenever the feature is on, so a quire-rs point release can break this crate's API without any change here. Either re-export it deliberately with a note, or project it into an owned local type.

6. **`preflight` folds three unrelated refusals into one function with three unrelated `Code`s** — `src/quire_source.rs:118-154` returns `ResourceExhausted`, `UnknownProfile`, or `InvalidModelBinding` from one body, and the third arm ORs together three independent mismatches (identity, path, package) into a single message "Quire context differs from the selected source or authored package". The test at `tests/quire_source.rs:271-298` has to assert all three cases produce the same code, so a bug that makes the *path* check vacuous is undetectable. Split the third condition so each mismatch is separately observable.

7. **`Limits` clamps caller values against a second `Limits::default()` built inside the function** — `src/quire_source.rs:124-128`: `let hard = Limits::default(); … limits.source_bytes.min(hard.source_bytes)`. So `Default` is doing double duty as "the default" and "the hard ceiling", and nothing prevents a future `Default` relaxation from silently raising the hard cap. Make the ceilings named `const`s and have `Default` read them.

8. **`Selection` carries three identity concepts as three unrelated fields** — `src/quire_source.rs:17-24`: `binding: ClauseBinding`, `body: SourceIdentity`, `formal: ir::SourceIdentity`. #23's `wire::BodyIdentity` then re-flattens `body` + `formal` into four stringly-typed JSON fields that duplicate `wire::SourceFile`'s quartet. The concept "an identity pair for a derived source" wants one struct used in all three places.

More findings exist in the lower-severity bands (doc-comment/behaviour drift on `Error::extraction`'s "absent when preflight prevented extraction", the `clause.language.clone()` String passed where `&str` would do, `find()` linear scan over an unbounded clause population); categories are: **magic literals**, **god function**, **layering (upstream types in public API)**, **missing abstraction**.

## Gap analysis

- FR-030-AC-1..5 → TC-108 → `#[trace]` at `tests/quire_source.rs:81,198,254,323`. All bound by `quire coverage`.
- FR-011-AC-1..4 were previously **unbacked**; this PR backs them via TC-108 and the coverage engine confirms it (they drop out of `unbacked_rows`). Good.
- **New tests are invisible to the default `cargo test`.** `tests/quire_source.rs:3` is `#![cfg(feature = "quire-extraction")]`, and the feature is not in `default`. Combined with finding 1, the matrix's `✅ Tested` for FR-030 and FR-011 is only true for someone who passes `--all-features` — and after this PR, CI is the only place that does, while nothing checks the other side. State the feature dependency in the matrix rows.
- Same repo-wide caveat as #20/#21: `status-column-matches-nothing` on all seven matrix docs means no `✅ Tested` mark in this repo is engine-classified.
- No code in the diff lacks an owning requirement. Dependency provenance (`docs/dependency-licenses.json`, `docs/dependencies.md`) is recorded as AGENTS.md requires.

kreneskyp:
Corrections adopt the fixed #21 parent (implementation 55649b3).

- Restored the minimal-feature lint/test lanes and retained separate all-feature lanes. The workflow remains `workflow_dispatch` only; none was dispatched.
- Removed all four IT-003-SC tags that resolved only as prose. Actual FR-030/FR-011/TC-108 traces remain, and TM-007 explicitly states their feature requirement.
- Named constants own Quire contract/semantic-core versions and hard byte/line ceilings. Public result/context types are deliberately re-exported as pinned Quire contracts.
- Preflight now returns distinct typed causes with actual/expected context for identity, path, package and both version selectors, plus measured byte/line ceilings. Each foreign field is tested against its exact variant before extraction; no display-message parsing determines the cause.
- Split clause selection, fence-coordinate validation, exact body-byte verification and map construction. The consumer still calls the actual pinned Quire extractor and borrows its language label; no second Markdown parser was introduced.
- Remaining low-severity identity grouping is carried into #23, where the duplicate command wire identity record is introduced.

Local gates at 55649b3: full all-feature suite passed 316 ordinary tests + 3 compile-fail doctests; full minimal suite passed 311 + 3. Both have 4 existing ignored assurance tests. Strict all-targets Clippy passed in both configurations, as did formatting, cached minimal build, warnings-denied rustdoc and scoped Quire validation. Five extraction tests exercise real producer/compiler/runtime behavior. FR-030 5/5, FR-011 4/4 and TM-007 14/14 trace bindings remain; matrix status classification is still tracked by #28. Existing QUOIN all-set, code/Rust and gap reviews were updated in place.

PR #23, opened 2026-09-09
feat: execute selected Markdown clauses in the standalone command

kreneskyp:
The standalone `run` command can now execute a selected clause directly from digest-verified Markdown, using the actual Quire consumer from #22. The optional `program.extraction.body` descriptor assigns the derived native/formal source identities. Results retain the original source, verified byte-map segments and unchanged Quire availability, clauses and diagnostics alongside the existing native runtime observations.

This is an optional clause-only mode with a Quire-validated context. Ordinary native requests are unchanged. Disabled builds reject the extraction field; selected-package execution and compile/lower export reject this mode explicitly before dependent reads. C retains installed-module and existing-system adoption. The existing Rust example generator now produces runnable Markdown healthy, violating and frame-refused cases.

Validation: 304 tests plus 3 compile-fail doctests pass (4 existing assurance tests ignored); the disabled-feature case passes separately. Strict all-targets/all-features Clippy, formatting, minimal build, rustdoc and artifact validation pass. Generated requests also run outside the harness with true, false and frame-refused outcomes. QUOIN all-set reviews SR-215–222, code/Rust PASS SR-223 and scoped gap CONDITIONAL SR-224 are included. Traces: FR-031 4/4, TM-007 15/15. Hosted workflows remain dispatch-only; none ran.

Stacked on #22; review the incremental diff. Delivers Task-030 under #6, which remains open for broader integration and assurance. Implementation: `1359f05`.


kreneskyp:
**CHANGES REQUESTED** — gates green and the Markdown path is genuinely the real Quire consumer from #22 (no second parser here either). The blocking-ish issue is that three distinct refusal conditions are collapsed into one untyped error carrying its only distinguishing information in a display string, with the guard duplicated in two files.

## Gates (run on `1359f05`)

```
$ cargo clippy --all-targets --all-features -- -D warnings   # no warnings
$ cargo test --all-features
   totals: 307 passed; 0 failed; 4 ignored
   totals: 300 passed; 0 failed; 4 ignored
$ cargo clippy --all-targets --no-default-features -- -D warnings   # clean
```

**Markdown/clause parsing check (your explicit ask): clean.** Nothing in this diff parses Markdown. `src/command/extraction.rs:74` delegates to `quire_source::compile`, which delegates to Quire's `extract_clauses`. No `starts_with`, no fence scanning, no `split` pyramid, no match-on-string over document structure. The only string matching added is the three-way `RunCause::ExtractionMode` message below, which is a different problem.

## Findings

1. **`RunCause::ExtractionMode(&'static str)` is a stringly-typed error: three distinct conditions, one wire code, discriminant only in prose** — `src/command.rs:47-49`, raised at `src/command.rs:363` ("extraction is supported by run only"), `:421` ("extraction cannot select a package artifact"), `:426` ("extraction requires exactly one authored clause binding"), and again at `src/command/extraction.rs:38-42`. All four serialize as `code: "unsupported-extraction-mode"` (`src/command/output.rs:62-66`) with `details: Value::Null`. Failure scenario: a caller that sends both `package` and `extraction` and a caller that sends two clause bindings get byte-identical error JSON and cannot tell which input to fix; the only signal is an English `message` field they'd have to substring-match. The tests can't tell either — `tests/extracted_command.rs:209,214,220` assert the same code three times for three different mutations, so a bug that made the package-conflict check unreachable would still pass. Fix: make it an enum (`ExtractionMode::PackageSelected | WrongClauseCount | CompileCommand`) with one code each.

2. **The "exactly one clause binding" guard is implemented twice, with the message duplicated verbatim** — `src/command.rs:425-428` and `src/command/extraction.rs:37-42`. Two copies of one rule that must be kept in sync by hand, and the second is dead in practice because the first already returned. Delete one (keep the `let [binding] = …else` form in `extraction.rs`, which is the one that actually needs the binding).

3. **The Quire version pin is now a second hardcoded copy** — `src/command/extraction.rs:52-53` constructs `{"contract_version":"1.0.0","semantic_core":"0.1.0"}` while `src/quire_source.rs:136` independently *checks* those same two literals. Drift between them is a runtime `UnknownProfile` refusal on every extracted run, with nothing to catch it at compile time. (Raised on #22 as well; it becomes concrete here.)

4. **`wire::BodyIdentity` duplicates the identity quartet already modelled by `wire::SourceFile`** — `src/command/wire.rs:107-113` declares `identity: String, revision: String, document: String, formal_revision: u64`; `wire::SourceFile` carries the same four plus `file`/`digest`. `src/command/extraction.rs:44-51` then hand-copies them into a `SourceIdentity` + `ir::SourceIdentity` pair. Three representations of one concept. Extract a shared `wire::Identity` and reuse it in both records.

5. **`#[cfg(feature = "quire-extraction")]` now appears at ten sites inside production control flow** — `src/command.rs:44,47,50,139,141,224,229,361,417,486,497`, `src/command/wire.rs:85,92,99,105`, `src/command/output.rs:61,67,71`. The public `RunCause` enum's *shape* changes with a feature flag, so a downstream `match` compiles or doesn't depending on how the dependent crate resolved features — and `RunCause` still has no `#[non_exhaustive]` (raised on #20). Concentrating the feature seam in one place (e.g. a single `Extraction` trait object or one `#[cfg]`'d submodule with a no-op counterpart) would leave `command.rs` feature-free.

6. **`run_bytes` has grown into a god function** — `src/command.rs:417-490`: request decode, three-way mode validation, intake budgeting, package preparation, snapshot reading, invocation reading, selection binding, limit overrides, execution, result formatting, extraction attachment. ~70 lines, four responsibilities, and it grows again in #25. The snapshot and invocation loops at :443-465 are the same block twice differing by one type name.

7. **`RunCause::QuireContext(Vec<SemanticFailure>)` puts an upstream type straight on the wire** — `src/command.rs:51-52` and `src/command/output.rs:68-70` (`json!(failures)`). The command's error schema for that stage is whatever quire-rs's `SemanticFailure` happens to serialize as, with no stable code and no version. A quire-rs field rename changes this repo's public error output silently.

8. **Test smells: conditional assertions and one giant test.** `tests/extracted_command.rs:186-266` is a single `#[test]` covering malformed descriptors, mode conflicts, clause-count conflicts, compile/lower rejection, duplicate JSON keys, stale digests, the line ceiling, two runtime limits, and a fresh retry — ~15 independent behaviours in one function, so the harness reports one name for any of them. Separately, `actual_markdown_cases_keep_native_results_and_exact_source_maps` at :66 is a doubly-nested loop (8 cases × ~20 assertions).

More findings in the low band (`attach` mutating a `RunResult` by string-keyed `Value` index rather than a typed field; `outcome()` at `extraction.rs:96` re-serializing four upstream fields by hand); categories: **wrong-shaped types**, **layering**, **test granularity**.

## Gap analysis

- FR-031-AC-1..4 → TC-109 → `#[trace]` at `tests/extracted_command.rs:36,65,141,187`. All bound by `quire coverage`; no unbacked FR-031 row.
- The behaviour is really exercised end-to-end through the compiled binary (`CARGO_BIN_EXE_quire-spec`), including byte-exact original↔body segment comparison against the on-disk `rules.md`. Not mock-shaped.
- **Feature-conditional coverage, unstated in the matrix**: everything except `extraction_request_obeys_the_build_feature` lives in `#[cfg(feature = "quire-extraction")] mod enabled` (`tests/extracted_command.rs:53`). `spec/native-workflow/tests.md` marks FR-031 `✅ Tested` with no note that four of five ACs require `--all-features`. Combined with #22 dropping the minimal CI lanes, say this explicitly in the matrix.
- Same repo-wide caveat: `status-column-matches-nothing` on all seven matrix documents, so no status mark here is engine-checked.
- No code in the diff lacks an owning requirement.

kreneskyp:
Review corrections are ready at b365b30 (implementation 0d3d294), incorporating the corrected #22 parent without rewriting branch history.

- Three extraction mode refusals now have distinct enum variants and catalogued wire codes. Independent binary tests require each exact code and details before deliberately missing dependent files are read; clause-count details include the actual count. One guard retains the selected binding.
- Original/body records share one closed wire identity and native/formal identity pair. Version constants have one owner. Quire context failures retain producer diagnostics with explicit selected contract/core versions.
- Preparation and optional extraction live behind the private compilation seam. Public error enums are non-exhaustive; necessary feature-gated declarations remain, without adding a trait object. Snapshot/invocation intake shares one helper.
- Typed extraction provenance is constructed before immutable result exposure. Schema negative controls reject missing provenance/map fields. Tests are split by refusal boundary, and TM-007 states the feature-specific evidence.

Local checks at 0d3d294: all features 327 ordinary tests + 3 compile-fail doctests; minimal features 312 + 3; four existing assurance tests ignored in each. Strict all-targets Clippy passes in both configurations, as do fmt, cached minimal build, warnings-denied rustdoc and scoped Quire validation. The controlled context-adapter test uses actual Quire validator failures; it is not claimed as an end-to-end reachable context-construction failure. CI remains manual-dispatch only and was not run remotely.

The owner-selected all spec review, code/Rust review and gap review were updated in place. Matrix trace counts remain distinct from execution evidence; issue #28 still owns the status-column mismatch. Broader adoption and backend assurance remain separately tracked and do not gate this engineering correction.

PR #24, opened 2026-09-09
feat: execute the concrete ConfigVersion model and update workflow

kreneskyp:
The native workflow now includes the concrete ConfigVersion model: bounded versionNumber, optional identity-bearing parent, config_history population and a version-only attemptUpdate frame. A Rust example generator emits actual native/Markdown, model, snapshot, invocation and command files for 13 independently expected healthy, violating, refused and incomplete cases.

The real binary exercises those inputs, actual Quire extraction and selected-package replay. This realizes the named native example; numeric/object/graph generated-backend parity remains open under IT-002.

Stacked on #23; review after #14–23. Implements FR-032 / Task-031 under #6.

Validation: 308 ordinary tests and 3 compile-fail doctests pass; 4 existing assurance tests remain ignored. The 3 native example tests also pass without default features. Formatting, strict Clippy, minimal build, rustdoc and Quire artifact validation pass. Fresh generated native and Markdown cases return true, false and frame refusal outside the harness.

QUOIN all-set reviews: SR-225–232; actual code/Rust review PASS: SR-233; scoped gap review CONDITIONAL for broader integration/assurance: SR-234. No dependency or workflow changes; all checks ran locally with serial Cargo phases.


kreneskyp:
**CHANGES REQUESTED** — gates green and the 13-case ConfigVersion workflow really runs through the compiled binary. But two of the three new tests wrap their substantive assertions in an `if` on a field's presence, so they pass vacuously if that field ever disappears; and the fixture generator is a 114-line god function with five parallel match arms that must be hand-synced.

## Gates (run on `8564909`)

```
$ cargo clippy --all-targets --all-features -- -D warnings   # no warnings
$ cargo test --all-features
   totals: 311 passed; 0 failed; 4 ignored
   totals: 303 passed; 0 failed; 4 ignored
```

## Findings

1. **Two tests put their real assertions behind a presence check, so they can pass while asserting nothing** — `tests/config_version.rs:162` (`if result["source"].is_object() { … }`) and `tests/config_version.rs:201` (`if markdown["extraction"].is_object() { … }`). Failure scenario: FR-032-AC-4 is discharged entirely by the block at :201-225 — original↔body digest correspondence, distinct source identities, distinct package digests. Delete the `extraction` attachment from the run result (the #23 feature this test exists to protect) and `markdown["extraction"]` becomes `Null`, `is_object()` is false, the block is skipped, and the test still passes green. The `if` needs to be an `assert!` with the loop-invariant stated, or the cases that legitimately have no `source`/`extraction` need to be an explicit list.

2. **`has_code` accepts the code in any of three different JSON locations** — `tests/config_version.rs:34-40`: top-level `code`, or `diagnostic.code`, or any element of `diagnostics[]`. So `assert!(has_code(&result, "frame_violation"))` cannot distinguish "the run was refused with this code" from "this code appeared somewhere in a diagnostics list on an otherwise successful run". Failure scenario: a regression that demotes `frame_violation` from a refusal to an advisory diagnostic leaves this assertion green. Assert the shape the AC actually names.

3. **`fixtures::write` is a god function** — `examples/config-version/fixtures.rs:251-364`, 114 lines: model emit, native-source string assembly, program JSON, snapshot file loop, invocation construction, request JSON, compile JSON, Markdown assembly, markdown-run JSON. Nesting to four levels, eight sequential responsibilities, and three separate ad-hoc `if case.operation() { … } else { … }` splits (:255, :276, :290) that each re-derive the same fact. Split into `write_model`, `write_program`, `write_inputs`, `write_requests`.

4. **The `Case` enum's data is spread across five parallel match arms plus a hand-maintained array** — `examples/config-version/fixtures.rs:41` (`CASES: [Case; 13]`, manually listing every variant), `:59` `id()`, `:77` `operation()`, `:84` `clause()`, `:155-170` the snapshot mutation match, and `tests/config_version.rs:131-145`'s 13-row expectation table. Six places to update for one new case, and only two of them (the matches) are compiler-enforced — `CASES` and the test table are not. Failure scenario: add `Case::Foo`, forget `CASES`, and it is silently never generated or exercised; the guard at `tests/config_version.rs:177-180` compares against `CASES`, not against the enum, so it cannot catch this. Fix: derive the case list (a `const` built from the variants, or a `strum`-style exhaustive slice asserted against a match) and fold `id`/`operation`/`clause` into one `fn spec(self) -> CaseSpec` returning a struct.

5. **Repeated tuple shapes that want a struct** — `clause() -> (&'static str, &'static str, &'static str)` at :84 (three same-typed positional fields: requirement name, clause id, expression — trivially swappable at the call site, `:254`); `(keyword, context, point)` at :255; `rows: Vec<(&str, i32, Option<&str>)>` at :154 mutated by index-and-field (`rows[0].1 = 3`, `rows[1].2 = Some("missing")` at :156-161). Any of these positional mixups compiles clean and produces a wrong fixture that the expectation table then blesses.

6. **`expect()` on seven paths in a first-party generator** — `examples/config-version/fixtures.rs:96,117,126,128,148,242,312`. The doc comment at :250 says "static fixture errors are defects", which is a defensible policy, but #25 then goes and converts the *sibling* generator's filesystem panics to propagated `io::Error` because a panicking generator was judged wrong there. Same binary, two policies. Pick one.

7. **The native source and the Markdown wrapper are assembled by `format!` string concatenation** — `:268` builds a full native program from a 5-line format string; `:348` builds the Markdown and then applies `.replace('\n', "\r\n")` to the whole document, which also rewrites the newlines inside the already-indented body. It happens to be correct because every newline in scope should become CRLF, but the transformation is global where the intent is local, and a future `\n` in a non-line-ending position (an escape inside a string literal in the expression, say) would be silently corrupted. Assemble line-wise, or emit CRLF directly.

8. **FR-032-AC-4's coverage is feature-gated and the matrix doesn't say so** — `tests/config_version.rs:187` carries `#[cfg(feature = "quire-extraction")]`, so under the default build the only test tagged `FR-032-AC-4` that runs is `exported_config_version_packages_reconstruct_the_same_native_outcomes` (:230), which covers the package-replay half and not the Markdown half. `spec/native-workflow/tests.md` marks it `✅ Tested` unqualified.

More findings in the low band (`snapshot()`'s guarded arms at :162-163 listing `Changed`/`ForbiddenParent` in two arms of one match; `if matches!(case, Case::Absent) {"root"} else {"child"}` inlined inside a `json!` at :322); categories: **primitive obsession**, **missing abstraction**.

## Gap analysis

- FR-032-AC-1..4 → TC-110 → `#[trace]` at `tests/config_version.rs:43,124,189,230`. `quire coverage` binds all four; no unbacked FR-032 row.
- Real-binary execution, real model frontend (`model_source::read(...).admit(...)`), real snapshot/invocation artifacts. Not stubs.
- Reverse gap: `examples/config-version/model.json` (31 lines, the concrete ConfigVersion model) is checked-in data with no owning requirement statement beyond FR-032's prose. Given AGENTS.md's provenance rule, worth a line in FR-032 saying it is authored fixture data, not a copied artifact.
- Same repo-wide caveat as #20–#23: `status-column-matches-nothing` across all seven matrix documents.

kreneskyp:
Review corrections are ready at 2883c98 (implementation 53431cb), with corrected #23 incorporated.

- Runtime cases now require source/report provenance. Markdown runtime cases also require extraction provenance. Missing-model compilation is the explicit exception and requires the corresponding error provenance. No acceptance check is guarded by the presence of the field it verifies.
- Independent exhaustive expectations require the actual exit, output stream, status, stage and diagnostic location. Validation diagnostics may locate the same refusal at several input paths; a successful/advisory result cannot satisfy these checks.
- One catalog generates Case, CASES and CaseSpec. Named clauses, object rows and current/update inputs replace positional mutations and repeated case matches. Adding a case requires extending the independent test oracle.
- The Rust generator separates model, program, runtime-input, request and Markdown emission. Model/runtime construction and I/O errors propagate; fixed identifier/bounded-arena assertions remain programmer invariants. Markdown uses explicit CRLF line joins. FR-032 owns the newly authored model fixture, and TM-007 states feature gating.

Local checks at 53431cb: all features 332 ordinary tests + 3 compile-fail doctests; minimal features 316 + 3; four existing assurance tests ignored in each. Strict all-targets Clippy passes in both configurations, plus fmt, cached minimal binary/example build, warnings-denied rustdoc and scoped Quire validation. The actual generator succeeded; missing arguments and an existing file as the output directory returned exit 2. Fresh native healthy/violating/frame-refused cases returned the expected true/false/absent-truth outcomes.

The all-set spec review, actual code/Rust review and gap review were updated in place at PR readiness. Matrix status reconciliation remains issue #28, separate from these measured local runs. No dependencies or hosted workflow changed; CI remains dispatch-only and was not run remotely.

PR #25, opened 2026-09-09
feat: export bounded integer IR with explicit lowering targets

kreneskyp:
The compiler can now export bounded integer expressions through the existing strict IR binder with `lower <request> --target integer-ir/v1`. It preserves actual scalar bounds, authored clauses, source locations and direct state/input correspondence. The existing Boolean target remains the default.

The shared lowerer translates integer literals, arithmetic, comparisons and Boolean operators; unadmitted forms and exhausted limits refuse the complete artifact. Both pinned IR readers accept the output. Runnable integer examples return true/false natively while exporting identical source-only IR. Current codegen explicitly refuses numeric expressions; this PR supplies A's compiler inputs for C's backend work.

The standalone example generator also propagates filesystem failures across native and Markdown output. Invalid directories and later writes now produce an error and exit 2 instead of panicking; partial example files may remain.

Stacked on #24; review after #14–24. Implements FR-033 / Task-032 under #5.

Validation: the unchanged numeric implementation at 5a7e5db passed 314 ordinary tests, 3 compile-fail doctests, minimal checks and rustdoc; 4 assurance tests remain ignored. The final generator correction at 7b5b663 passes all 19 affected command tests, six minimal-feature tests, strict Clippy, formatting and the actual example build. Invalid-directory/late-Markdown failures and a successful fresh retry were executed outside the harness. Quire validation passes. No dependencies or hosted workflows changed.

QUOIN all-set reviews SR-235–242 and actual code/Rust review PASS SR-243 are retained. SR-244 keeps full Plan-008 acceptance open because Task-020 activation assurance is deferred; the owner permits engineering delivery. Numeric/object/graph generated-backend parity remains separate work.


kreneskyp:
**CHANGES REQUESTED** — gates green and the integer wire encoder is well tested (exact JSON asserted per field, read back by both pinned IR revisions). Two real design defects: the operator lowering uses `_` catch-all arms that will silently mis-map a future `BinaryOp` variant, and the CLI is now a hand-rolled arg parser spread over four dispatch sites with the target-name strings duplicated in three of them.

## Gates (run on `7b5b663`)

```
$ cargo fmt --check                                                 # clean
$ cargo clippy --all-targets --all-features -- -D warnings          # no warnings
$ cargo clippy --all-targets --no-default-features -- -D warnings   # no warnings
$ cargo test --all-features
   totals: 318 passed; 0 failed; 4 ignored
   totals: 310 passed; 0 failed; 4 ignored
```

**Integer-boundary check (your explicit ask): no `as` casts were added.** `src/lowering/wire.rs`'s `IntegerType` serializer emits `minimum()`/`maximum()` at their native width, and the literal path at `src/lowering.rs:507` uses `text.parse()` with an explicit `InvalidCorrespondence` refusal on failure — not `as`. Literal range against the declared `[minimum, maximum]` is already enforced upstream at `src/checking/constraints.rs:558-560`, so an out-of-bounds literal cannot reach the wire. The one bare cast in the touched files is pre-existing (finding 5).

## Findings

1. **`binary()`'s inner operator maps end in `_` arms that silently absorb new variants** — `src/lowering.rs:565-616`. Each of the three groups matches `op` a second time and closes with a catch-all: `_ => ir::BooleanOperator::Implication` (:572), `_ => ir::NumericOperator::Remainder` (:588), `_ => ir::ComparisonOperator::GreaterEqual` (:604). Failure scenario: add `BinaryOp::Xor` to `src/syntax.rs` and include it in the outer `BinaryOp::And | Or | Implies` arm — it compiles clean and exports every `xor` as **implication**, producing a semantically wrong IR that both pinned readers happily accept and the backend then executes. Nothing fails; nothing warns. This is precisely the case exhaustive matching exists to prevent. Fix: one flat `match op` from `BinaryOp` to a typed operator (or `impl TryFrom<BinaryOp> for …`), so a new variant is a compile error at the single place that owns the mapping.

2. **`binary()` matches on `op` three times to answer one question** — same site, `src/lowering.rs:557-616`: once for the Boolean-target guard (:563), once to pick the group (:566, :578, :591), once inside each group to pick the operator. Three parallel arm lists over one enum that must stay consistent by hand. Collapsing to a single `match op` yielding `(Group, Operator)` removes both this and finding 1.

3. **`ProjectionTarget` has `name()` but no `FromStr`, so the string↔enum mapping is duplicated in three places** — `src/lowering.rs:44-49` (`name()`), `src/main.rs:94-95` (the parse), `src/main.rs:29` (the usage string). Failure scenario: rename a target in `name()` and the CLI keeps accepting the old spelling while reporting the new one in `details.profile`, and the usage text describes a third thing. `rust-review §0b`: implement the standard trait rather than an inherent ad-hoc parser. `impl FromStr for ProjectionTarget` + `impl Display` collapses all three to one.

4. **`main()` is a 70-line hand-rolled argument parser with four dispatch sites for one grammar** — `src/main.rs:83-152`: a `match arguments.as_slice()` with three guard arms (:86, :89, :92), then a separate `if let [command, path]` for `run` (:121), then a fallback into `run()` which does its *own* `matches!(command, "parse" | "format")` check (:41), with the usage string in a fifth place (:29). Concrete consequence today: `quire-spec compile req.json --target integer-ir/v1` falls through every arm into `run()` and reports **"command must be parse or format"** — a message about a command the user did not name. One table (command → arity → handler) fixes the dispatch and the diagnostics together.

5. **Bare `as` cast plus unchecked `+ 1` at a file-read boundary, next door to the correct idiom** — `src/main.rs:50`: `file.take(limits.source_bytes as u64 + 1)`. `src/command.rs:158-162` does the same job correctly with `u64::try_from(limit).map_err(io::Error::other)?` and `.saturating_add(1)`. Pre-existing, but this PR is the one editing `main.rs`'s I/O path and the inconsistency with the repo's own established idiom is the finding.

6. **`ProjectionTarget::admits` is a boolean formula where a match belongs** — `src/lowering.rs:51-55`: `matches!(ty, Boolean) || (self == Self::IntegerIrV1 && matches!(ty, Integer{..}))`. Reads as "Boolean is always allowed, plus a special case", which happens to be true but states the *policy* nowhere. A `match (self, ty)` makes each target's admitted domain explicit and exhaustive over `ProjectionTarget`, which matters as soon as a third target lands.

7. **`ProjectionTarget` is a new `pub` enum with no `#[non_exhaustive]`** — `src/lowering.rs:33`. Same class as `RunCause`/`Code` (raised on #20); this is a fresh one, so it costs nothing to add now. `RunCause::Lowering` also gains a fifth field (`target`) in this PR — the variant is now a 4-field struct variant reachable from a `pub enum` with no stability marker.

8. **`ValueType`'s serializer defers an unadmitted-type refusal to serialization time** — `src/lowering/wire.rs:76-81` returns `serde::ser::Error::custom("unadmitted primitive wire type")` for anything not Boolean/Integer. `admits()` should already have refused, so this is defence in depth, but a bug that got past `admits()` surfaces as a generic serde error mapped through `LoweringCode` rather than a located `Unsupported` diagnostic with a clause and span. The `_ =>` at `src/lowering/wire.rs:165` has the same shape.

More findings in the low band (`ProjectedRead` doc still says "Boolean parameter" in one place while the type widened; `usage()`-derived `LoweringLimits` construction repeated four times in `tests/integer_lowering.rs:311-330`); categories: **naming/altitude drift**, **test duplication**.

## Gap analysis

- FR-033-AC-1..5 → TC-111 → `#[trace]` at `tests/integer_lowering.rs:68,146,207,290,390` and `tests/lower_command.rs:45,84`. `quire coverage --scope .` binds all five; AC-5 (generator I/O propagation) is backed by a real `NotADirectory`/`IsADirectory` test, not a mock.
- Coverage rollup at the top of the stack: **319/323 backed**. The 4 unbacked are pre-existing and non-symbol by design: `TC-010` (Manual) and `FR-017-AC-2` (Inspection).
- `untracked_symbols` at this head is 24: the 20 pre-existing `NFR-007-M-2..5` tags, plus the 4 `IT-003-SC-0N` tags introduced in #22 that bind to nothing (detailed in that review). `unmatched_tags` holds three `IT-004` tags in `tests/fixture_audit.rs`. None are introduced here, but the stack now ships seven tracking tags that grep and verify nothing.
- Repo-wide, unchanged across the whole stack: `quire coverage` reports `status-column-matches-nothing` for **all seven** matrix documents (`functional-coverage` reads `Status`, the tables authored `Coverage Status`), so status classification is skipped and no `✅ Tested` mark in this repo is engine-verified. Worth one follow-up ticket for the stack rather than a change here.

kreneskyp:
Review corrections are ready at 53dc9ef (implementation b0c02c7), incorporating the corrected #24 parent.

- One exhaustive BinaryOp conversion maps to typed Boolean/numeric/comparison operators. The typed group controls target refusal and IR construction; no fallback operator remains.
- One catalog owns ProjectionTarget variants, names and published list. FromStr and Display share it; the public enum is non-exhaustive. The existing CLI parser owns lower's suffix. Compile/run with that suffix return their own usage before I/O, and non-UTF-8/unknown targets refuse before opening a missing request.
- Target admission is an exhaustive target match. Typed command errors retain the selected profile and metadata-only source context from #21.
- The wire boundary now constructs fallible typed borrowed views before serialization. Unadmitted prepared declarations/nested expressions return located Unsupported errors with clause identity; invalid correspondence is distinct. A private conversion control exercises that boundary without claiming a publicly constructible invalid checked package.
- Existing integer/operator/bounds tests and both pinned readers remain real. Generator I/O propagation is preserved alongside the corrected named fixture payloads. The pre-existing file-size conversion is corrected by the adopted shared CLI parent.

At b0c02c7: 60 focused tests pass; full all-feature suite 342 ordinary tests + 3 compile-fail doctests; full minimal suite 326 + 3. Four existing assurance tests remain ignored in each. Both strict all-targets Clippy lanes, fmt, cached minimal binary/example build, warnings-denied rustdoc and scoped Quire validation pass. Actual generator usage, early output-path and late Markdown-write failures return exit 2; fresh integer cases return true/false and identical projection bytes.

All-set spec, actual code/Rust and gap reviews were updated in place. No dependencies/workflows changed; hosted CI remains dispatch-only and was not run. Matrix status reconciliation (#28), full activation assurance and C's numeric backend qualification remain separately tracked; this PR does not claim their completion.

PR #26, opened 2026-09-10
feat: project state fields and materialize validated pre/post inputs

kreneskyp:
Primitive native self fields can now be exported with `state-scalar-ir/v1`. Separate pre/post reads retain their original declaration, model, source and observation identities. `NativeProjection::inputs` supplies their actual values only from the matching, fully validated native context; cancellation and exhaustion return no partial input set.

The review correction replaces message-based correspondence failures with typed defensive invariants carrying the original read. Public tests pin the real earlier rejection boundaries for invalid observations and missing invocation inputs. Integer support comes from one exhaustive target policy, and field-alias searches consume the existing lowering budget, including collisions.

Implements FR-034 / TC-112 / Task-033. The common command implementation tracked by #29 is included. Remaining work: #28 matrix reconciliation and Task-020 generated activation qualification; graph/nonself receivers and numeric codegen acceptance remain outside this increment. No broader acceptance issue is closed by this PR.

The old parked base is integrated with the current language stack. Claude's code/Rust and selected QUOIN specification finding recheck resolves all four original findings; only nonblocking lows and deferred activation assurance remain. The corrected implementation passes 26 focused tests. The integrated stack passes 576 minimal-feature and 592 all-feature tests, with zero failures and four inherited assurance ignores per lane, including five doctests. Both strict all-target Clippy configurations, formatting, minimal binary/example build, warnings-denied rustdoc, both Rust fixture audits and 432-document scoped Quire validation pass. All checks ran locally with one Cargo job; hosted workflows remain manual-dispatch only.


kreneskyp:
## Review — #26 Project state fields and materialize validated pre/post inputs

Reviewed `main...agent-a/lc04-state-scalar-projection` at `f444d03` (`/code-review`, `/rust-review`, `/gap-analysis`).

**Verdict: CHANGES REQUESTED** — one High, one Medium. The design is sound and the traceability is complete; the findings are about the error surface and one enum guard, not the projection itself.

### Findings

| # | Severity | Location | Finding |
| --- | --- | --- | --- |
| 1 | **High** | `src/lowering/inputs.rs:146,223-307` | `InvalidCorrespondence` is one code for 11 distinct refusals, none of them tested |
| 2 | **Medium** | `src/lowering.rs:543,584` | `!= BooleanOracleV1` is an open-ended negative test over a closed enum |
| 3 | Low | `src/lowering.rs` `field()` | alias-search `loop` has no ceiling and is not charged to `LoweringUsage` |
| 4 | Low | `src/lowering/inputs.rs` | `InputProjectionError` is `pub` with all-`pub` fields and no `#[non_exhaustive]` |

---

**1 — High. The variant set is the API, not the message.**

`Budget::missing` funnels 11 call sites into a single `InputProjectionCode::InvalidCorrespondence`, with the actual discriminant living in `message: &'static str`:

```
selected snapshot unavailable          selected invocation unavailable
selected self has a different model…   selected self object unavailable
validated self field unavailable       captured invocation unavailable
captured parameter unavailable         validated state value unavailable
projected read origin disagrees…       arena index exceeds host width
validated projected value is not primitive
```

These are not one condition. "Captured invocation unavailable" is a *caller* error — they offered a context without the invocation this clause needs, and retrying with a different context fixes it. "Arena index exceeds host width" is a 32-bit host limit. "Validated projected value is not primitive" is an internal invariant break that should never survive validation. A caller that needs to tell them apart has to compare `message` against string literals.

Failure scenario: a caller builds a `ValidatedContext` whose selection is `ObservationSelection::Current` and asks for a clause whose read is a captured `Input`. They get `InvalidCorrespondence` + `"captured invocation unavailable"` — byte-identical in structure to the invariant-break case. If a refactor swapped those two branches, nothing would notice.

Which brings the second half: **`grep -rn InvalidCorrespondence tests/` returns nothing.** All 11 paths are unexecuted by the suite. `ResourceExhausted`, `Cancelled` and `ContextMismatch` are each pinned by `state_scalar_lowering.rs:481,491,518`; the fourth variant has zero assertions.

Ask: split the caller-reachable conditions into their own variants with typed fields (which artifact, which declaration), keep one variant for genuine invariant breaks, and pin at least the caller-reachable ones with tests.

This is the sixth PR in this programme carrying the same pattern (also #16, #17, #18, #19, #22, #25). It is worth a tracking issue rather than six separate fixes — happy to file it if you want.

**2 — Medium. One fact, two spellings.**

`ProjectionTarget::admits` states positively which targets accept integers:

```rust
Self::IntegerIrV1 | Self::StateScalarIrV1 => {
    matches!(ty, ir::ValueType::Boolean | ir::ValueType::Integer { .. })
}
```

Two guards state the same fact negatively:

```rust
ExprKind::Integer(text) if self.target != ProjectionTarget::BooleanOracleV1 => {   // :543
} if self.target != ProjectionTarget::BooleanOracleV1 => {                          // :584 (NumericNegate)
```

Failure scenario: someone adds `TextIrV1`. `admits` forces a decision — the `match` won't compile without an arm. The two `!=` guards silently opt the new target into integer-literal lowering and numeric negation, and a text target exports an integer literal with no panic and no failing test.

Ask: `matches!(self.target, ProjectionTarget::IntegerIrV1 | ProjectionTarget::StateScalarIrV1)`, or a `fn admits_integers(self) -> bool` sitting next to `admits` so both facts live in one place.

**3 — Low. Unbounded alias search.**

```rust
let name = loop {
    let name = ir::SymbolName::new(format!("nativeField{}", self.next_alias))…?;
    self.next_alias += 1;
    if !self.declarations.contains_key(&name) && !self.used.contains_key(&name) { break name; }
};
```

Terminates (the catalog is finite) but has no ceiling and its iterations are not charged against `LoweringUsage`. A model declaring `nativeField0…nativeField9999` makes every field lowering scan 10,000 candidates outside the node budget. §11 wants a stated error at the ceiling. Charging `self.enter(node.span, depth)` per attempt would do it.

Credit where due: the *collision* hazard itself is properly tested — `aliases_do_not_collide_and_direct_reads_keep_state_and_parameter_origins` (`tests/state_scalar_lowering.rs:293`) declares a real value named `nativeField0` and asserts the generated alias moves to `nativeField1`. That is the test I went looking for and expected not to find.

**4 — Low.** `InputProjectionCode` is `#[non_exhaustive]`; `InputProjectionError` is not, and exposes `code`/`message`/`work` as public fields. Adding a field later is a breaking change for consumers.

### What is right

- `usize::try_from(arena_value.index())` at the width boundary — no bare `as`, explicit fallback (§7).
- Work charged for *unselected* reads, with the reason in a comment: a large package cannot hide scan work.
- The `std::ptr::eq` package-identity gate is documented in the `inputs()` doc comment, including the instruction to re-validate an independently reconstructed context.
- `ProjectedReadOrigin` is `#[non_exhaustive]`; the `_ =>` arm in `materialize` maps to an *error*, not a value.
- Splitting `used` (per name) from `reads` (per name+observation) is the right shape for pre/post — one declaration, two reads.

### Gap analysis

Clean. FR-034-AC-1..5 → TC-112, TC-112 row added to `spec/native-lowering/tests.md`, `spec/spec.md` relationship added. AC-5 claims the standalone command exports the new target; `tests/state_scalar_lowering.rs:250` runs the real binary with `--target state-scalar-ir/v1`, so the ✅ is earned rather than asserted.

### Gates — run, not assumed

Isolated worktree at `f444d03`, private `CARGO_TARGET_DIR` (shared target dirs serve stale artifacts across branches). Both CI lanes:

| Gate | Result |
| --- | --- |
| `cargo test --locked --no-default-features` | **334 passed, 0 failed** |
| `cargo test --locked --all-features` | **350 passed, 0 failed** |

Issue #30, opened 2026-09-10
[LC01–LC05] Reconcile native source-profile admission with the ADR-0053 ruling


kreneskyp:
## Outcome

Reconcile the native compiler line with the ADR-0053 source-authority ruling
without changing historical profile meaning or treating a backend as the
language definition.

Parent: [quire-specification #1](https://github.com/agent-ix/quire-specification/issues/1).

## Final disposition

- [x] Contract-IR [PR #72](https://github.com/agent-ix/quire-contract-ir/pull/72) accepted ADR-0053 with native Quire as the sole editable source language and OCL/SysML/FRETish as output mappings.
- [x] Specification [PR #35](https://github.com/agent-ix/quire-specification/pull/35) accepted the Quire/native Quire/`ix:native` charter and composed profile identity hierarchy.
- [x] The temporary `ix:native` / `0-draft` / `state-finite/1-draft` replacement candidate is superseded and unavailable; it is not aliased to historical or composed profiles.
- [x] Historical `ix:native` / `0-draft` / `state-finite/0-draft` source, definition and evidence meaning remains unchanged.
- [x] Compiler LC01–LC05 (#2–#6) and composed implementation #35–#40 are closed; their merged bounded implementation is released as `v0.2.0` at `6eccb05238c22ec15dd41704874ee8d4482ca68f`.
- [x] Bounded-scalar verification #83 and ConfigVersion #84 are closed with explicit profile/backend support separation.

## Preserved non-claims and remaining owners

- Exact rational source/runtime/backend implementation remains FS03 work;
  refusal is not rational support.
- The original minimum-only ConfigVersion integer and recursive
  `parent: ConfigVersion[0..1]` remain unqualified until the model authority
  supplies an explicit finite domain and relationship correspondence. The
  bounded 0..1000 corpus proves only its authored worked example.
- Composed state/query/graph semantic revisions retain FS02/FS03 acceptance;
  merged native implementation does not adopt those drafts by implication.
- Contract-IR #69 retains object/graph/collection Kani parity.
- Agent B retains the entire TL-* library lane. This tracker changes no TL
  ownership or file.

The former replacement-profile checklist is superseded by the accepted composed
hierarchy and closed child implementation tickets above. No additional compiler
PR is created solely to rename that obsolete plan.

kreneskyp:
[...]

Issue #28, opened 2026-09-10
[QSL#28] [ASSURANCE] Repair TestMatrix status-column contract and verify status claims


kreneskyp:
Owner: Agent A for compiler adoption; coordinate any shared catalog correction with Agent C. Parent: #1. Workstreams: #2–#6.

The review batch reproduced `status-column-matches-nothing` for all seven compiler TestMatrix documents: the active declaration selects `Status`, while authored tables use `Coverage Status`. Existing trace binding counts remain useful, but no authored `✅ Tested` status is engine-verified.

- [ ] Identify the authoritative catalog/table contract and correct the mismatch at its owning boundary; do not silently fork a shared catalog in the compiler.
- [ ] Run `quire coverage --scope . --json` and verify that every matrix's declared status column is actually read, including a negative control for a false Tested claim.
- [ ] Correct existing readiness claims that conflate trace binding, status validation and executed tests; retain unmet assurance criteria.
- [ ] Update only the affected spec/review artifacts at PR readiness and record the engine/catalog revision and exact local evidence.

This is an assurance defect to resolve without blocking unrelated proof-of-concept engineering. Keep producer coordination private. New executable checks, if needed, must be Rust; no hosted CI dispatch.

kreneskyp:
Consumer compatibility is now demonstrated in PR #33: https://github.com/agent-ix/quire-spec-language/pull/33 (6d7de6a, stacked on #32). No authored matrix header, row reference or status was changed.

The authoritative contract requires `Coverage Status`; the fix belongs to process PR85 at e6ea515 with ISO PR37 at a60ee12. Both are already authored and still OPEN. A clean existing CLI binary ff638b9 / engine d3bc2ba supports them (CLI PR77 and engine PR410 are merged). The installed CLI 4f6ed024 / engine ca7362d4 rejects the new field, so an ambient run cannot substitute for the explicitly selected stack.

Actual compiler observations at 7619105 with the clean candidate stack: zero missing status columns, zero status lies, 325/329 bound targets, and 392/392 specification documents grammar-clean. Manual TC-010 and inspection FR-017-AC-2 have catalogued no-source exemptions; StR-001's two demonstration targets remain unbacked in the source rollup. These numbers do not establish executed-test or stakeholder-demonstration completion. TC-094's generated activation stays deferred.

[...]

PR #31, opened 2026-09-10
Reject native sequence declarations above 10,000

kreneskyp:
Native model admission previously accepted sequence maximum 10,001 despite the source profile's 10,000 ceiling. It now refuses oversized maxima with `unsupported_construct` before publishing a model, through both direct Rust construction and JSON source-draft admission. The check applies to each collection wrapper, including nested and unused declarations, independently of runtime length and caller work budgets.

Three regression tests cover 21 boundary scenarios at 10,000, 10,001 and `u32::MAX`. The Unicode stress fixture retains 40,000 logical occurrences using two admitted 200-wide levels, preserving its original Unicode and work-exhaustion checks. Integration retains current V1/V2 model-profile handling. Two newer composed-query tests now assert the earlier typed admission refusal with original source/locus, while retaining their real 10,000-element type/proof success cases.

The [external review](https://github.com/agent-ix/quire-spec-language/pull/31#issuecomment-5613404849) is **MERGEABLE**. The reviewed parent stack and final state-scalar correction are integrated. Remaining work: broader source-profile reconciliation under #30 and matrix-tooling adoption under #28.


kreneskyp:
## Review — #31 Reject native sequence declarations above 10,000

Reviewed `agent-a/lc04-state-scalar-projection...agent-a/lc02-sequence-admission-limit` at `8866239` (`/code-review`, `/rust-review`, `/gap-analysis`).

**Verdict: MERGEABLE.** Twelve lines of source, a real refusal, and spec text that says exactly what the refusal does and does not cover. Two Low notes, neither blocking.

### Findings

**1 — Low, question. Per-wrapper ceiling, not aggregate.**

`check_type` bounds each `Collection` wrapper and then recurses into `value.element()`, so `Collection<Collection<T>>` with both maxima at 10,000 declares 10^8 elements and is admitted.

The FR anticipates this — "The sequence ceiling applies to every collection wrapper … It constrains declared maxima independently of actual runtime length and caller-lowered work budgets" — so this reads as deliberate, with the aggregate bounded downstream by artifact byte limits and arena admission rather than here. Confirming that is the intent, and that no caller treats this ceiling as a bound on total declared elements.

**2 — Low. The bound is a private const while every neighbouring bound is a `Limits` field.**

`MAX_SEQUENCE_ITEMS` is `const` in `admission.rs`. Every other bound in this crate is a documented `pub` field on a limits struct, and every one of them is also 10,000:

```
ModelLimits::roles / entries / nodes      LinkingLimits::nodes
CheckingLimits::nodes / proof_values      EvaluationLimits::graph_steps / events
```

The FR is explicit that this one is deliberately *not* caller-tunable ("Raising ModelLimits cannot raise this admission ceiling"), so a const is the right call — the divergence from the neighbouring pattern is justified, not accidental. The remaining gap is visibility: a consumer that wants to pre-validate a model before calling `NativeModel::new` has no way to read the value. `pub const MAX_SEQUENCE_ITEMS` would close that without making it configurable.

### What is right

- The refusal carries `Code::UnsupportedConstruct` and names both the offending maximum and the ceiling, so the message is diagnostic rather than decorative.
- Spec text is unusually careful about scope: it states that the rule covers unused and nested declarations, that it is independent of runtime length, that IR construction already guarantees a positive maximum, that both `NativeModel::new` and `ModelDraft::admit` enforce it, and that earlier failures may pre-empt it and keep their own codes. That paragraph answers the questions a reviewer would otherwise have to ask.
- `docs/native-model-checking.md` adds the matching operator-facing note, including the point that the general IR collection constructor has a wider domain and is not this gate.
- FR-015-AC-2 was amended rather than a new AC bolted on, and the Status section explicitly says the older checks do not qualify the amendment — that is the honest form.

### Gap analysis

FR-015-AC-2 amended → TC-041; TC-065 touched for the bound/cancellation case; `tests/native_model_cases/admission.rs` adds the 10,000-admitted / 10,001-refused boundary pair through both public Rust and source-draft admission, matching the AC wording exactly. `spec/model-linking/tests.md` updated. No status inflation.

### Gates — run, not assumed

Isolated worktree at `8866239`, private `CARGO_TARGET_DIR` (shared target dirs serve stale artifacts across branches). Both CI lanes:

| Gate | Result |
| --- | --- |
| `cargo test --locked --no-default-features` | **337 passed, 0 failed** |
| `cargo test --locked --all-features` | **353 passed, 0 failed** |

Issue #27, opened 2026-09-10
[QSL#27] [LC05] Replace message error discriminants and unify command wire codes


kreneskyp:
Owner: Agent A. Parent: #1. Workstream: #6 (LC05), with #5 lowering diagnostics included.

The review batch found that distinct refusals share `&'static str` error variants and that command output mixes catalogued codes with ad-hoc literals. Consumers must be able to identify a failure without interpreting English prose.

Scope: native runtime intake, model source intake, command request/resource/extraction failures and output codes introduced by PRs #16, #17, #18, #20, #21 and #23. PR #16's correction begins this work; this issue tracks the complete cross-cutting inventory rather than declaring completion from that one fix.

- [ ] Replace message discriminants with explicit Rust variants carrying relevant typed context; derive stage/code/incomplete classification exhaustively from the cause.
- [ ] Put every emitted command error code in one owned catalog and document stable spellings. Decide enum extensibility at the public boundary.
- [ ] Update the affected existing requirements and acceptance tests. Each distinct refusal must be asserted independently so bypassing a guard cannot pass through a later failure with the same generic code.
- [ ] Apply QUOIN specify and the selected all-review set at PR readiness, plus code/rust review; report exact local checks and remaining follow-ups.

First-party implementation and tests stay Rust. Hosted workflows remain dispatch-only; local checks run serially with one Cargo job. No B/C repository changes are authorized by this issue.

Review remediation status: #16/#17/#18 corrections are pushed with exact checks and dispositions on the PRs. Later command/extraction cases are still open.

- [ ] Coordinate with C for a producer-owned Contract IR resource-classification predicate; pinned 690bde7 has no such helper. This follow-up does not block the local corrections or proof-of-concept engineering.

## Approved language-foundation roadmap amendment — 2026-09-10

The owner approved one ecosystem roadmap with native Quire as its language foundation. This scoped amendment supersedes conflicting earlier state-only or blanket research-only scheduling; detailed semantics still require the selected specification/review cycle before implementation.

Existing typed-diagnostic remediation remains the owner for stable phase/reason/wire discrimination as composed families are added. New language slices consume that contract rather than inventing message-parsed codes. Preserve existing PR/review history; this roadmap does not restart completed fixes.

kreneskyp:
PR #32 is ready for independent review: https://github.com/agent-ix/quire-spec-language/pull/32 (head 7619105, implementation 3c6a0e6; stacked on #31).

[...]

PR #32, opened 2026-09-10
Own native runtime wire decoding at type boundaries

kreneskyp:
Direct Serde decoding of runtime drafts and value nodes previously accepted positional arrays even though the artifact reader required objects. Closed decoding now belongs to each public type through private wire records and validated scalar wrappers, so direct and nested callers enforce the same grammar.

Public draft field types and `native-state-input/1` serialization are preserved. Required null results remain distinct from omitted fields, and bare-hex digest fields use the byte codec. Three public-API tests cover eleven record types, all ten value variants, raw duplicate keys, missing/unknown fields, positional arrays and scalar boundaries.

The [external review](https://github.com/agent-ix/quire-spec-language/pull/32#issuecomment-5613405024) is **MERGEABLE**, and the reviewed parent stack is integrated. The companion structural schema is in #34. Remaining work: producer-owned resource classification under #27 and broader source-profile reconciliation under #30.


kreneskyp:
## Review — #32 Own native runtime wire decoding at type boundaries

Reviewed `agent-a/lc02-sequence-admission-limit...agent-a/lc05-runtime-owned-wire` at `7619105` (`/code-review`, `/rust-review`, `/gap-analysis`).

**Verdict: MERGEABLE.** Two Low notes only.

This is the change I was hoping to see. The previous shape had the wire grammar spread across a dozen `deserialize_with` function pointers, so the contract for any given record was "whatever these helpers happen to do." Moving it into `wire.rs` as closed `Deserialize` types with `deny_unknown_fields`, and letting the public records reach it through `#[serde(from = "Object<wire::T>")]`, makes the type the schema. The scalar newtypes (`Requirement`, `Symbol`, `Digest`, `Identity`) put the fallible construction in `TryFrom` where it belongs instead of in ad-hoc closures.

### Findings

**1 — Low. One stringly error left in an otherwise typed module.**

```rust
impl TryFrom<String> for Digest {
    type Error = &'static str;
    fn try_from(value: String) -> Result<Self, Self::Error> {
        ByteDigest::from_hex(&value).map(Self)
            .map_err(|_| "expected 64 lowercase hexadecimal digits")
    }
}
```

Every sibling carries a typed error — `Requirement` and `Symbol` return `ir::Diagnostic`, `Identity` returns `Box<InputError>`. This one discards `InvalidDigest` and substitutes a literal. Serde stringifies it either way so nothing observable breaks, but it is the same pattern this PR is removing, sitting inside the PR that removes it. `type Error = InvalidDigest` with a `Display` impl would be consistent.

**2 — Low. The new `LowerHex` impl ignores formatter flags.**

```rust
impl fmt::LowerHex for ByteDigest {
    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
        for byte in self.0 { write!(f, "{byte:02x}")?; }
```

`{:>70x}` and `{:#x}` are silently ignored because the bytes go straight to the writer rather than through `f.pad`. Standard nit for a hand-written `LowerHex`; harmless at every current call site.

**3 — Note, no action. The wire shape is now written twice, and the tests are what bind it.**

`Serialize` still comes from the public struct's own fields; `Deserialize` comes from the parallel type in `wire.rs`. The `From` impls force *field presence* to agree, but not field *names* — renaming a public field would keep the `From` impl compiling while the emitted key and the accepted key drift apart.

`closed_record` in `tests/runtime_reading_cases/wire.rs:9` is what closes that, and it closes it properly: it round-trips through `to_string` → `from_str`, asserts `object.len() == fields.len()`, and indexes `object[field]` by literal name. A rename fails on the index; an added field fails on the length. Flagging it so the invariant is visible and nobody weakens that helper later thinking it is just a convenience wrapper.

### What is right

- `Absent {}` as an empty struct variant, with the comment explaining that Serde's *unit* variant slips past `deny_unknown_fields` on an internally-tagged enum. That is a real hole and an easy one to leave open.
- `InvocationResult` as an untagged enum so a missing `result` key is an error while a present `null` stays the explicit no-result form — with the reason (Serde's missing-field deserializer turning `Option` into `None`) written down at the definition. `wire.rs` line 179 onward tests both directions, plus `true`, `-1`, `1.5`, `2^32`, `"0"` and `[]`.
- Duplicate-key rejection tested on raw text rather than through `Value`, with the comment saying why — `Value` coalesces before the decoder ever sees it.
- `all_value_variants_own_closed_object_decoding` asserts `kinds.len() == 10`, so an eleventh `ValueNode` variant that the fixture does not exercise fails the test rather than passing silently.
- The digest domain is pinned hard: 63/65 chars, uppercase, non-hex, multibyte, and a `sha256:`-prefixed value are all refused, while `Display` keeps the prefix and `{:x}` does not.
- `reading.rs` dropping the outer `Object<T>` wrapper is correct rather than a loosening — each public body type now applies it internally through its own `from` attribute.

### Gap analysis

FR-024 amended, TC-100 extended, `spec/reviews/runtime-owned-wire/` carries the eight review artifacts. `docs/native-runtime-inputs.md` records the closed-decoding contract. Binding is complete.

### Gates — run, not assumed

Isolated worktree at `7619105`, private `CARGO_TARGET_DIR` (shared target dirs serve stale artifacts across branches). Both CI lanes:

| Gate | Result |
| --- | --- |
| `cargo test --locked --no-default-features` | **340 passed, 0 failed** |
| `cargo test --locked --all-features` | **356 passed, 0 failed** |

PR #33, opened 2026-09-10
Record the verified matrix status consumer stack

kreneskyp:
The recorded CLI/catalog mismatch prevented matrix-status verification even though the compiler used the required `Coverage Status` and `Status` headers. This documentation supplies the tested candidate tool/module configuration and repeatable commands, keeping trace binding, status classification and executed tests distinct.

The recorded candidate-stack check bound 325/329 targets, with the remaining manual, inspection and demonstration cases explained. Negative controls detected all seven deliberately false functional completion claims and distinguished report from strict exit behavior. Historical evidence and the deferred generated-activation criterion are preserved.

The [external review](https://github.com/agent-ix/quire-spec-language/pull/33#issuecomment-5613405203) is **MERGEABLE**, and the reviewed parent stack is integrated. Remaining producer and installed-stack adoption work is tracked under #28. This change edits documentation only.


kreneskyp:
## Review — #33 Record the verified matrix status consumer stack

Reviewed `agent-a/lc05-runtime-owned-wire...agent-a/lc05-matrix-status-evidence` at `6d7de6a` (`/code-review`, `/gap-analysis`). Documentation only — 147 lines added, no source, no tests.

**Verdict: MERGEABLE.**

This closes out the `status-column-matches-nothing` finding from the previous review round, and closes it the right way round. The finding then was that `quire coverage` reported the diagnostic on all seven matrices, which would have meant no `✅ Tested` mark in the repo was engine-verified. The answer here is that the compiler's headers are correct — `Coverage Status` for functional coverage, `Status` for test summaries — and the installed catalog is what is wrong, with the fix owned upstream in process PR85 / ISO PR37. The instruction not to rename headers to work around the installed catalog is the right call and worth having in writing.

### Findings

**1 — Low. Four pinned external revisions with no automated check.**

The runbook pins CLI `ff638b98`, engine `d3bc2baf`, process `e6ea5151` and ISO `a60ee121`. Nothing in this repo verifies they still resolve or still produce the recorded result, so the doc rots silently once those repos move. It does say "checked against compiler `7619105`", which is the important half. Consider adding the verification date alongside, so a reader can judge staleness without resolving four SHAs.

**2 — Low, not a defect. Results are not reproducible from this checkout.**

Reproducing requires `QUIRE_STATUS_BIN` plus two module roots at specific revisions, none of which are vendored here. The doc says so plainly and separates what the check establishes (zero status-column diagnostics, zero status lies, the per-matrix binding counts) from what it does not (execution evidence, test adequacy, StR-001's demonstration criteria). That satisfies AGENTS.md's "do not fabricate successful tests" — recording it as a known limitation rather than a finding.

### What is right

- The negative controls are the reason to trust the rest. Injecting TC-999998 into all seven matrices and confirming the report identifies seven `functional-coverage` lies — and that report mode exits 0 while strict mode exits 1 with byte-identical JSON — tests the detector rather than assuming it. Running them against an isolated copy, with no live matrix or producer module altered, is the right hygiene.
- The rollup is stated as 325/329 *with the four explained*: TC-010 manual, FR-017-AC-2 inspection, two StR-001 demonstration criteria. The explicit "The rollup is not a count of four missing automated tests" heads off exactly the misreading that number invites.
- FR-009-AC-5 / TC-094 is left at `🚧 Generated activation deferred` even though its test symbol exists — not quietly promoted.
- The installed CLI `4f6ed024` rejecting the new `status_column` field is recorded rather than glossed, along with the fact that a passing candidate-stack check does not upgrade that installed setup.

### Gap analysis

No spec or matrix rows change, which is correct for a doc that records an investigation. Compiler #28 stays open for producer merge and installed-stack adoption; SR-275 records the bounded review.

### Gates — run, not assumed

Isolated worktree at `6d7de6a`, private `CARGO_TARGET_DIR` (shared target dirs serve stale artifacts across branches). Both CI lanes:

| Gate | Result |
| --- | --- |
| `cargo test --locked --no-default-features` | **340 passed, 0 failed** |
| `cargo test --locked --all-features` | **356 passed, 0 failed** |

PR #34, opened 2026-09-10
Publish the native runtime input structural schema

kreneskyp:
Runtime artifacts had an exact Rust reader but no standalone structural schema. This change publishes a local Draft 2020-12 schema for `native-state-input/1`, covering snapshots, invocations, all ten value variants and closed required records. Explicit null results and duplicate vector entries remain valid.

Four Rust integration tests validate actual producer output and independently mutate populated records and scalar domains. They also show that schema-valid parsed data can fail selected-byte verification, raw JSON decoding or arena admission. Model validity, population closure and predicate truth remain responsibilities of the existing Rust pipeline.

The [external review](https://github.com/agent-ix/quire-spec-language/pull/34#issuecomment-5613405408) is **MERGEABLE**; it retains a regex-engine portability limitation before stable schema use. The reviewed parent stack is integrated. Remaining work: producer-owned resource classification under #27, broader admission reconciliation under #30 and tooling adoption under #28.


kreneskyp:
## Review — #34 Publish the native runtime input structural schema

Reviewed `agent-a/lc05-matrix-status-evidence...agent-a/lc05-runtime-input-schema` at `4f0b42f` (`/code-review`, `/rust-review`, `/gap-analysis`). 974 lines added, no source changes — a published schema plus 346 lines of tests binding it to real generated documents.

**Verdict: MERGEABLE.** One Medium worth fixing before the `$id` URN is treated as stable, one Low.

### Findings

**1 — Medium. The schema will not load on RE2-based validators, and the file anchors two ways.**

`$defs/symbol` and `$defs/package` end with a negative lookahead:

```json
"pattern": "^[A-Za-z][A-Za-z0-9._-]*(?![\\s\\S])"
"pattern": "^[A-Za-z0-9][A-Za-z0-9._-]*(/[A-Za-z0-9][A-Za-z0-9._-]*)*(?![\\s\\S])"
```

`$defs/digest`, in the same file, uses a plain anchor: `"^[0-9a-f]{64}$"`.

Failure scenario: a Go consumer calls `Compile()` with `santhosh-tekuri/jsonschema`, which is backed by RE2. RE2 has no lookahead, so the call fails on the *schema* before a single document is validated — the artifact is unusable there, not merely lax. Any Rust validator backed by the `regex` crate behaves the same way. The Rust test passes because `jsonschema 0.17.1` uses a lookahead-capable engine.

The lookahead is not gratuitous — it is what makes `"Name\n"` fail. Under Python's `re`, `$` matches before a trailing newline, so the plain form would accept it. Under ECMA-262 (the dialect JSON Schema actually specifies) and under Go, `$` already means end-of-input and the lookahead is unnecessary. So the file currently picks Python-compatibility for two patterns and ECMA-compatibility for the third, and says so nowhere.

There is a latent hazard in the inconsistency too: `digest`'s `$` is only safe because `minLength`/`maxLength: 64` reject the trailing newline first. `schema_checks_kind_pairing_...` proves `"{64 zeros}\n"` is refused, but the length constraint is doing that work, not the anchor. If someone later removes those as redundant with the pattern, the digest pattern starts accepting a trailing newline on any `re`-backed validator.

Ask: pick one end-anchor spelling and apply it to all three, and state the required regex dialect in `$comment` — the file already uses `$comment` to record its other non-obvious limits, so there is a natural place for it.

**2 — Low. `revision.maximum` is not exactly representable.**

`"maximum": 18446744073709551615` — validators that parse schema numbers as `f64` compare against 18446744073709551616 instead. The test asserts `u64::MAX` is *accepted*, which passes under either reading, so nothing here would catch the drift. Worth a `$comment` noting that u64 bounds are advisory in this dialect and the Rust reader is the authority.

### What is right

The test design is the strongest part of this PR, and it is doing genuine work rather than restating the schema.

- `schema_refuses_missing_unknown_mistyped_and_positional_records` walks every object in real generated envelopes and, at *every* nesting level, asserts rejection of an unknown member, a positional array form, each field removed, and each field mistyped — with `records >= 20` so the walk cannot quietly stop finding records. This is what makes the schema a second source of truth that is actually pinned to the first, rather than a document that drifts.
- `kinds` pins all ten `ValueNode` variants by name.
- `schema_pass_does_not_replace_raw_byte_selection_or_arena_admission` proves the contract in the honest direction: duplicate keys, `1.0`, `1e0`, a stale digest and a cyclic arena all pass the schema and are refused by the reader, each with its expected `InputReadStage` *and* the matching typed cause. The `match` on `(stage, cause)` with an explicit `panic!` on mismatch is better than the `matches!` it could have been.
- The schema's own `title`, `description` and `$comment` state plainly that validity establishes none of byte identity, budgets, arena correspondence, model validity, population closure, frames or predicate truth — and that parsed JSON cannot retain duplicate members or integer token spelling. The document does not oversell itself.
- `jsonschema` is a dev-dependency, pinned exactly, `default-features = false`. No runtime cost and no new runtime dependency.
- Round-trip is asserted alongside schema validity: every snapshot and invocation is also `read_verified` and compared on both `draft()` and `bytes()`.

### Gap analysis

FR-024-AC-5 added, TC-099 and TC-100 extended, `spec/native-workflow/tests.md` row added, eight review artifacts under `spec/reviews/runtime-input-schema/`. `docs/native-runtime-inputs.md` records the schema's standing. Binding is complete and the AC wording matches what the tests actually assert.

### Gates — run, not assumed

Isolated worktree at `4f0b42f`, private `CARGO_TARGET_DIR` (shared target dirs serve stale artifacts across branches). Both CI lanes:

| Gate | Result |
| --- | --- |
| `cargo test --locked --no-default-features` | **344 passed, 0 failed** |
| `cargo test --locked --all-features` | **360 passed, 0 failed** |

Issue #40, opened 2026-09-10
[L6/B] Publish a versioned compiled-protocol artifact contract


kreneskyp:
## Outcome

Publish and accept one versioned compiled-protocol artifact contract from the native Quire compiler so parser-free Rust consumers can consume typed protocol declarations without interpreting native source or specification prose.

This is the A/compiler-owned producer seam consumed by [quire-protocol#11 "Integrate the native compiler artifact"](https://github.com/agent-ix/quire-protocol/issues/11), under [EPIC quire-protocol#14](https://github.com/agent-ix/quire-protocol/issues/14). (Corrected 2026-09-11: the earlier reference to quire-protocol#1 is stale; #1 is B's original coordination ticket, and #11 is the actual native-artifact intake ticket.) It belongs under composed package admission/linking in [#35](https://github.com/agent-ix/quire-spec-language/issues/35) and supplies the typed input consumed by the L6 protocol path in [#39](https://github.com/agent-ix/quire-spec-language/issues/39). Quire Specification PR #15 now maintains one coherent prototype draft and freezes exact profile/rule/dependency/source identities at acceptance; its package contract still does not register a wire format or canonicalizer.

The consumer-side numeric requirements and adverse cases are reviewed and accepted in [quire-protocol#2](https://github.com/agent-ix/quire-protocol/pull/2), merge `5ce1e09c64d5d5c29242aa91aa22933b5774db13`. The compiler contract must make those requirements executable rather than leaving a consumer to infer a representation.

## Required specification cycle

- [x] Use QUOIN `/specify` in this repository for the compiler-owned output contract.
- [x] Review the output contract with the selected `/spec-review` set and validate the final artifacts. PR #52 records conditional base/failure-domain reviews; minor residuals remain visible.
- [x] Plan and implement only the reviewed producer surface. Keep native Quire the sole editable formal source.

## Minimum contract

The accepted artifact must carry or bind, at minimum:

- media/schema/type identity and canonical encoding rules;
- producer implementation and revision identity;
- native source revision/digest, edition, profile, rule identities, and exact source references/spans;
- dependency and required/optional feature identities;
[...]

PR #69, opened 2026-09-11
Prove producer and protocol-result digest domains stay separate

kreneskyp:
## What this changes

Three TC-121 controls in `tests/protocol_artifact.rs` prove the digest-domain
separation that FR-042-AC-3 requires but no existing test named: producer-owned
source/model/config digest domains stay separate from the consumer's RFC
8785/JCS protocol-result domain, and admitted producer references are copied
byte exactly rather than recanonicalized.

No production change was needed. Each refusal already carried its own typed
cause; the gap was that nothing exercised them across the two domains.

## The two directions proved

**(a) A producer digest offered where a compiled-artifact digest belongs
refuses.** `a_producer_source_model_or_config_digest_is_never_reinterpreted_as_the_compiled_artifact_seal`
takes each producer-owned domain the artifact retains — raw model package bytes,
the model's canonical IR object, original native source text, the selected
producer binary — and offers it in two consumer-side roles. The compiled-artifact
external seal refuses on `Invalid(Seal)`; the accepted contract reference refuses
independently on `Invalid(Selection)`. A positive control admits the same
references under the real seal, so each refusal is attributable to the
substituted digest alone.

**(b) A protocol-result JCS digest offered where a producer reference belongs
refuses.** `a_protocol_result_jcs_digest_is_never_accepted_in_a_producer_owned_reference`
builds a small RFC 8785 canonical protocol-result document (sorted ASCII member
names, no insignificant whitespace, bare safe integers; its canonical form is
checked against an independently spelled document rather than asserted) and
offers its digest in four producer-owned roles. Each refuses on its own typed
cause, no two alike:

| Producer-owned role | Typed cause |
| --- | --- |
| Selected producer binary reference | `Invalid(Selection)` |
| Model package byte authority | `Invalid(Model)` |
| Model's original producer source | `Invalid(ForeignLocus)` |
| Native source text | `Invalid(Seal)` |

The model case deliberately bypasses the earlier guards rather than stopping at
the first one: the offered dependency bytes are the result document itself, so
the raw-byte seal passes, and every embedded copy of that reference in the
declaration binding authorities is substituted too, so the reader's separate
binding-authority selection check cannot mask the model cause. The refusal is
delivered by the model byte comparison in `models::validate` -- `same_bytes`
against the admitted model's own artifact bytes -- not by the reference-digest
comparison at `src/protocol_artifact/models.rs:144`.

Correction, 2026-09-11, from adversarial review of this PR: an earlier revision
of this body said the refusal came from the model's raw-byte digest remaining the
authority. That is wrong. Deleting `|| artifact.digest != selected.model.digest()`
at models.rs:144 leaves all 27 tests passing, because `intake::selected()` already
pins `ByteDigest::of(bytes) == artifact.digest` and `same_bytes` pins
`bytes == model.artifact_bytes()`, so digest equality follows. That comparison is a
fast path, not a second guard, and it is currently untested -- a refactor could drop
it with the suite staying green. Recorded as a follow-up rather than silently fixed.

**(c) A legitimate producer reference is retained byte exactly.**
`an_admitted_producer_reference_is_retained_byte_exactly_without_recanonicalization`
checks that the retained model reference hashes the producer's exact artifact
bytes and is not the canonical IR projection of the same model, that original
source text and its producer digest survive unparsed, that the compiled-artifact
digest stays in the consumer's own byte domain and collapses onto no producer
reference or the result JCS domain, and that re-encoding the admitted package
reproduces the offered bytes unchanged.

## Gates

Run serially in this worktree, one cargo job, warnings denied, both lanes and
both strict clippy configurations. Every command was run as
`env -u CARGO_TARGET_DIR cargo ...` so the build used this worktree's own
`./target` rather than the shared target directory.

| Command | Result |
| --- | --- |
| `cargo clippy --all-targets -j 1 -- -D warnings` | clean |
| `cargo clippy --all-targets --features quire-extraction -j 1 -- -D warnings` | clean |
| `cargo test -j 1 -- --test-threads=1` | 608 passed, 0 failed, 4 ignored |
| `cargo test -j 1 --features quire-extraction -- --test-threads=1` | 624 passed, 0 failed, 4 ignored |

The four ignored tests are the inherited release-binary producer cases. The
`tests/protocol_artifact.rs` target itself is 27 passed, 0 failed (24 before).

## Scope and residual follow-ups

- Deliberately not built: the deferred coordination with C for a producer-owned
  Contract IR resource-classification predicate. `Correspondence` records remain
  `Unsupported::ProducerCorrespondence`, so the `SelectedDigest` producer-domain
  record is not yet an admitted path and is not exercised here.
- No language semantics were invented and no B or C repository was touched.
- Traceability to the owning native Quire requirements and the consumer handoff
  to quire-protocol#11 is a separate increment and stays unticked.
- Remaining work: #27, #40. Full FR-042 acceptance and the actual source-to-B
  handoff stay open.

https://claude.ai/code/session_01PcWw91irFwpGVuBbfU8HMN


kreneskyp:
Mergeable. All five local gates ran serially in this worktree's own `./target` (every command as `env -u CARGO_TARGET_DIR cargo ...`, one cargo job, warnings denied): `cargo fmt --all -- --check` clean; `cargo clippy --all-targets -j 1 -- -D warnings` clean; `cargo clippy --all-targets --features quire-extraction -j 1 -- -D warnings` clean; `cargo test -j 1 -- --test-threads=1` 608 passed / 0 failed / 4 ignored; `cargo test -j 1 --features quire-extraction -- --test-threads=1` 624 passed / 0 failed / 4 ignored. The change is test-only: no production file is modified, so no existing admission surface moves. The same account authors and reviews here, so this comment is the merge signal rather than an approving review.

Issue #68, opened 2026-09-11
Native admission refusal can report a locus whose source and span name different sources


kreneskyp:
Found while writing the FR-043 temporal evaluation controls under #38. Diagnostic quality only; no admitted artifact is affected.

## Observed

A two-unit package where unit 0 is a `temporal` declaration and unit 1 a `protocol` declaration with `role Service on M::Plain`. `M::Plain` is a bare record with no `objects` entry, so it is not an object type and the role is correctly refused with `Invalid(Type)`.

The refusal's locus came back as `Locus { source: 0, span: Span { start: 1286, end: 1311 } }`. Slicing that span out of source 0 yields `"n { always[0,1] holds(vie"` — meaningless. Slicing the same span out of **source 1** yields exactly `"role Service on M::Plain;"`, the 25 bytes that actually caused the refusal.

So the span is right and the source index is wrong. A reader following the locus lands in an unrelated declaration in an unrelated unit.

## Where to look

`src/protocol_artifact/native/runtime.rs` around line 567: the protocol role loop calls `model_type(...)`, `nominal(...)` and `builder.ty(...)` without setting `work.locus` first. Every other emission leg sets it — `channels.rs` sets it five times, `controls.rs`, `metadata.rs`, `populations.rs`, `values.rs` and `runtime/compensations.rs` all set it per node. `runtime.rs` sets it only once, at line 644.

Two candidate causes, not yet distinguished:

1. The role path inherits a stale `work.locus` from whatever ran last, and that stale value happens to pair an old source index with a fresh span.
2. `DeclLayout::locus` stamps `self.source`, and `self.source` is not the unit that owns the span at this point.

`families.rs` derives its source correctly — `sources.get(proof.unit().index())` per declaration — so the defect is unlikely to be there.

## Suggested fix

Set `work.locus` at the top of the role loop the way `channels::lower` does for each channel, and add a control that refuses a role on a non-object model in a multi-unit package and asserts the locus resolves to the offending text in the named source. A locus that names the wrong source is worse than no locus, because it reads as a confident pointer.

Not blocking #38: the temporal evaluation controls work around it by slicing the span out of every input source until one matches.

kreneskyp:
[...]

Issue #38, opened 2026-09-10
[QSL#38] Implement native temporal compilation and bridge integration


kreneskyp:
## Approved ecosystem roadmap scope

Owner-approved 2026-09-10. Stable plan key: **L5**. One roadmap under [program #36](https://github.com/agent-ix/quire-research/issues/36); owning parent: [quire-spec-language#5](https://github.com/agent-ix/quire-spec-language/issues/5).

Parse, type-check and compile native temporal clauses with shared predicates and captured values through the reviewed contract-IR/TL bridge. Native Quire remains the only editable clause authority.

## Acceptance

- [x] Preserve exact source/model/anchor/capture, native/TL profile, clock, history and observation identities.
- [x] Exercise future and past operators, event-position/sampled/event-time distinctions, open prefixes, closed executions, missing history and silent deadlines against independent expected cases.
- [x] Use existing TL closure behavior only under its actual semantic identity; a differing native interpretation requires reviewed correspondence and distinguishing vectors.

## Dependencies and boundaries

Acceptance prerequisites: [L2](https://github.com/agent-ix/quire-spec-language/issues/35), [quire-specification#5](https://github.com/agent-ix/quire-specification/issues/5).

Related owned interfaces/work: [agent-ix/quire-contract-ir#64](https://github.com/agent-ix/quire-contract-ir/issues/64), [agent-ix/quire-contract-ir#63](https://github.com/agent-ix/quire-contract-ir/issues/63), [agent-ix/tl-syntax#33](https://github.com/agent-ix/tl-syntax/issues/33).

IR #63/#64 currently own bridge specification; their existence is not an implemented bridge. Required reviewed contracts and actual TL capabilities gate dependent implementation. FRETish remains output-only.

## Delivery discipline

Detailed specification artifacts and the selected base-plus-seven QUOIN review set must be accepted before dependent implementation. Ticket creation does not claim that review has occurred. Reviews happen at coherent PR readiness; continue unrelated unblocked engineering. Preserve existing owners, work and historical identities. Do not create repositories or dispatch agents from this ticket. First-party executable production/qualification paths are Rust; all hosted CI is manual-dispatch only and testing runs locally with resource-conscious concurrency.

## Required implementation language and containment

[...]

PR #70, opened 2026-09-11
L5: bounded native temporal evaluation, activation and mapping support

kreneskyp:
Implements the compiler-owned half of L5 for #38: evaluating the temporal body that FR-042 already emits. Parsing, linking, type admission and artifact emission of temporal declarations were delivered earlier; nothing here re-does them.

## Spec cycle

Ran before implementation, as the ticket's gate requires. Reviewed revision `1f0ff95`; artifacts under `spec/reviews/native-temporal/`.

Owner policy is applicable analyses, not all seven. Selected **base + failure-domain + scope-boundary + evidence** — the change is semantics-heavy with a dense failure surface, crosses four agents' ownership boundaries, and carries an NFR whose source requirement prescribes specific catalog verification methods.

| Cycle | Revision | Result |
| --- | --- | --- |
| 1 | `4c1eee8` | 44 findings, 14 high |
| 2 | `87bc83f` | base / failure-domain / evidence PASS; scope-boundary FAIL |
| 3 | `1f0ff95` | scope-boundary's five resolved; one new blocking finding, fixed |

The three findings that changed the work most:

- **The first draft claimed every local rule restated an owned shared rule. It did not.** The pointwise reading of the temporal Boolean connectives is pinned by no owned rule, and the first draft asserted the *opposite* (untimed) reading. FR-043 now carries a per-subsection source table and records the pointwise reading as an open question for agent E, with the selected reading stated so the choice is visible in results. FR-044 was making the same over-claim and gets the same treatment.
- **Three dimensions the shared requirements carry were missing** — the four independent closure/completeness axes, silent-deadline progress, and an authoritative origin against a bare history cutoff. All three are named in #38's own acceptance list.
- **The unsupported-mapping obligation had no owner** while the bridge is absent, so a stub would have satisfied it. It became FR-045: a total function over FR-095's reviewed support table at baseline `4d6230eb`, consulting no backend capability report.

## What landed

`temporal::evaluate` takes an admitted `quire.compiled-protocol/1` package, a declaration index, a caller-supplied observation trace and explicit ceilings. Admission through the reader is the constructor-private evidence; there is no entry point taking a freely constructed wire package.

Evaluation is three-valued. `Unknown` becomes pending with an `unsettled` basis and is never reported as false; a missing required fact inside the decision support becomes an `unavailable` basis with no truth; a reached ceiling becomes a typed stop. Truth, settlement basis, activation, both closure axes, assessment execution and input completeness are independent fields.

The three profiles keep three meanings over identical valuations and identical intervals — the two false-extension profiles evaluate atoms false outside a closed-complete scope while constants keep their authored value, the finite-window profile quantifies only over admitted instants. `temporal::mapping_support` classifies native-to-TL support from the reviewed table alone and emits no TL artifact.

New spec: FR-043 (13 AC), FR-044 (9), FR-045 (5), NFR-008 (5), TC-122–125, TM-008, and `docs/native-temporal-evaluation.md` publishing the counter contract `quire.native.temporal-work/1`.

## Rules the tests exposed

Writing the controls found five rules the contract stated and the code did not implement, plus one bug. These are the reason the test commit touches `src/`:

1. Watermark settlement of a silent deadline — an open scope over complete valuations now settles a silent offset once progress reaches it.
2. Event-position silence — fixed-sample and timestamped progress advance the clock with no business event; an event-position watermark settles nothing.
3. Finite-window decisiveness — the completeness-through-upper-endpoint test applied only to an *empty* window, so an observed-but-incomplete window settled immediately.
4. Progress-regression and completeness-revision contradictions — there was no per-binding retained progress at all, so no contradiction could be detected. Added `temporal::Ledger` and `evaluate_with_progress`.
5. **Bug**: sparse offsets resolved as indices into the ordered position list, so under the finite-window profile an instant at clock distance 2 with two admitted positions resolved to nothing.

Activation reporting was also weaker than FR-044 requires: an unestablished guard was discarded whenever a healthy sibling existed, a failed activation dropped the trigger identity it already knew, and an incomplete capture named its dimension but not which capture. All three now report per trigger.

## Gates

All local, per the owner directive; hosted workflows remain manual-dispatch.

- `cargo fmt --all -- --check` clean
- `cargo clippy --locked --all-targets --no-default-features -- -D warnings` clean
- `cargo clippy --locked --all-targets -- -D warnings` clean
- `cargo test --locked --no-default-features -j 1 -- --test-threads=1` — full suite passes, no failures
- `quire validate --scope . "spec/**/*.md"` clean
- 39 new controls: TC-122 (14), TC-123 (11), TC-124 (7), TC-125 (7), all bound through trace attributes. `quire coverage` reports 0 unbacked rows for this scope and `rust: 658/658/658 bound`, read on the **installed** stack — per `docs/matrix-status.md` a status-classification claim needs the pinned stack, so none is made here.

## Deliberately not done

- **Mutation testing** of NFR-008's exhaustion paths. NFR-040 requires it separately from fault injection; no metric row claims it.
- **Artifact-side checking of the declared clock parameters.** The emitted body carries the clock binding name and the definition entry's identity and revision, but no digest and no declared period, epoch, unit or sequence authority. Those are retained trace premises that change result identity and are not authenticated, so `quire-specification/FR-090`-AC-2 is satisfied for identity and revision only; the rest needs an FR-042 wire extension.
- **The emission half of `quire-specification/FR-095`** — TL formula, valuation request and correspondence record — stays blocked on agent-ix/quire-contract-ir#63, agent-ix/quire-contract-ir#64 and actual TL capability.
- **FR-043-AC-10's basis-substitution half** has no expressible control: the evaluator computes the basis and accepts none, so there is no substitution to submit. The independence half is verified in full. Recorded, not hidden.

## Open questions referred for a ruling

1. **Pointwise temporal Boolean connectives** (agent E). `temporal-common.md` admits "temporal Boolean connectives" without fixing whether they evaluate at each offset the enclosing operator ranges over. Under false extension this decides whether `always[0,1] not holds(p)` is false (pointwise, selected) or true.
2. **Ceiling versus already-settled sibling obligation.** FR-043 says a fact missing outside completed decision support cannot erase a settled truth; NFR-008 says a reached ceiling never emits a Boolean for that obligation. Selected: an obligation already settled over complete exact support keeps its truth, and only the unassessed obligation carries the stop.
3. **FR-095's two axis vocabularies** (IR/TL owners). Its table keys row 1 on "complete execution" and row 2 on "open prefix"; FR-045 normalizes both to surrounding-execution closure, which is a selection for row 2, not a restatement.

Also filed from this work: #68, a diagnostic defect where a native admission refusal can pair a span from one source with another source's index.

Remaining work: #38

https://claude.ai/code/session_01FtMrr5iLjuejhqi2gj47Bd


kreneskyp:
Temporal subsystem review/reconciliation is ready as follow-up PR #74 against this branch.

Verified locally on the PR #70 base plus the spec reconciliation:
- temporal evaluation: 14/14
- activation: 11/11
- limits: 7/7
- mapping classification: 7/7
- `cargo fmt --all -- --check` and strict all-target minimal-feature Clippy pass.

#74 ratifies the merged shared semantics and records an all-lens review. It does **not** claim the A-owned FR-042 `/2` wire contract, F-owned observation authority, or Contract IR/TL correspondence. Those remain explicit gates.

kreneskyp:
## Focused NFR-008 mutation evidence

Executed locally with installed `cargo-mutants 27.1.0`, restricted to `src/temporal/budget.rs` and `tests/composed_temporal_limits.rs`:

- `Work::charge -> Ok(())`: **caught**
- `Work::charge` limit comparison `<= -> >`: **caught**
- `Work::visit -> Ok(())`: **caught**

This demonstrates that the focused limits suite rejects the principal bypasses of charge, ceiling comparison and visit accounting. It is **partial** mutation evidence only: it does not claim an exhaustive mutation campaign across every temporal, activation, retention or result-construction path.

Issue #66, opened 2026-09-11
[L3/L4/L6] Resolve the missing spec cycle on three implemented stages


kreneskyp:
Three epic children were implemented without the five-step spec cycle each of them carries, because no compiler-owned requirement was ever specified for their scope. Found during the tracker reconciliation at `bb30eac` (see the summary on #1).

`spec/functional/` runs FR-001…FR-042. Nothing in it scopes:

| Ticket | Stage | Work rode in on | Missing |
| --- | --- | --- | --- |
| #36 | Reusable predicates and bounded collection queries | FR-040 composed values, and its proof obligations | A scoped requirement, test cases, matrix rows, and a `/spec-review` campaign |
| #37 | The explicit finite graph profile | FR-041 rational model profile, FR-036 traversal | Same |
| #39 | Native ecosystem choreography compilation | FR-035/036, FR-042 | Same |

The behaviour is built and tested in every case — each ticket's acceptance boxes have now been ticked against named test functions, and the evidence is recorded in a comment on each. What is missing is the specification and review record for the stage as a scope. That is fifteen unticked boxes across the three tickets, and it is what stops any of them closing.

## Two ways to resolve, and this needs an owner decision

1. **Run the cycle.** `/specify` a scoped requirement for each of the three surfaces, `/spec-review` with the applicable analyses, bind test cases into a matrix, then tick. Honest, and it produces the artifacts the gate exists to produce. It is also retrospective specification of shipped code, which the epic's own policy warns against presenting as a prior review: "do not imply a retrospective review occurred before the code was written."

2. **Record an explicit waiver.** State on each ticket that FR-040, FR-041 and FR-035/036/042 subsume the stage, name which criteria carry the stage's meaning, and tick on that basis. Cheaper, and defensible if the subsumption is real — but it needs to be written down, not assumed.

Option 2 looks right for #36 and #37, where FR-040 and FR-041 genuinely carry the semantics. #39 is the weaker case: choreography spans channels, delivery, attempts, compensation and recovery across FR-035, FR-036 and FR-042, and no single requirement owns it.

## Do not

Do not create plan bundles for these stages. Epics and tickets serve that purpose in this repository; `plan/` covers the LC01–LC05 era only and is already reconciled at 35/36 tasks done. PR #41 attempted plan reconciliation and was correctly closed.

## Acceptance

- [ ] Owner decision recorded here: run the cycle, or waive with named subsuming requirements, per ticket.
[...]

Issue #39, opened 2026-09-10
[QSL#39] Implement native ecosystem choreography compilation


kreneskyp:
## Approved ecosystem roadmap scope

Owner-approved 2026-09-10. Stable plan key: **L6**. One roadmap under [program #36](https://github.com/agent-ix/quire-research/issues/36); owning parent: [quire-spec-language#5](https://github.com/agent-ix/quire-spec-language/issues/5).

Compile roles, messages, owned choices, parallel branches/joins, bounded repetition, deadlines and compensation into PT01's typed protocol representation. Include internal/external participants, related workflow instances and the selected population-scoped obligations.

## Acceptance

- [x] Resolve role/component/model and payload bindings; diagnose wrong ownership, ambiguous correlation, invalid instance relationships and scope/type errors.
- [x] Compile split shipments, payment retries and related refunds with separate transport delivery, attempt and business-effect identities.
- [x] Preserve causal/visibility assumptions, captured obligations, commit/recovery semantics and source correspondence for PT02 conformance and OB monitoring.

## Dependencies and boundaries

Acceptance prerequisites: [L2](https://github.com/agent-ix/quire-spec-language/issues/35), [L3](https://github.com/agent-ix/quire-spec-language/issues/36), [quire-specification#12](https://github.com/agent-ix/quire-specification/issues/12), [quire-research#47](https://github.com/agent-ix/quire-research/issues/47).

No single universal trace/correlation key is assumed. Global conformance is distinct from realizability. The compiler does not execute business actions or compensation. Timed execution also consumes L5/OB02 interfaces when required.

## Delivery discipline

Detailed specification artifacts and the selected base-plus-seven QUOIN review set must be accepted before dependent implementation. Ticket creation does not claim that review has occurred. Reviews happen at coherent PR readiness; continue unrelated unblocked engineering. Preserve existing owners, work and historical identities. Do not create repositories or dispatch agents from this ticket. First-party executable production/qualification paths are Rust; all hosted CI is manual-dispatch only and testing runs locally with resource-conscious concurrency.

## Required implementation language and containment

Owner directive (2026-09-07): first-party production and possible qualification paths must be **Rust**, including libraries, CLI/tools, generators, validators, canonicalization, result adapters, test assertions and fixture/CI audits. "Glue", prototypes, scripts and tests are not exemptions; a Rust wrapper around non-Rust semantic logic is not remediation.

[...]

PR #71, opened 2026-09-11
Admit observed repeat guards and verify choreography scenarios

kreneskyp:
Observed Boolean guards now use the existing ownership, visibility and bounded
valuation proof when admitting repeats. A repeat that may exit before its first
iteration supplies no progress to an enclosing loop. The emitted guard, finite
bound and static attempt identity retain their authored form.

The branch combines payment retry, split-shipment and related-refund scenarios
through actual parsing, linking, typing, proof, native emission and independent
reading. Tests distinguish transport deliveries, forward attempts, business
effects, and each refund's activation/retry/commit/recovery requirements. No
wire-format or dependency change is introduced.

Validation: minimal full suite 625 passed, all-feature full suite 641 passed,
zero failures and four inherited ignores each. Added boundary controls also pass
in the minimal 23-test choice target. Both strict Clippy configurations,
formatting and pinned specification validation pass.
The PR remains draft while the scoped specification-cycle record under #66 is
reconciled; local implementation checks are complete.

Code/Rust review: reviews/26-09-11-choreography-scenarios-code.md (SR-379),
PASS for this implementation increment after resolving the two boundary-test
gaps. The historical SR-378 describes an earlier revision; its static
attempt-identity defect and the later enclosing-progress/budget defects have
been addressed. Gap-analysis plan completion is unavailable because this stage
uses issues rather than a plan bundle; issue #66 explicitly prohibits creating
replacement plan bundles. The scoped specification-cycle reconciliation remains
part of #66, with the true retrospective sequence retained.

This delivers the three compiler scenarios, not runtime execution or full
ecosystem acceptance. The real B consumer criterion remains outstanding;
FR-042 has 9/10 trace-backed criteria. Hosted CI was not dispatched.

Remaining work: #39, #40, #66.


kreneskyp:
## Coordinator note: the SpecReview artifact committed on this branch is STALE

`reviews/26-09-11-payment-retry-gaps.md`, committed as `57d7a7c`, records **verdict FAIL** on blocking finding FND-001. **That finding was fixed later on this same branch and the verdict is no longer current.** Do not read it as the state of this PR.

FND-001 said the amended FR-042 sentence requiring "one attempt identity per bounded occurrence … carrying its enclosing iteration ordinal", and its FR-042-AC-6 clause, had no implementing code and no verifying test, while `tests/native_payment_retry_emission.rs:906` asserted the opposite cardinality. That was correct when written. The requirement was the thing that was wrong, not the code: the enclosing iteration ordinal belongs to the canonical runtime occurrence, and the emitted artifact is a static graph with no ordinal field. `c85b5ad` retracted the over-reaching sentence, narrowed AC-6, and updated TC-121 to match what is actually implemented and tested.

The gap-analysis agent reported FAIL twice after that fix landed, because it was reporting a verdict it had already finalised rather than re-running against the branch head. Its scoped evidence stands (`native_choice_emission` 22/22, `native_payment_retry_emission` 9/9); its verdict does not. It also never cleared the shared cargo lock for a full suite, and its artifact says so rather than implying a green full run — that part is honest and worth keeping.

Whoever reviews this should either re-run gap analysis against `957bf97` or treat the committed artifact as historical. It is left in place deliberately rather than deleted, since the finding was real when made and the record of it being raised and answered is worth more than a tidy branch.

## What is on the branch

Fourteen commits over `bb30eac`, covering all three legs of #39's acceptance sentence:

- `d62ee1b` — `/specify`. This was the gate, not ceremony: FR-042 lines 119-121 **forbade** dynamic repeat guards outright, so the work was unbuildable as "implement against existing authority". The upstream accepted surface already permits the construct (`choreography-surface.md:89`, worked example at `:278`), so this is compiler catch-up rather than a new capability.
- `64ce9e4` — observed Boolean repeat guards, generalising the existing `decisions`/`formula` partition proof from Choice to Repeat. No wire change; `flow.rs` untouched.
- `dc199fa`/`c3a3224`/`7e89134` — the TC-121 payment-retry scenario.
- `c85b5ad` — the retraction described above.
- `03eb482`/`a3d94b8` — split shipments and related refunds. **No compiler change was needed for either**, justified against the code: `channels.rs` already emits one channel-owned `Delivery` requirement and `controls.rs:425-465` already anchors each `receive` requirement at its own control, so two receives of one send already yield two delivery identities under one effect; `runtime/compensations.rs` already builds registration → activation → attempt → effect, and `scopes/protocol.rs:436-444` already refuses a send/receive/attempt compensation target before family admission.
- `9c46d79`/`f2bfb93`/`957bf97` — a **High** admission-soundness fix found by review: `families.rs:281` discarded the exhausting-branch feasibility bit, so an observed-guard repeat claimed `Progress::Observable` from its body alone even when the guard may be false at the first decision and the body never runs. An enclosing repeat could then take it as its progress source and be admitted while observing nothing. The fix was verified **both ways** — against a temporarily restored HEAD the nested reproducer was admitted (`err()` = `None`); after the fix it refuses with `Invalid(Control)`.

## Deliberately not done

A static check that a send's authored delivery count falls within the channel's `delivery [l,u]` interval. `families.rs:529` validates only `lower <= upper`; `docs/compiled-protocol-v1.md:692` makes concrete delivery correspondence a **consumer** input requirement; FR-042 is silent. Accepted text does not authorise it, so it was reported rather than built. It needs a `/specify` cycle first.

Remaining work: #39, #40

kreneskyp:
Closing as superseded: all code, tests and spec are on main at 04254a3 (via #75), including the High soundness fix (tests/native_choice_emission.rs:1858, :1904; FR-042 at :332-333). Not carried over: the two review files under reviews/26-09-11-*, whose SR-378/379 IDs are now used by other reviews on main. Open follow-up, not built: a static check that a send's delivery count stays within the channel's delivery [l,u] bounds (needs a spec cycle). Branch retained.

PR #72, opened 2026-09-11
Expose signed-64 and rational boundary values in the producer handoff

kreneskyp:
The real producer fixture now exposes full signed-64 integer and exact-rational
domains so B can test boundary values through its public reader. Workflow's
`wide` and `exact` fields are referenced by the Healthy invariant through the
Bounded predicate. The fixture emits integer MIN/MAX and reduced rational pairs
`(i64::MAX, i64::MAX-1)` and `(i64::MIN, i64::MAX)`, reaching both numerator
endpoints and the actual maximum denominator without weakening any limit.

The actual stripped release producer regenerated the artifact, expected selectors
and seal in `/tmp/quire-native-handoff-signed64-endpoints/`. The independent
reader accepts those bytes and their measured SHA-256 matches the producer seal.
Model export indices are **Exact 16, Tally 19, Total 20, Wide 21**; B must
re-derive its frozen selections rather than patch indices or digests.

Validation: focused numeric emit/read test and actual stripped-release producer
test pass; formatting and both strict Clippy configurations pass. The inherited
equivalent pre-correction tree passed full suites 606/622 with zero failures and
four inherited ignores; the endpoint addition was verified with the focused and
release tests above. Code/Rust review SR-380 passes for this producer change.

No production compiler, dependency pin, wire contract or hosted workflow changes.
Actual B consumer acceptance and the issue-based campaign gaps remain separate;
no plan-completion gate is claimed for this stage.

Remaining work: #40.


kreneskyp:
Closing as superseded: 4 of 5 commits are identical patches on main at 04254a3 (via #75); tests/native_numeric_domain_emission.rs matches main. Not carried over: review file reviews/26-09-11-numeric-handoff-code.md, whose SR-380 ID is now used on main. Branch retained.

PR #67, opened 2026-09-11
Compare declared static components and retain requested capabilities (TC-115)

kreneskyp:
## What

TC-115 was the only test case in the program with a written procedure, P1 priority and no implementation. Its three criteria — FR-036-AC-5, AC-6 and AC-8 — were the last open acceptance criteria of FR-036, and the four rows they occupy have been carried as inherited `FND-001` debt by roughly a dozen consecutive gap-analysis reviews.

The compiler could not express two of the three criteria, so this builds them rather than writing a test around the gap.

## Source changes and why they were needed

**`src/linking/composed/subject.rs` (new).** AC-5 needs static meaning compared across configurations through **declared semantic components, not an invented canonical hash**. `StaticSubject` retains exactly the six components the [package contract](https://github.com/agent-ix/quire-specification) names — Sources, Language, Profiles, Models, Native dependencies, Binding requirements — and `differences()` reports which of them moved. No canonical digest, structural hash or JSON fingerprint is derived: the contract holds canonical identity unavailable until its exact domain, version and algorithm are selected and qualified. `BuildProvenance` retains the full compile configuration (package, parser and binding limits plus charged usage) separately, so a resource-only change is visible there and in no static component.

**`src/linking/composed/requests.rs` (new).** AC-6 needs capability disposition visible in the request report; nothing of the kind existed anywhere in the crate. `requests::report` records a typed disposition for every requested clause/capability pair against a caller-declared backend and the supplied assessment inputs. `UnsupportedCapability`, `UnsupportedFamily`, `InapplicableCapability`, `RefusedSubject`, `UnfinishedSubject` and `UnknownSubject` are distinct outcomes; none is dropped or merged; a required non-admitted request makes complete aggregate success unavailable. `admitted_bodies()` exposes only admitted family-check subjects, so an unsupported family body is never represented as checked. Assessment selections and backend support are retained as provenance and never written back into the static subject.

**Derives.** `syntax::Limits` and `scopes::BinderType` gain `Eq`/`PartialEq` so the compile configuration and the declared roles can be compared at all.

## Tests

[...]

PR #77, opened 2026-09-13
FR-045: authenticated /2 mapping classification under the closure ruling

kreneskyp:
Refs #38

## What this does

- **Closure ruling applied.** Following the owner ruling on [quire-contract-ir#64](https://github.com/agent-ix/quire-contract-ir/issues/64#issuecomment-5649214124), surrounding-execution closure picks the TL row. Decision-scope closure is not the selector. FR-045's "open question" hold is gone. `docs/native-temporal-evaluation.md` wrongly said "decision-scope closure"; that is corrected.
- **`mapping_support_v2` now actually authenticates.** FR-050 and the docs already said it did, but the code did not. Before classifying, it now checks that the admitted `/2` temporal binding selects the declaration's own definition. A missing or different binding is refused.
- **Proof of authentication is kept on the result.** `Classification::authenticated()` returns the package digest, declaration, definition identity, namespaced revision and definition artifact digest.
  - A `/1` classification returns `None`.
  - Code outside the crate cannot construct the field.
  - The disposition and retained native subject match `/1` exactly.
- **What classification still does not do.** It still emits no TL formula, valuation request or correspondence record. That emission half stays blocked on quire-contract-ir#63/#64 (spec PRs #67/#68 are still open) and on actual TL capability.
- **Single `Status` column.** Upstream spec-artifacts-process#87 merged `Coverage Status` into `Status`. The installed quire stack enforces this, so `quire validate` failed on all 8 compiler matrices on main. Only the header cells were renamed, and the prose telling authors to keep `Coverage Status` was replaced. No row status changed.

## Spec

- **FR-045:** the ruling section replaces the open question. A new "Authenticated classification" section and FR-045-AC-6 are added, plus a `depends_on` FR-050.
- **TC-125:** group 6 added.
- **TM-008:**
  - Adds the AC-6 row.
  - Counts updated to 41 controls and 33 criteria.
  - Drops the clock-premise caveat "until `/2` lands", which FR-050/TC-138 now own.
- **`spec/spec.md`:** the FR-045 row is updated.
- **No new IDs allocated.** AC-6 sits inside the existing FR-045. Remote branches were scanned before deciding.

## Decisions (made without holding on the owner)

1. **Ticket.** Refs #38, which owns FR-043/044/045. There were no open duplicate PRs; the only open PR, #76, touches only `handoff.rs`.
2. **No clock parameters on the classification.** FR-050 forbids a parameter-map claim here, and classification takes no trace. A bridge reads the clock from the package the digest names.
3. **`/2` supported results still list all eight outstanding premises.** Naming a premise is not discharging it, and there is no bridge.
4. **Defensive refusal paths.** A missing binding, or one that selects a different definition, cannot happen after the strict reader runs. These paths are refused anyway and verified by Inspection, not by an unreachable test.
5. **New test file** `tests/composed_temporal_mapping_v2.rs` (agent E), instead of editing agent A's `tests/compiled_protocol_v2.rs`.
   - Covers all three profiles with bounded future operators, under both closures.
   - The reader selection is rebuilt from the authored sources and the registered catalog.
6. **Status rename covers all 8 matrices, not just TM-008.** The owner's rule is a single `Status` column, and a partial rename would still fail `quire validate`. Historical review records under `reviews/` and `spec/reviews/` were not edited.
7. **Agent A's files are unchanged** (`docs/compiled-protocol-v2.md`, FR-050, `tests/compiled_protocol_v2.rs`, `tests/composed_temporal_mapping.rs`). Two follow-ups for A are noted on #38: FR-050's "formula closure" wording, and adding `AuthenticatedSelection` / `Classification::authenticated()` to the v2 API list.

## Review fixes (/rust-review + /gap-analysis)

8. **`AuthenticatedSelection` is sealed.** Private fields, read-only accessors, a `compile_fail,E0451` doctest (same pattern as `ExactRational`). The revision is kept as the full namespaced `wire::Revision`, not just its value.
9. **`Classification.support` / `.retained` stay public.** Making them private would break agent A's `tests/composed_temporal_mapping.rs`. Only the selection is sealed; FR-045, TM-008, the guide and the type doc now say so, and that a consumer needing the disposition bound to the evidence re-classifies from the named package.
10. **Selection is resolved through the binding.** The definition and artifact digest are read from `binding.definition`, not the declaration. The binding-equals-profile check stays as a re-check of the reader invariant and refuses with `Refusal::Reference`. It is unreachable after the strict reader, so it is an Inspection clause in AC-6, not a test.
11. **`AuthenticatedBinding` is not merged with `AuthenticatedSelection`.** Changing its `String` fields would break A's test. Instead both v2 entry points derive through one `authenticated_selection` helper, so they cannot drift.
12. **Exact refusals in tests.** Out-of-range, unrepresentable and non-temporal (`Flow`) requests each assert the exact located `Error::Refused(..)`. The package digest is compared with `ByteDigest::of(emitted bytes)`, not the admitted package's own digest.
13. **Past-operator and no-operator rows under `/2`** are covered by inspection of the shared path (stated in TC-125 group 6), not by extra fixtures.
14. **Wording.** "The *disposition* is a total function of…" in FR-045, `mapping.rs` and the guide. TM-008's stale pinned-stack caveat is replaced: the installed `quire coverage` reports no `status-column-matches-nothing` diagnostic.
15. **Second review round (at b3b4385):** rust-review found no blockers; gap-analysis PASS. Low findings were fixed. The missing-binding refusal is now listed as inspected in AC-6, TC-125 and TM-008. The `usize::MAX` test no longer claims to reach the binding-index conversion. `docs/matrix-status.md` puts the pinned-stack-only rule in the past tense and keeps `Remaining work: #28` for its negative control (also restored in model-linking). Two lines were rewrapped. Not done: storing `Profile` instead of the identity `String` in `AuthenticatedSelection`. It was optional style, and scope is frozen.
16. **Gate env:** the installed `spec-artifacts-process` module was stale (`d605caa`, which asserts `Coverage Status`). It was upgraded to `375fc2a` (#87) instead of renaming back.

## What changed in behavior

- `mapping_support_v2` used to classify any temporal declaration without looking at the binding. It now resolves the definition through the binding and refuses a missing or mismatched one with `Refusal::Reference`.
- `mapping_support` (`/1`) behaves the same. `select` now takes the subject instead of a budget, and classification is infallible after selection.
- `evaluate_v2` / `evaluate_with_progress_v2`: `authenticate_v2` builds `AuthenticatedBinding` from the shared helper. Output is unchanged for every admitted package.

## Gate

Local only; hosted CI stays manual. The gate result and head SHA are posted as a comment.

Remaining work: #38 (TL emission half, NFR-008 mutation testing).

https://claude.ai/code/session_011HFnqWrGMDuUiaULxxE2Tc


kreneskyp:
Mergeable at head 3047338ecc8d5d54e2a340305ad2152c87bbd3be (tree 5248cd0d598ba5b1b73ea0ada1678403998d4ff1).

Local full gate at that exact head: exit 0, GATE-PASSED. Ran parse/format fixture, cargo fmt --check, clippy -D warnings (no-default-features and all-features), cargo test incl. doctests (both lanes), clean build, fixture-audit self-test and model-bytes, quire validate (spec-artifacts-process 375fc2a). Hosted CI not dispatched (manual-only).

PR-time /rust-review (no blocking findings) and /gap-analysis (PASS) were run; all findings are fixed in 3047338. Not merged.

Refs #38

PR #137, opened 2026-09-17
Drop the Producer 1.2/Filament correspondence adapter (#131)

### LINKED ISSUE #131: [QSL#131] Implement model intake from domain packages
## Goal

QSL admits a Semantic IR 2.0.0 domain package and builds Quire model declarations from it; the model linker resolves ports, channels, components and connections from those declarations.

## Scope

* Intake: artifact bundle -> `quire_rs::semantic::extract_semantic` with the real module manifests -> `agent-ix-extraction-frontend` lift -> IR 2.0.0 read by `agent-ix-semantic-ir` -> model declarations. Admission order follows FR-154.
* Kinds are module-qualified; each construct's embedded declaration supplies its `meaning` id, which binds to the Quire meaning.
* Remove `agent-ix-baseline-producer`, `src/linking/composed/producer.rs`, `AdmittedProducerModel`, `ModelSelection.producer`, and the `state/input.rs` contract revision constants that belong to it; remove the hand-built semantic context in `src/command/extraction.rs`.
* Tests: an end-to-end test over the architecture fixture bundle; `compiled_protocol_v2`, `composed_*`, `protocol_artifact`, `native_*emission` and `support` use domain packages; delete `tests/producer_correspondence.rs`.

## Acceptance criteria

* The end-to-end test turns the architecture bundle ([agent-ix/filament-core-data#173](<https://linear.app/agent-ix/issue/PLAT-751/fcd173-architecture-fixture-bundle-lifts-to-ir-200-goldens>)) into model declarations, and the linker resolves its ports and connections.
* No QSL source matches on a kind name; binding goes through `meaning` ids.
* The removed producer items are absent from the tree.
* Traced tests back the [#130](<https://linear.app/agent-ix/issue/QSL-114/qsl130-specify-model-intake-from-domain-packages>) TCs; full gates pass.

## Scope amendment (ADR-011 FB-13, from [agent-ix/quire-spec-language#250](<https://linear.app/agent-ix/review/adr-013-o-04-correct-adr-011s-s3-minting-boundary-today-list-9982c52ac47c>))

* Wire enum and unit admission behind the `check` stage. `EnumDeclaration::admit`/`admit_member` (`src/value/enumeration.rs`) and `UnitGraph::admit` (`src/value/unit.rs`) mint `NodeKey`s today with no check-stage caller: no `src/` module constructs either type, and `PackageDeclarations::enums` is populated only in `tests/collection_algebra.rs`.
* ADR-011 §6.1's minting rule, ADR-013 O-04 and T-12 stay as written. The T12-B API-surface check correctly fails those four call sites until this wiring lands.
* Evidence: SR-508 (`spec/reviews/nodekey-minting-boundary/base.md`) and ADR-011 §3 FB-13.

Plan merry-snuggling-moler.md, Lane A2.

Depends on [#130](<https://linear.app/agent-ix/issue/QSL-114/qsl130-specify-model-intake-from-domain-packages>)
Depends on [agent-ix/filament-core-data#172](<https://linear.app/agent-ix/issue/PLAT-750/fcd172-semantic-ir-200-module-declared-construct-kinds>)
Depends on [agent-ix/filament-core-data#173](<https://linear.app/agent-ix/issue/PLAT-751/fcd173-architecture-fixture-bundle-lifts-to-ir-200-goldens>)

🤖 Generated with [Claude Code](<https://claude.com/claude-code>)

## Pin-bump follow-ups (2026-09-19)

The intake tail was written against FCD#200's target shape while the FCD pin still names `7dcb2f2c` (FCD main). At the bump:

* Write the duplicate-identity test. Its expected diagnostic is read from filament-core-data's own committed `expected/diagnostics.json` for that fixture — message `` identity `ix://agent-ix/identity-service/FR-001/revision` is already minted by an earlier node ``, locus line 23 col 1, related line 18 col 3. Upstream is where that fact lives; a downstream test does not need to hold a copy of it in the meantime.
* Vendor `tests/fixtures/fcd-199-golden-semantic-ir.json` properly. It is a one-time manual copy of filament-core-data's architecture golden at `ea09341`, placed outside the trees `revendor_check` walks so it does not force a pin bump now. If it drifts from the original before then, a test fails in a prerelease repo, which is the system working.
* Re-verify the real `Err` assertion, a true end-to-end `read_records` pass, and the actual per-type breakdown.

No ignored test and no tripwire guards any of this. Ruling (Peter, 2026-09-19): an ignored test verifies nothing, and preserving reproducible evidence of a transitional state in a prerelease system with no users is ceremony. There is no regression to guard against and no consumer to protect.

### What the pin bump is load-bearing for

`validate_with_semantic_ir` emits TWO whole-document errors at the current pin, measured by running it: `SCHEMA_VIOLATION` at `/ir/types/0/fields/0` ("the member `constraints` is not one this schema admits", `Count.value`'s inline constraints) and at `/ir/types/2/relationships/0` ("a required member `verb` is absent", `Flow2`'s relationship against the old shape). `decide()` runs `schema::decide` first and only reaches `rules::decide` if diagnostics are empty, so both short-circuit the rules pass. Nothing about per-node intake behaviour is observable until the pin moves; the static 5-of-12 breakdown cannot be validated before then.

kreneskyp:
## Summary

Implements the half of #131 ("Implement model intake from domain packages")
that does not depend on agent-ix/filament-core-data#173: deleting the whole
Producer 1.2/Filament correspondence adapter. Nothing in this prerelease
repo consumes that path, so it is deleted outright rather than bridged with
a compatibility shim (no migration/fallback layer, per repo policy).

The actual Semantic IR 2.0.0 intake wiring (the other half of #131, which
depends on filament-core-data#173) is a separate, later PR. This PR does
not close #131 — remaining work: #131 still needs that intake-wiring PR.

## Removed (all owned by #131)

- `agent-ix-baseline-producer` Cargo dependency, and its lockfile entry
- `src/linking/composed/producer.rs` and `tests/producer_correspondence.rs`
- `AdmittedProducerModel`, `ModelSelection.producer`,
  `ProducerCompatibilitySelection`, `PRODUCER_CONTRACT_REVISION`
  (`OBSERVATION_CONTRACT_REVISION` is untouched — different contract)
- The whole `_with_producers` API family (`read_with_producers`,
  `admit_with_producers`, `admit_v2_with_producers`,
  `validate_with_producers`), collapsed into their plain siblings since the
  `producers`/`expected_producers` argument was always `&[]`
- `ExpectedProducerModel`, `ProducerSelection`, `Error::Producer`,
  `Target::Producer`
- Dead helpers left with no other caller once the above fell away:
  `producer_relationship`, `producer_path_matches`, `RelatedContext`,
  `type_export`, `validate_correspondence`, `producer_export_key`,
  `validate_producer_locus`, `relationship_binding`,
  `require_endpoint_type`, `producer_relation`, `producer_correspondence`,
  `producer_export`, `producer_locus`, `export_owned_key`

Native compilation paths that used to consult the producer model
(`relationship_authority` in `runtime.rs`, `related_occurrences` in
`controls.rs`, export validation in `protocol_artifact/models.rs`) now
refuse as `Unsupported::ProducerCorrespondence` / `Unsupported::Export`
after confirming the referenced name resolves, reusing the composed-model
layer's existing error codes rather than inventing new ones.

`docs/dependency-licenses.json` and `docs/dependencies.md` updated by hand
to drop the `agent-ix-baseline-producer` entry and correct the lock's
package count (181 -> 180).

## Deferred to the intake-wiring PR (depends on filament-core-data#173)

- `src/command/extraction.rs`'s hand-built `json!({...})` block is left
  as-is; removing it now breaks `Selected::compile`. It will be replaced
  when Semantic IR 2.0.0 intake actually lands.
- The tests #131 lists as moving to domain packages
  (`compiled_protocol_v2`, `composed_*`, `protocol_artifact`,
  `native_*emission`, `support`) stay coupled to the current native
  fixtures for now; only the producer-specific pieces inside them were
  removed here (dead imports/constructions in `tests/compiled_protocol_v2.rs`,
  the removed `_with_producers` helpers in
  `tests/support/native_protocol/mod.rs`).
- Re-vendor tooling for `resources/native-v1/` and
  `resources/complete-value/` — out of scope for this PR; see comment below.

## Verification

`git grep -n 'AdmittedProducerModel\|baseline-producer\|baseline_producer\|ProducerCompatibilitySelection\|PRODUCER_CONTRACT_REVISION'` returns nothing.

## Gates

- `cargo check --all-targets --features quire-extraction`: clean
- `cargo clippy --all-targets --all-features -- -D warnings`: clean
- `cargo fmt --check`: clean
- `cargo test --all-features`: 0 failures (unit + integration + doc tests)


kreneskyp:
## Re-vendor tooling for resources/native-v1/ and resources/complete-value/ — out of scope here

This PR does not build re-vendor tooling for `resources/native-v1/` or
`resources/complete-value/`. Findings, for whoever picks this up:

- Neither directory has any existing vendor script, Makefile target, or
  CLI subcommand in this repo today. Both READMEs describe a **manual**
  process: `resources/native-v1/README.md` points at a specific merged
  `quire-specification` PR (#15) with no re-run command; `resources/complete-value/README.md`
  says the files are unmodified bytes copied by hand with
  `git show <revision>:<path>` at a pinned commit
  (`a25c93c875231664d90f5d4b3f2ac51f1dc77a75`), plus `unicode-17.0.0/`
  downloaded by hand from unicode.org URLs.
- Both are checked for tamper, not freshness: `tests/complete_value_lock.rs`
  recomputes SHA-256 against `complete-value-lock.json`, and
  `tests/unicode_17_text.rs` checks against digests named in
  `value-text-unicode-17.md`. There's no test asserting the vendored bytes
  match a *live* re-fetch, so there's nothing today that would catch drift
  if the tooling didn't exist.
- Issue #131 itself never mentions re-vendoring these directories — it's
  about Semantic IR 2.0.0 model intake. If a re-vendor command is wanted,
  it needs its own ticket: design would need to cover pinning (explicit
  commit/rev per source, no floating refs), byte-exactness (raw `git show`/
  download, no reformatting), idempotency (re-run produces no diff), and a
  test that fails on drift — none of which exists as a pattern anywhere
  else in this repo to follow.

Not attempting an invented design for that here — flagging it so it's a
deliberate decision rather than a silent gap.

kreneskyp:
## Review addressed (commit 91f1a67)

An Opus review pass on this diff flagged four fixable findings, all applied in 91f1a67:

- **Refusal-code inconsistency (Medium)**: `relationship_authority` (native/runtime.rs) and `related_occurrences` (native/controls.rs) returned `Unsupported::ProducerCorrespondence`, while the wire-read equivalent (`models.rs::validate_related`) returns `Unsupported::Export` for the identical condition. Both native-path functions now return `Unsupported::Export`, matching every other unauthorized-export refusal already in those modules. `ProducerCorrespondence` stays reserved for a wire package that explicitly claims a correspondence (`models.rs:172`).
- **Missing test coverage (Medium)**: `validate_related`'s bounds check (`Invalid::Reference` for an out-of-range `Related.relationship` index) is now exercised by a new mutant in `choice_repeat_await_and_operation_events_preserve_static_edges_and_binding_kinds` (tests/protocol_artifact.rs). Its `Unsupported::Export` "found" branch is *not* independently tested — I traced the two independent checks that gate it (`models.rs`'s own per-model correspondence rejection, and `Graph::protocol`'s requirement that a live relationship's model *have* a correspondence) and confirmed they're mutually exclusive, so no wire package can reach that branch with a resolvable index in this build. Documented in place instead of faking an impossible fixture.
- **Style (Low)**: folded a stray `std::collections::BTreeSet::new()` into the existing `BTreeMap` import; replaced a hand-rolled boolean flag in `validate_related` with the equivalent `!related.is_empty()`.
- **Stale doc comment (Low)**: `SymbolKind::Relationship` in `scopes.rs` no longer describes producer-owned relationship semantics that don't exist anymore.

One finding is **not** applied, by design — flagging for the ticket owner rather than fixing unilaterally:

- **`state/evaluation.rs`/`state/input.rs` residual "1.2.0" pin (Medium, reviewer marked "owner decision")**: `AuthorityEvidence.static_selection.interface_version` is still hardcoded to check `== "1.2.0"`, and `StaticAuthority`/`CanonicalDigest` still carry doc comments naming "Producer 1.2". This is a *different* contract (FR-049 runtime authority/evidence, `state/input.rs`) than the `linking/composed/producer.rs` adapter this PR deletes, and #131's removal list only names `src/linking/composed/producer.rs`, `AdmittedProducerModel`, `ModelSelection.producer`, and the `state/input.rs` contract-revision constant (`PRODUCER_CONTRACT_REVISION`, already removed here). Widening the deletion to this residual pin would be scope creep beyond what #131 asks for and beyond what I was asked to do in this PR -- flagging it here so it's a deliberate decision, not a silent gap.

Gates re-run clean after the fixes: `cargo check`/`clippy -D warnings`/`fmt --check` clean, `cargo test --all-features` 0 failures (60 binaries, 3 pre-existing `#[ignore]`).

PR #153, opened 2026-09-18
QSL #138: re-vendor command for native-v1 and complete-value

### LINKED ISSUE #138: Build the re-vendor command for resources/native-v1 and resources/complete-value
Blocks #132.

#132 says to re-vendor `resources/native-v1/` and create `resources/complete-value/` "with the normal re-vendor command, never by hand". **That command does not exist.**

Verified on `origin/main`: no Makefile, no xtask crate, no `scripts/` directory, and no hits for "vendor" in any `.rs`, `.toml`, `.md` or CI file. Both resource trees are hand-maintained today — `git show <revision>:<path>` for complete-value pinned at `a25c93c`, and direct unicode.org downloads for unicode-17.0.0. `resources/native-v1/README.md` documents the hand-copy process explicitly.

#131 never mentioned re-vendoring these, so this is not that ticket's scope either. It needs its own.

## Why it cannot be improvised

`resources/native-v1/` is a **historical selection**, not a mirror of current source. Its README states the invariant directly: "Updating current compiler diagnostics does not update this historical selection. A changed interpretation requires an explicit registry/definition change, not synchronizing these resources with current source files." The closed registry at `src/linking/composed/definition_source.rs` holds the embedded originals, and a test compares retained resource bytes against it.

A re-vendor tool that naively syncs from a source tree would silently destroy that invariant. That is the whole design problem, and it is why the deleting agent stopped rather than inventing one.

## What needs deciding

- **Pinning strategy.** The command takes an explicit commit or revision and records it. It never resolves "latest" and never re-synchronizes against a working tree.
- **Byte-exactness.** Vendored bytes match the pin exactly. The existing registry byte-comparison test keeps passing.
- **Idempotency.** Running it twice against the same pin is a no-op, provable by a test.
- **Drift detection.** A check that fails when the vendored tree no longer matches its recorded pin, so drift is caught rather than discovered.
- **Shape.** A `Makefile` target wrapping an `xtask` binary, or a checked-in script. The repo currently has none of these, so whichever is chosen sets the convention — say why in the PR.
- **`resources/complete-value/` does not exist yet.** The command creates it from a pinned quire-specification commit, on the same rules.

## Out of scope

Not a migration or compatibility tool. It vendors one explicit pin at a time. There is no "upgrade path" between vendored revisions — a new pin replaces the old bytes wholesale, per the standing no-compatibility ruling.

## Sequencing

This lands before #132, which assumes it exists.

kreneskyp:
## Summary

Adds an `xtask` workspace crate providing `cargo xtask revendor` / `cargo xtask revendor-check`, wrapped by a top-level `Makefile` (`make revendor`, `make revendor-check`). It replaces the hand-copy vendoring process for `resources/native-v1` and `resources/complete-value` with a command that:

- reads every path from an **explicit, checked-in commit sha** via `git show <sha>:<path>` against a local clone path (or this repo's own history), never resolving "latest" and never reading a working tree;
- vendors **exactly** the path list recorded in a new `VENDOR.json` manifest per tree — it never globs or scans a source tree to discover files, because `native-v1` is a historical selection, not a mirror;
- treats the six `unicode-17.0.0/` files as external-pinned-by-digest (recorded URL + sha256, `revendor`/`revendor-check` only ever verify the digest, never fetch over the network);
- is idempotent: running `revendor` twice at the same pin writes nothing the second time (proven by test);
- detects drift offline: `revendor-check` (also wired into `cargo test --workspace` via `xtask/tests/revendor.rs`) fails if any vendored byte differs from its recorded digest, or if an untracked file exists outside the manifest and README.

**Why xtask.** The program is porting tooling to Rust; an `xtask` binary crate (the cargo-xtask convention, aliased via `.cargo/config.toml`) lets this command share code directly with the crate's own `ByteDigest` type for the byte/registry checks, rather than standing up a separate script runtime.

## Pins established (verified by byte-comparison, not assumed from the issue text)

- **native-v1**: `agent-ix/quire-specification` commit `4d6230eb8aa9766ff3017360962f2d6368d74cb3` (QSpec PR #15's merge commit) — confirmed by diffing every one of the 67 vendored `qspec`-sourced paths against `git show <sha>:<path>` in a local clone; zero mismatches.
- **native-v1**'s `external/quire-spec-language/` historical selection: this repository's own commit `bf9960e8d818e4d72182c4691c8aaf4e2cd7dffd` (the parent of the commit that vendored it) — confirmed by exact byte match.
- **complete-value**: `agent-ix/quire-specification` commit `7d7943ab1482e091f6d126401ada38957c4a1ccf` — the issue text cites `a25c93c`, but that is a stale pin from an earlier re-vendoring round; the current README and the most recent re-vendoring commit both cite `7d7943a...`, and all 58 vendored paths byte-match it exactly. Went with the verified-current pin, not the issue's outdated reference.
- **complete-value**'s six `unicode-17.0.0/` files: no git revision exists to pin (they're downloaded from `unicode.org`), so they're recorded by origin URL + sha256 only; `revendor-check` verifies the digest, `revendor` never fetches them.

Manifest format is JSON (`VENDOR.json`, not `.toml`): the repo has zero `toml` dependency, uses `serde_json` pervasively, and already has a "lock file" JSON convention (`complete-value-lock.json`).

## CI

Added `--workspace` to the 5 relevant `cargo clippy`/`cargo test`/`cargo build` steps in `.github/workflows/ci.yml`. Without it, the new `xtask` workspace member would be silently unlinted and untested — this repo's root `Cargo.toml` is both the workspace root and a package, so bare `cargo` invocations there only ever touch the root package.

## Gates (this checkout)

- `cargo fmt --all -- --check` — clean
- `cargo clippy --workspace --all-targets --all-features -- -D warnings` — clean
- `cargo test --workspace --all-features` — all passing, including the existing byte-comparison tests (`tests/complete_value_lock.rs`, `tests/composed_definition_source.rs`, `tests/unicode_17_text.rs`) and all 13 new xtask tests
- `cargo xtask revendor-check` — `native-v1: 69 matched, 0 drifted, 0 stray`, `complete-value: 64 matched, 0 drifted, 0 stray`

**Pre-existing, unrelated failure noted but not fixed**: `cargo clippy --all-targets -- -D warnings` run *without* `--all-features` fails to compile on a clean `origin/main` checkout (confirmed via a throwaway worktree), because several `tests/model_*.rs` files call `ProducerKey::fixture(...)`/`ModelSelection::fixture(...)` gated behind the `test-support` feature, which isn't enabled by that invocation. This is unrelated to this PR (reproduces identically on `origin/main` with no xtask changes at all) and out of scope for #138 — flagging it for a separate fix.

## Judgment call for review

I did not add a new spec FR/TC for this tooling. No existing requirement precisely owns "how vendored resources are re-vendored/checked," and the issue didn't ask for spec authoring; inventing one risked misrepresenting what it actually verifies. Happy to add one if the maintainers want this tracked as a formal requirement.

Closes #138


kreneskyp:
## Review: PR #153 (closes #138), head `858195f`

**Verdict: FIX.** One High: the in-repo idempotency test fails on CI's shallow checkout. One Medium: the spec has no requirement or test case for this tooling. The pins, byte-exactness and drift detection are correct.

### Verified
- **Pins, checked against the real clone.** `cargo xtask revendor --qspec-clone ~/dev/quire-specification` printed `native-v1: 0 written, 69 unchanged` and `complete-value: 0 written, 58 unchanged, 6 external verified`. `git status` was clean afterwards, and neither VENDOR.json changed. `7d7943ab…` is the correct complete-value pin: README, FR-055 and IT-011 all cite it, and `a25c93c` was stale.
- **Independent spot checks** (sha256 of `git cat-file blob <sha>:<path>` compared with the vendored file and the manifest). 3 random files from each of qspec@4d6230eb, self@bf9960e8 (both files) and qspec@7d7943ab matched. All 6 unicode digests matched.
- **`cargo xtask revendor-check`**: `native-v1: 69 matched, 0 drifted, 0 stray`, `complete-value: 64 matched, 0 drifted, 0 stray`.
- **Registry invariant.** `src/linking/composed/definition_source.rs` is untouched. `tests/composed_definition_source.rs` passes, and so do `complete_value_lock.rs` and `unicode_17_text.rs`. No vendored byte changed in this diff.
- **CI `--workspace`** is correct. Root `default = []`, and `cargo tree --workspace --no-default-features -e features -i quire-spec-language` shows that xtask's path dependency does not unify extra features into the minimal-feature lanes.
- **Git calls** use `Command` with separate args and no shell. The `commit:path` argument always starts with validated 40-char hex, so it cannot be read as an option.
- **No `unwrap` on I/O or git output in non-test code.** The three `expect`s (`fsutil.rs:334`, `manifest.rs:128`, `main.rs:551`) guard structural invariants, not I/O.
- **Both READMEs describe current state only.** The false line "no separate per-file checksum inventory" is correctly removed.
- **Drift detection runs in plain `cargo test`.** It fails on a changed byte and on a stray file, and those tests can fail.

### Gates (858195f, this worktree)
- `cargo fmt --all -- --check`: clean
- `cargo clippy --workspace --all-targets --all-features -- -D warnings`: clean
- `cargo test --workspace --all-features`: all pass (xtask: 6 unit + 7 integration)

### Punch list

1. **High: `xtask/tests/revendor.rs:137-178`.** `revendor_is_idempotent_against_this_repositorys_own_history` needs commit `bf9960e8…` in local history. CI uses `actions/checkout@v4` with the default `fetch-depth: 1` (`ci.yml:12`). Reproduced: a depth-1 clone of this branch fails `git cat-file -e bf9960e8…^{commit}`. The test then hits `UnknownCommit`, `.unwrap()` panics at :160, and both `cargo test --workspace` CI steps go red. The other idempotency test is skipped unless `QSPEC_CLONE_PATH` is set, so no idempotency test runs in CI.
   **Fix:** make the test hermetic. In a tempdir, `git init`, commit a file with `git -c user.name=t -c user.email=t@t -c commit.gpgsign=false commit`, and point a `Source::SelfRepo` at that repo via `Sources.workspace_root`. Run `revendor` twice. Assert that the second run `is_noop()`, the bytes are unchanged, and the `Manifest::save` output is byte-identical. Add a negative control: alter the dest file between runs and assert that the second run rewrites it. That proves the test can fail. Keep the `QSPEC_CLONE_PATH` test as an extra.
2. **Medium: spec gap (no FR/NFR/TC; new xtask tests carry no `/// Tracing:` tag).** QSL traces tests to requirements, and review gaps get ACs and tests written. Add **NFR-011 "Vendor resources from exact pins"** in `spec/non-functional/` (next free locally; I found no NFR-011 or TC-149 on any `origin/*` branch), with:
   - AC-1: every git source records a full 40-hex commit, bytes are read only from that commit, and a short sha, ref or missing clone is refused.
   - AC-2: vendored bytes equal the blob at the pin, and the registry byte comparison holds.
   - AC-3: a second run at the same pin writes nothing and leaves the manifest byte-identical.
   - AC-4: `revendor-check` fails offline on a changed byte, a missing file and a stray file, and runs in the default `cargo test`.
   - AC-5 (if you take #3): a new pin replaces the tree wholesale.

   Add **TC-149…TC-152** in `spec/test-cases/`. Add rows to the repo-level `spec/tests.md`, which already covers repo tooling (FR-012/NFR-005), and widen its Overview scope line. Add `/// Tracing: TC-…` to each xtask test.
3. **Medium: `xtask/src/lib.rs:165-189`, `resources/native-v1/README.md` ("A new pin replaces these bytes wholesale via `revendor`"), and the same sentence in `resources/complete-value/README.md`.** `revendor` never deletes anything. Scenario: a new pin drops a path from the manifest. The old file stays, and `revendor-check` then reports STRAY, which leaves a hand deletion as the only remedy. That contradicts both the README and #138's "a new pin replaces the old bytes wholesale". **Fix:** after writing, have `revendor` remove any regular file under the tree that is not in the manifest (other than README.md and VENDOR.json), list it in a `removed` field on `RevendorReport`, and add a test.
4. **Low: `xtask/src/manifest.rs:133-166` and `fsutil.rs:345`.** `validate` does not constrain `path`, `dest` or `dest_prefix`. `tree_root.join` with an absolute path, or with `..`, writes outside the tree. **Fix:** reject absolute paths, `..`, empty components and `\` in `validate`, and add a test.
5. **Low: `xtask/src/fsutil.rs:329-338`.** Symlinks and other non-regular entries are skipped silently, so a stray symlink in a vendored tree passes `revendor-check`. **Fix:** report any non-regular, non-directory entry as stray.
6. **Low: `xtask/src/git.rs:258,275`.** The `git` child process inherits `GIT_DIR`, `GIT_WORK_TREE` and `GIT_INDEX_FILE`. When those are set (for example when run from a git hook), `GIT_DIR` takes precedence over `-C`, so reads come from the wrong repository. **Fix:** `.env_remove` each of them on both commands.
7. **Low: `xtask/src/main.rs:437`.** `USAGE` advertises `[--qspec-clone <path>]` for `revendor-check`, but `run_check` (:503-511) refuses that flag. **Fix:** remove it from the usage text.
8. **Low: `xtask/tests/revendor.rs:212-244`.** When the real-clone test fails, it has already written into the checked-in `resources/` and `VENDOR.json`, so a failed run leaves a dirty tree. **Fix:** copy the tree to a tempdir and revendor the copy.
9. **Low: `xtask/src/lib.rs:222-231`.** Nothing tests the "missing file" drift branch. Add one test in which a manifest entry has no file on disk.
10. **Low: docs.**
    - Neither README says how to move a pin (edit `commit` in `VENDOR.json`, then run `make revendor`).
    - Neither README says that the native-v1 `self` source needs non-shallow history.
    - The comment at `.cargo/config.toml:3` points to "xtask/README topics", which do not exist.
11. **Nit: `xtask/src/error.rs:340`.** `Usage(&'static str)` covers two distinct refusals as prose (`tree.rs:390`, `main.rs:508`). Give the qspec-clone refusal its own variant.

The pre-existing `clippy` failure without `--all-features` (`test-support` fixtures) is out of scope here. It needs its own ticket.

kreneskyp:
## Re-review: PR #153, head `8ba5af9` (diff `858195f..8ba5af9`)

**Verdict: FIX.** Two small spec-wording edits remain (R1, R2). Every code item from the first review is resolved, and I found no new code problems.

### First-review items
| # | Status | Evidence |
|---|---|---|
| 1 High, idempotency in CI | ✅ | I cloned the pushed branch with `git clone --depth 1` (`is-shallow-repository: true`). `cargo test --locked -p xtask --all-features` passes there, 21/21 including `revendor_is_idempotent_against_a_hermetic_git_repository`. The negative control (tamper the file, third run rewrites it) makes the no-op assertions able to fail. |
| 2 Medium, spec coverage | ✅ (see R1, R2) | NFR-011 and TC-149…152 are added, the xtask tests carry `#[trace]` and `/// Tracing:` tags, and `spec/tests.md` has the rows. On every `origin/*` branch other than this one, no document declares `id: NFR-011` or `id: TC-149`–`TC-152`. `quire validate` on the 5 new docs plus `tests.md`: 6/6 grammar-clean. |
| 3 Medium, wholesale replace | ✅ | `remove_files_the_manifest_no_longer_lists` runs only after every source succeeds, reports what it removed, and `is_noop` now includes removals. The test would fail if removal were missing. |
| 4 Low, path safety | ✅ | `is_safe_relative_path` covers `path`, `dest` and `dest_prefix`, and tests cover absolute paths and `..`/`.` components. Unix-only is accepted per repo convention. |
| 5 Low, symlinks | ✅ | Non-regular entries are now reported as stray; a `cfg(unix)` test covers it. `remove_file` unlinks the link, not its target. |
| 6 Low, git env | ✅ | `git_command` removes `GIT_DIR`, `GIT_WORK_TREE`, `GIT_INDEX_FILE`, `GIT_OBJECT_DIRECTORY` and `GIT_ALTERNATE_OBJECT_DIRECTORIES`. |
| 7 Low, USAGE | ✅ | Fixed, and a test covers it. |
| 8 Low, real-clone test dirtied tree | ✅ | It now works on a tempdir copy. I ran it with `QSPEC_CLONE_PATH=~/dev/quire-specification`: it passes and `git status` stays clean. |
| 9 Low, missing-file drift | ✅ | `revendor_check_reports_a_missing_vendored_file` added. |
| 10 Low, docs | ✅ | Both READMEs now explain how to move a pin, and native-v1's also covers the shallow-clone caveat. The `.cargo/config.toml` comment is fixed. |
| 11 Nit, error variant | ✅ accepted | `CheckRefusesQspecClone` is typed. The unknown-`--tree` case staying as `Usage` is fine; it really is a usage error. |

### Remaining (spec wording only)
- **R1 Low: `spec/non-functional/NFR-011-vendor-resources-from-exact-pins.md`, Scope, 2nd sentence.** "It does not resolve a branch, tag or "latest" ref, fetch over the network, or change how the closed registry … interprets the resulting bytes" lists behaviour the tool doesn't have. Our spec rule is to state what is, not what isn't. **Fix:** state it positively, e.g. "It reads only the full commit recorded in each manifest from a local repository, and the closed registry's interpretation of the resulting bytes stays owned by FR-036." The Statement and ACs already say this positively.
- **R2 Low: `spec/tests.md`, "Six coverage rules", first sentence.** It still reads "All nine FR-012 ACs and both NFR-005 metric obligations are verified above." The Overview now puts NFR-011 in scope. **Fix:** add "and all five NFR-011 ACs (TC-149…TC-152)".

### Pre-existing, not this PR
`quire validate` fails `spec/tests.md:16` because the column header `Coverage Status` doesn't match the `Status` the module asserts. That header predates this PR, so it needs its own ticket.

### Gates at 8ba5af9 (worktree `target/`)
- `cargo fmt --all -- --check`: clean
- `cargo test --locked --workspace --all-features`: all pass, 0 failures
- `cargo xtask revendor-check --tree all`: `native-v1: 69 matched, 0 drifted, 0 stray`; `complete-value: 64 matched, 0 drifted, 0 stray`

With R1 and R2 fixed, this is MERGE with no further review round needed.

PR #156, opened 2026-09-18
Skip test-support fixture tests under default features

### LINKED ISSUE #154: cargo clippy --all-targets without --all-features fails: tests/model_*.rs call test-support-gated fixture()
On `origin/main`, `cargo clippy --all-targets -- -D warnings` (no `--all-features`) fails to compile: several `tests/model_*.rs` call `fixture(...)` methods gated behind the `test-support` feature. Reproduced on a clean main worktree by the #138 author (PR #153).

CI passes only because it runs with `--all-features`, so a default-feature build of the test targets is broken and unchecked.

## Fix
Either declare the feature requirement on those test targets (`[[test]] required-features = ["test-support"]` in Cargo.toml) or make the fixtures available to integration tests without the feature — whichever the crate's feature design intends. Add a CI step that builds tests with default features so this can't regress.

kreneskyp:
## Summary

`cargo clippy --all-targets` (default features) fails to compile
`tests/model_conformance.rs`, `tests/model_dispatch.rs`,
`tests/model_normalization.rs`, `tests/model_population.rs` and
`tests/model_systems.rs`: they call `ProducerKey::fixture`,
`ModelSelection::fixture` and `ProducerDigest::of_identity`, which are
gated `#[cfg(any(test, feature = "test-support"))]`. Integration tests
compile the crate as an ordinary dependency (never `cfg(test)`), so they
need the `test-support` feature — and this crate's `default = []` makes
"default features" and "no features" identical, so any default-feature
build hits this.

## Fix

- `Cargo.toml`: declare `required-features = ["test-support"]` on the
  five affected `[[test]]` targets. This is the mechanism the feature
  design already implies (`test-support` is a real, documented feature
  gate) — Cargo now skips compiling these targets under an insufficient
  feature set rather than failing to compile them.
- `Makefile`: add `ci-default-features` (fmt/clippy/test under default
  features specifically) and `ci-all-features`, combined into `ci`, as
  the durable local/CI regression check the issue asks for. A future
  `test-support`-gated test file added without `required-features`
  fails `ci-default-features` the same way the production build already
  fails today.

## Decisions

None beyond the mechanism named above — this is a mechanical fix using
the feature design already in place, not a new design choice.

## Test plan

- [x] `cargo fmt --all -- --check`
- [x] `cargo clippy --locked --workspace --all-targets -- -D warnings` (default features)
- [x] `cargo clippy --locked --workspace --all-targets --all-features -- -D warnings`
- [x] `cargo test --locked --workspace` (default features)
- [x] `cargo test --locked --workspace --all-features`
- [x] `make ci` (`ci-default-features` + `ci-all-features`)

Closes #154


kreneskyp:
## Review: PR #156 (fixes #154), head 80c60a6d388dec91c7ef2cf7c1e3c9d05d1b2fa6

**Verdict: FIX.** The Cargo.toml change is correct. The one thing to fix is that `make ci` is not the documented gate.

### What checks out
- `required-features = ["test-support"]` covers exactly the five integration tests that call the gated fixtures: `model_conformance`, `model_dispatch`, `model_normalization`, `model_population` and `model_systems`. A grep for `::fixture(` and `of_identity(` over `tests/` finds no others. This is the first option #154 offers.
- It does not hide tests from the all-features run. `make ci` log: the default `cargo test --locked --workspace` never runs the five targets. `cargo test --locked --workspace --all-features` runs all five. `autotests` is still on, so the other test targets are still found.
- `make ci` exits 0 locally. It runs fmt, then clippy and test under default features, then clippy and test under `--all-features`.
- `.github/workflows/ci.yml:20-22` already had a `--no-default-features` clippy/test lane. That is the same as default here, because `default = []`. This PR makes that lane pass, so the "add a CI step" part of #154 is covered by the existing workflow.

### Findings
1. **README.md:252-260 (Medium).** The README still lists raw cargo commands as the local check sequence and never mentions `make ci`. The PR body presents `make ci` as "the durable local/CI regression check", but nothing documented points a contributor to it. Failure scenario: a contributor follows the README, which does cover default-feature clippy/test, and `make ci` stays unused and drifts (see 2). Fix: make `make ci` the documented command in the README checks block, or drop the target.
2. **Makefile:31-40 (Low).** The gate is now defined in two places, and they already differ. `ci.yml` also runs the clean build (`--target-dir target/clean`), `fixture-audit -- self-test` and the parse example. `make ci` runs none of these, and `ci.yml` does not call `make ci`. Failure scenario: a change breaks `fixture-audit self-test`, `make ci` reports green locally, and the dispatched workflow fails. Fix: make `ci` match `ci.yml`, or have the workflow call `make ci`.
3. **Makefile:27 (Low, comment only).** "reachable only under `--all-features`" is wrong: `--features test-support` also turns them on. Suggest "reachable only with the `test-support` feature".

Gates run at the head: `make ci`, exit 0.

kreneskyp:
## Re-review: PR #156 at 1a025dd1ad1b516064b8d2d4b8821cd6aedf0ddf

**Verdict: MERGE.** All three earlier findings are resolved.

**`make ci` against `.github/workflows/ci.yml`, step by step:**
| ci.yml step | make target |
|---|---|
| Install toolchain, install cargo-llvm-cov | not mirrored (environment setup, not a gate) |
| fmt `--check` | `ci-default-features` |
| clippy `--all-targets --no-default-features` | `ci-default-features` (default features; the same build because `default = []`) |
| test `--no-default-features` | `ci-default-features` (same note) |
| clippy `--all-targets --all-features` | `ci-all-features` |
| test `--all-features` | `ci-all-features` |
| clean build `--target-dir target/clean` | `ci-clean-build` |
| `fixture-audit -- self-test` | `ci-clean-build` |
| parse example | `ci-clean-build` |

Every gating step is covered.

**Other checks:**
- `make ci` at this head: exit 0. `fixture-audit` passed 6 negative controls, and the parse example returned `"status":"parsed"`.
- The README now points at `make ci`.
- The Makefile:27 comment is fixed.
- The only closing keyword, in the PR body and in the commits, is `Closes #154`.

**Non-blocking note:** the README calls these "the exact commands" `ci.yml` runs. The two minimal-feature steps differ in flag spelling (none vs `--no-default-features`). They build the same thing only while `default = []`.

Issue #146, opened 2026-09-18
[QSL#146] FR-151/FR-152 cleanup: depth-ceiling miscounts, redundant re-cloning, dead code, collapsed causes


kreneskyp:
PR #144 review, non-blocking findings filed as a ticket rather than fixed in
that PR.

## MAX_CONFORMANCE_DEPTH / MAX_DISPATCH_DEPTH count nodes, not depth

`MAX_CONFORMANCE_DEPTH` (`src/model/conformance.rs`) and
`MAX_DISPATCH_DEPTH` (`src/model/dispatch.rs`) are documented as depth
ceilings, but the walks they bound (`type_conforms`, `build_family`) count
total nodes visited, not path depth. A type with a wide (not deep)
ancestor cone — more than 128 generalization edges reachable from one
type, regardless of how shallow the graph actually is — spuriously refuses
`resource_exhausted`. Either the doc should describe this as a node-count
ceiling, or the counters should measure actual path depth.

## ConformanceIndex::build re-clones per entry point

`ConformanceIndex::build` (`src/model/conformance.rs`) re-clones its
generalization/field/operation maps once per caller (`check_*` entry
point), rather than being built once and shared. Not a correctness issue;
worth revisiting if `ConformanceIndex::build` shows up in profiling.

## Dead code

- `EstablishedFact::field()` — unused accessor.
- `OperationMemberRecord::has_own_precondition` — written by every fixture,
  never read anywhere in `src/model`.

## conformance.rs collapses two distinct redefinition-target failure causes

Around `src/model/conformance.rs:805-811`, `resolve_redefinition_target`
reports "no target" and "ambiguous targets" (zero vs. several distinct
valid targets) under the same `redefinition-target` cause, losing the
distinction between the two failure shapes in the refusal detail.

## Scope

Independent small fixes; can land separately or together. None are
blocking — filed per PR #144 review's "also ticket, do not fix" list.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #120, opened 2026-09-15
[QSL#120] [V1-A04] Implement model graph binding, closed lookup, inheritance, and dispatch


kreneskyp:
Parent: #1. Lane: Agent-A. Depends on V1-A00, V1-A02, and the accepted Filament producer interface.

Primary capabilities: V1-TYPE-021 through V1-TYPE-029, V1-EXPR-017 through V1-EXPR-022.
Normative contracts: FR-150–153 plus existing state/model FRs, AD-006, I03–I05; TC-195–198 and existing graph/state cases.

Implement immutable model correspondence, original/effective provenance, object identity/references, closed populations, relationship traversal, inheritance/conformance/subsetting/redefinition, systems-model references and unique most-specific dispatch.

Exit: no ambient registry or display-name identity; derivations replay exactly; closure and ambiguity failures expose no partial substitute. Do not edit Filament producer code.

kreneskyp:
blocked on agent-ix/filament-core-data#143 because Producer interface 1.2.0 emits no generalization, subsetting, redefinition, operation-signature or port-direction records, and FR-150/151/152 require those producer facts for correspondence normalization, conformance and dispatch (slices A/B, part of C).

Unblocked and proceeding: a QSpec amendment for #120's other readiness gaps (definition artifact `quire.model.complete/v1`, effective identity domain, charge points for deref/navigation/reaches/lookup/allInstances/dispatch, catalog codes, lookup and allInstances typing, reference identity under inheritance, TC-195..198 vectors), then runtime closed environments (slice D). The amendment starts after QSpec PR #74 merges, because both touch value-accounting and native-diagnostics.

kreneskyp:
The spec contracts for #120 merged in QSpec PR #75 as 7d7943ab1482e091f6d126401ada38957c4a1ccf (FR-150..153, TC-195..198, `quire.model.complete/v1`).

The remaining block is narrower. blocked on agent-ix/filament-core-data#143 because inheritance, subsetting, redefinition, operation/interface signatures, port direction and typed multiplicity need producer interface 1.3.0 records. QSL work that does not need those records proceeds after #119: reference identity and upcast, lookup and populations, graph reachability, and the 1.2.0 per-item refusals.

kreneskyp:
[...]

PR #157, opened 2026-09-18
QSL #120 slice 4: FR-151 conformance completion (R01, R06, R07, R08)

kreneskyp:
## Summary

Part of #120. Implements and tests, on TC-196 (FR-151) vectors, rules **R01, R06, R07 and R08**. R01, R07's second shape, and R06 are enforced at their real pipeline boundaries (`normalize.rs` / `population.rs`'s `admit_binding`), not in a standalone checker. **V1-TYPE-026 (inheritance/conformance) and V1-TYPE-027 (subsetting/redefinition) are in progress, not backed**: composing `conformance.rs`'s conformance checks into `normalize.rs` phase 4 with `conformance.axis` `f(T)` accounting is QSL #165. Refs #165.

- **R01** (FR-151-AC-2, specialization-cycle): `normalize.rs`'s cycle-detection refusal (phase 3) now names the full contributing declaration chain, rotated to start at its least key, not just the closing edge.
- **R07** (FR-151-AC-2, redefinition-target ambiguity): R07 has two shapes. The **first shape** (one redefining member queried in isolation resolves to zero or multiple genuinely-inherited targets) is covered by `conformance.rs`'s `resolve_redefinition_target` — this function is unwired from `src/`'s pipeline (like its `conformance.rs` siblings) and stays that way; #165 composes it later. The **second shape** (several distinct members all redefine one shared inherited target, e.g. `B/z` and `B/z2` both redefining `A/x`) cannot be seen by a single-member query, so it is now refused directly in `normalize.rs` **phase 4** (`apply_redefinitions`), the real boundary that sees every redefiner of a target at once: when no owner dominates every other undominated edge, the check restricts to the *most-derived* owners among those edges (an owner properly dominated by another edge's owner has already lost, exactly as it would if that owner had only one redefiner, and takes no part in the ambiguity test) and refuses `redefinition-target` when they're all one owner, or falls through to `derivation-conflict` for genuinely incomparable lineages.
- **R08** (FR-151-AC-6, refinement obligation): `check_field_refinement_obligation` no longer trusts a caller-stated `EstablishedFact`. `bundle.rs`'s `EstablishedFact` is renamed `PostconditionClause` (`Interval{lower,upper}` -> `Comparison{operator,literal}`, one guard form per clause). `conformance.rs` rebuilds each clause as a small synthetic `Node` guard tree (seeded from the *redefined parent* field's own scalar domain, per FR-146's rule) and runs it through a crate-internal `crate::value::established_field_fact`/`Established` (reusing FR-146's own `Definedness` fact derivation in `value/expression/facts.rs`, not a second implementation).
- **R06** (FR-151-AC-10, `binding.subset-value`): `population.rs`'s `admit_binding` now actually charges `binding.subset-value` (previously declared but never charged) once per admitted member's subsetting-feature value, and refuses `invalid_runtime_input`/`subsetting-violation` when a value is absent from the subsetted feature's values. `PopulationMember` gained `field_values: Vec<MemberFieldValues>`. `AdmissionChargePoint::is_charged()` is removed since both charge points are now genuinely charged.

## Deviations from the literal task wording (flagging explicitly)

- The task described R07 as "split the two `redefinition-target` causes." The two shapes are genuinely different ambiguities that TC-196 gives the same cause tag: kept `cause` unified as `redefinition-target`, moved the second shape's detection to the real boundary (`normalize.rs` phase 4) instead, since a standalone per-member checker (`resolve_redefinition_target`) structurally cannot see sibling redefiners and can't ever detect that shape.
- QSL #146 is **not** closed by this PR (no closing keyword used): only the redefinition-target parts (inside this PR's own touched files) are addressed. `MAX_CONFORMANCE_DEPTH`/`ConformanceIndex::build` re-cloning, mentioned in #146, is untouched — out of scope for this slice.
- QSL #165 (composing `conformance.rs`'s checks into `normalize.rs` phase 4 with `conformance.axis` `f(T)` accounting) is **not** closed by this PR either. Refs #165.
- The task instruction to update a "test matrix tracking artifact" was checked against `spec/tests.md` (an unrelated TM-001/FR-012 matrix) and `plan/Plan-013-complete-v1-delivery/plan.md` (a coarse checklist covering far more than this slice, including out-of-scope dispatch/systems rows). Neither is updated by this PR; traceability is via the `#[trace(...)]` tags below and this PR description instead.

## Scope boundaries honored

- `src/model/key.rs`: untouched.
- `normalize.rs`: phase 3 (cycle-detection refusal) **and** phase 4 (`apply_redefinitions`, R07's second shape) changed. Phases 1-2 untouched.
- `bundle.rs`: touched only for the `EstablishedFact` -> `PostconditionClause` rename this R08 fix requires.
- `conformance.rs`: `resolve_redefinition_target`'s doc comment now states explicitly that it covers R07's first shape only; it stays unwired (dead code, like its siblings) pending #165.
- `value/expression/{syntax,check,evaluate}.rs`, `value/reference.rs`, `population.rs`'s `lookup`/`all_instances`: untouched.
- `population.rs`: the only functional change is the `binding.subset-value` charging loop in `admit_binding`.
- Dispatch (D01-D08) is untouched, as instructed.

## Rows/ACs now backed

| Row | AC | Real boundary | Test(s) |
|---|---|---|---|
| R01 | FR-151-AC-2 | `normalize.rs` phase 3 | `model_normalization.rs::r01_a_closing_generalization_cycle_names_the_full_rotated_chain`, `r01b_the_cycle_listing_excludes_a_type_that_only_leads_into_it` |
| R07 (first shape) | FR-151-AC-2 | `conformance.rs::resolve_redefinition_target` (unwired; #165) | `model_conformance.rs::r07_zero_or_multiple_inherited_targets_refuse_redefinition_target` |
| R07 (second shape) | FR-151-AC-2 | `normalize.rs` phase 4 (`apply_redefinitions`) | `model_normalization.rs::r07_two_redefiners_owned_by_the_same_type_refuse_redefinition_target_through_normalize`, `r07_a_less_derived_owners_redefiner_is_excluded_from_the_same_owner_test` |
| R08 | FR-151-AC-6 | `conformance.rs::check_field_refinement_obligation` | `model_conformance.rs::r08a`-`r08i` |
| R06 | FR-151-AC-10 | `population.rs::admit_binding` | `model_population.rs::r06_subsetting_violation_refuses_after_the_charged_subset_value`, `r06_subsetting_satisfied_admits_with_the_charged_subset_value`, `r06_duplicate_field_values_refuse_rather_than_silently_keep_the_first` |

V1-TYPE-026/V1-TYPE-027 are **not** claimed as backed by this table: the rows above are tested directly against `normalize`/`admit_binding`/the standalone conformance checks, not against a composed `conformance.axis` accounting pipeline. That composition is #165.

## Gates (CARGO_TARGET_DIR + CARGO_PROFILE_DEV_DEBUG=0/CARGO_PROFILE_TEST_DEBUG=0 on every cargo invocation, per current disk-conservation practice)

- `cargo fmt --check`: clean
- `cargo test --locked --workspace --all-features`: all green (every suite, including the new/updated tests above)
- `cargo clippy --locked --workspace --all-targets --no-default-features -- -D warnings`: pre-existing failure in `tests/model_conformance.rs` etc. (`ProducerKey::fixture`/`ModelSelection::fixture` need the `test-support` feature, not enabled by `--no-default-features`) — **not introduced by this PR**; already tracked separately (see the in-progress `fix/154-default-feature-tests` branch). Verified the equivalent `--all-features` clippy/test lanes are clean, and `cargo build --locked --workspace --no-default-features` (no tests) succeeds.
- `cargo run --locked --no-default-features --bin fixture-audit -- self-test`: passed
- `cargo run --locked --no-default-features -- parse test:parent fixture:1 tests/fixtures/parent.native`: passed
- `quire coverage`: new trace tags (`TC-196`/`FR-151-AC-2`/`FR-151-AC-6`/`FR-151-AC-10`) report identically to the pre-existing R02-R06 tags in the same files ("matches no declared row" — FR-151/TC-196 rows aren't in `quire coverage`'s declared-row source for this crate at all, pre-existing and unrelated to this change).

## Remaining work

- Dispatch (D01-D08) is untracked here, as instructed — separate slice.
- QSL #146's `MAX_CONFORMANCE_DEPTH`/`ConformanceIndex::build` re-cloning concern is untouched.
- QSL #165 composes `conformance.rs`'s checks (including `resolve_redefinition_target`'s first-shape check) into `normalize.rs` phase 4's real pipeline with `conformance.axis` `f(T)` accounting, backing V1-TYPE-026/V1-TYPE-027. Refs #165.
- The `--no-default-features` + `test-support` feature gap (affecting every `model_*.rs` integration test file, not just ones this PR touches) is tracked on the separate `fix/154-default-feature-tests` work already in progress.


kreneskyp:
## Review of PR #157 at head 50bf2ba: **FIX**

This review covers code, Rust idioms and gaps, scoped to the diff. The normative source is QSpec origin/main: FR-151, TC-196 and FR-146.

**Gates at 50bf2ba all pass:** `fmt --check`, `clippy --locked --workspace --all-targets --all-features -D warnings`, and `test --locked --workspace --all-features`.

I checked each finding below with a throwaway probe test, then reverted it.

### Findings

1. **High: multiple postcondition clauses are never combined.** Location: `src/model/conformance.rs` about lines 826 and 869.
   - Each clause is derived on its own, starting from `Facts::default()`, and `find_map` keeps only the first interval.
   - The effective postcondition is a conjunction, so the facts should be intersected.
   - Probe: `A.c` is `Int[-5,9]` and `B.cs` is `Small [0,5]`. The clauses are `[cs >= 0, cs <= 5]`.
     - Actual result: `Refused("established interval [0, 9] ...")`.
     - Expected result: `Compatible`.
   - Before this PR, `Interval{0,5}` could express this case. With the one-sided `Comparison` shape it cannot, so this is a regression.
   - Fix: fold every applicable clause into one `NodeKind::Connective(Connective::And, ..)` tree. Call `established_field_fact` once on that tree. Have it return both presence and interval, not `Presence` *or* `Interval`.
   - Add a test for the two-clause case.

2. **High: the R01 cycle listing includes types that are not in the cycle.** Location: `src/model/normalize.rs:433`.
   - `frame.visited` is the whole path from the root type, not just the cycle.
   - Probe: `A→C`, `C→B`, `B→C`.
     - Actual listing: `[model.A, model.C, model.B]`.
     - Expected listing: `[model.B, model.C]`.
   - Fix: first `drain(..position(ancestor_key))`, then rotate. Use `min_by_key(|(_, k)| *k)`; the `.clone()` is not needed.
   - The r01 test also cannot fail if the rotation is removed. `sorted_type_keys` walks `model.A` first, so `visited` is already `[A, B]`.
   - Add the probe above as r01b. Its prefix-plus-rotation case actually exercises the rotation.

3. **Medium: bad caller input causes a panic.** Location: `src/model/conformance.rs:211`, the `expect`.
   - `check_field_refinement_obligation` is public and takes a raw `Bundle`. Nothing validates that `ScalarTypeRecord.lower <= upper`.
   - Probe: `Count` set to `(9, 0)` panics at line 666 of the file.
   - Fix: return a `ModelRefusal`, or treat the domain as `None`. Add a test.

4. **Medium: R07's second shape follows the old vendored TC-196, not origin/main.**
   - origin/main R07 (after QSpec #86) uses two redefiners, `B/z` and `B/z2`, both redefining `A/x`. It expects "the same refusal listing `B/z`, `B/z2` and the one target `A/x`".
   - The PR implements and tests the vendored shape instead: one redefiner with two targets.
   - Keeping a single `redefinition-target` cause is correct under either version ("the same refusal"). `valid_targets` is a reasonable answer to QSL #146.
   - But the origin/main second shape is neither refused as `redefinition-target` nor tested. Phase 4 gives either `derivation-conflict` or a silent winner.
   - Fix: implement and test the origin/main shape here, or ticket it together with #6.

5. **Low: the R06 refusal does not name the record.** Location: `src/model/population.rs:646`.
   - The vendored R06 says "naming the record". Add `record.key.identity` to `detail`.
   - `field_values: Vec<MemberFieldValues>` silently accepts duplicate `field` entries, because `find` keeps the first. Use a `BTreeMap<ProducerKey, Vec<String>>`, or refuse duplicates.

6. **Low / repo-wide (ticket): the vendored QSpec snapshot is stale.**
   - `resources/complete-value` is pinned at `7d7943ab`, which is before QSpec #86.
   - R01 in the snapshot lists records; origin/main lists types. R07 differs as described in #4.

### Verification of the claimed fixes

- **R08: partly real.**
  - The derivation is genuine. It seeds the redefined parent's declared domain and runs the same `outcomes`/`relate` code that FR-146 uses. r08e depends on that seed, and r08g would fail under a "trust any clause" implementation.
  - The clause is still written by the caller: it is a typed copy of the source, not the parsed `post QC {...}` text. That is acceptable because `crate::model` has no parser.
  - `EstablishedFact`→`PostconditionClause` is a real shape change, from result-shaped to source-shaped, so it is not a churn rename.
  - Presence derivation is essentially 1:1.
- **R06: correct.**
  - Charges are 3 × `binding.member` + 1 × `binding.subset-value` = 4 work units.
  - Charges come before the test and follow the order in value-accounting.md. `max(1, n)` is correct.
  - The refusal is `invalid_runtime_input`/`subsetting-violation`.
  - `field_values` is justified: R06's population carries `all`/`some` values.
- **R01:** shape matches origin/main, but see #2.

### Coverage

This is a pre-existing gap across the repo, not specific to this PR.

- QSL has no row that binds QSpec-owned TC-180..198 or FR-15x ids.
- TM-010 (`spec/complete-v1/tests.md`) says those "remain pending product tests". No vendored matrix binds them.
- The only matrix that declares these rows is the central QSpec TM-009 (`V1-TYPE-026/027` → TC-196, `planned`).
- This PR's tags are formatted correctly.
- Recommendation: ticket a TM-010 extension, or a vendored TM-009 row subset, that declares FR-151 AC rows → TC-196. With that, the existing `#[trace]` tags bind with no change to the tests.

### Scope

- `key.rs` is untouched.
- The normalize.rs change is in phase 3 only (`ancestor_paths`).
- The bundle.rs change is limited to the clause type.

kreneskyp:
## Re-review of PR #157 at head 6adb4d1: **FIX**

This re-review is scoped to the diff `50bf2ba..6adb4d1`. The normative source is QSpec origin/main: FR-151, FR-272 and TC-196.

**Gates at 6adb4d1 all pass:** `fmt --check`, `clippy --locked --workspace --all-targets --all-features -D warnings`, and `test --locked --workspace --all-features`.

**Revert probes (all passed):** I reverted each fix one at a time and re-ran its test, then restored the tree.

| Test | Fix reverted | Result |
| --- | --- | --- |
| r01b | the `drain` call | fails |
| r07b | the contention block | fails |
| r08h | the And-fold | fails |
| r08i | the refusal, replaced with a panic | fails |

**Outside `src/model`:** nothing changes behaviour. `established_field_fact` and `Established` are `pub(crate)`, and they have no caller outside `model::conformance`. The only other change is re-exporting `Connective`.

### Findings

1. **High: R07's second shape still refuses with the wrong cause at the real boundary.**
   - The bundle is origin/main's R07 second shape: `B/z` and `B/z2` both redefine `A/x`.
   - `normalize` refuses it as `derivation-conflict` ("type model.B has 2 undominated redefinitions of model.A.x"). TC-196 R07 expects `invalid_model_binding`/`redefinition-target`.
   - The new check exists only in `resolve_redefinition_target`, and nothing calls that function (no caller in `src/`). r07b tests only that standalone query, so it can pass while `normalize` gives the wrong answer.
   - Fix, in phase 4 (`normalize.rs` ~1023): when the undominated edges share one owner (sibling redefiners), refuse `redefinition-target`. Keep `derivation-conflict` for edges with different owners.
   - Add an r07 test that goes through `normalize`.

2. **Medium: the R07 refusal names the wrong keys.**
   - FR-272 says `redefinition-target` carries "every redefining member's declaration key and the resolved targets". R07 says "listing `B/z`, `B/z2` and the one target `A/x`".
   - The contention refusal and r07b's assertions use the redefinition record keys instead (`model.redef.z`, `model.redef.z2`).
   - Fix: list the redefining members (`model.B.z`, `model.B.z2`).

3. **Medium: `invalid-scalar-domain` is not a cause name the spec defines.** Location: `conformance.rs` ~696, `field_domain_type`.
   - FR-272 has a closed cause list for `invalid_model_binding`. The fitting cause there is `malformed-declaration`, whose detail is the declaration and its invalid member path.
   - Fix: use `malformed-declaration` and name the scalar type's key and `lower`/`upper`. Update r08i.

4. **Medium: `duplicate-field-values` is not a cause name the spec defines.** Location: `population.rs` ~653.
   - FR-272's causes for `invalid_runtime_input` already include `duplicate-member`.
   - Fix: use `duplicate-member` and update the test.

5. **Low: r08h's doc comment is wrong, and the test relies on clause order.**
   - The comment says neither clause alone establishes the interval. But with `A.c` as `Count [0,9]`, `cs <= 5` alone gives `[0,5]`; that is exactly r08e.
   - The revert probe fails only because `cs >= 0` comes first. Swap the two clauses and the old first-clause-wins code passes.
   - Fix: seed a domain that neither clause bounds by itself, for example `Int[-5,9]`, and correct the comment.

### Checked and correct

- r01b.
- The R06 detail now names `record.key`.
- The `Established { presence, interval }` struct.
- The `min_by_key` cleanup.

kreneskyp:
## Round 3 re-review of PR #157 at head a976344: **FIX**

Only small changes are needed; there are no correctness blockers in the code. This review covers the diff `6adb4d1..a976344`. I ran it in my own clone.

**Gates at a976344 all pass:** `fmt --check`, `clippy --locked --workspace --all-targets --all-features -D warnings`, and `test --locked --workspace --all-features` (107 suites ok).

**Revert probes:** I reverted each fix on its own and re-ran its test. Every test failed when its fix was removed.

| Fix reverted | Test that fails |
| --- | --- |
| Phase-4 `same_owner` refusal | `r07_…_through_normalize` |
| Detail built from `record_key` instead of `redefining` | the same test |
| `malformed-declaration` changed back | r08i |
| `duplicate-member` changed back | `r06_duplicate_field_values…` |
| The And-fold | r08h, which now uses `Int[-5,9]` and needs both clauses in either order |

The two causes, `malformed-declaration` and `duplicate-member`, are both in FR-272's closed list.

### Findings

1. **Medium: the row claim goes beyond what the code delivers, and the PR body is out of date.**
   - The body still says "`normalize.rs`: only phase 3 … changed". It still describes R07 through `valid_targets`, and it still lists `r07b`. It also claims V1-TYPE-026/027 are backed.
   - Narrow the claim as described in the section below, and file the ticket.
2. **Low: a mix of same-owner contention and an ancestor redefiner falls through to the wrong cause.** Location: `normalize.rs:1042`, the `same_owner` check.
   - The check requires *every* undominated edge to share one owner.
   - Example: `C ≤ A` with `C/w redefines A/x`, and `B ≤ C` with both `B/z` and `B/z2 redefines A/x`. B's edge set is {C, B, B}. No edge wins, because B never dominates B, so the mixed set reports `derivation-conflict`. The real ambiguity is the same-owner contention in B.
   - Fix: run the same-owner test on the edges whose owner no other edge's owner properly dominates. Add this case to the test.
3. **Low: one rule is now implemented in two places, and they disagree.**
   - `conformance::resolve_redefinition_target` returns `Resolved(A/x)` for `B/z` on the R07 bundle, while `normalize` refuses the same bundle.
   - Fix: document that `resolve_redefinition_target` covers only R07's first shape, or delete it if nothing needs it once phase 4 owns R07.

### Ruling on the "unwired conformance" question

**Current state:**
- `model::conformance` is public (`pub mod model`, `pub mod conformance`), and its tests go through the public API.
- Nothing in `src/` outside `model` calls anything in `model`. That includes `normalize`, `dispatch` and `admit_binding`. `model/mod.rs` documents this: the module is pure functions over a caller-built `Bundle` until QSpec #131 / FCD #173 intake lands.
- So for this rung, `normalize` is the model boundary. The conformance checks are public but not *composed into* that boundary.
- Concretely, `normalize` admits a bundle that has a variance-failing redefinition or an unproved refinement.

**What the spec requires:**
- FR-151 has one Output: "Validated effective declarations and a linked dispatch table, **or** a located cycle, incompatibility, unproved refinement or ambiguity refusal."
- TC-196 runs every vector "through the real local Rust boundary; compare the complete typed result and artifact accounting".
- R02–R04 expect the `conformance.axis` charges and per-axis work units *inside* the normalization result. `conformance.rs`'s own module docs say it does not reproduce `f(T)`.
- FR-151-AC-2 (incompatible variance), AC-4 and AC-6 are therefore met only as standalone rules, not by the FR-151 Output.

**Recommendation: narrow the claim and file a ticket. Do not wire it in this PR.**
- Wiring the checks means composing the axes into phase 4 with exact `f(T)` accounting. That is a slice of its own, and larger than this one.
- Change the body to say: FR-151 R01, R06, R07 and R08 implemented as rules and tested on TC-196 vectors; R01, R07 (second shape) and R06 go through their real boundaries (`normalize` / `admit_binding`); V1-TYPE-026/027 remain in progress.
- File the ticket under #120: "compose conformance variance/refinement/subsetting checks into `normalize` phase 4 with `conformance.axis` `f(T)` accounting (TC-196 R02–R05, R07 first shape, R08)."
- With #1 done, this merges. #2 and #3 are quick in-PR fixes.

kreneskyp:
## Delta check of PR #157 at 847e1e17fa3eed5420157577dc65166f4d9a22ca: **MERGE**

This check covers the delta `a976344..847e1e1`. I ran it in my own clone.

- **Gates at 847e1e1 all pass:** `fmt --check`, `clippy --locked --workspace --all-targets --all-features -D warnings`, and `test --locked --workspace --all-features` (107 suites ok).
- **Most-derived-owner fix (`normalize.rs` ~1053):**
  - I reverted the restriction by skipping the `dominates` check. `r07_a_less_derived_owners_redefiner_is_excluded_from_the_same_owner_test` then fails. The N06 tests and the same-owner R07 test still pass.
  - `dominates` is a proper-descendant test and is never reflexive, so the filter is correct.
- **`resolve_redefinition_target` doc:** it now states that it covers R07's first shape only, and points to phase 4 and #165. Resolved.
- **PR body:** V1-TYPE-026/027 are stated as in progress, not backed. It says `Refs #165` (#165 is open), and it no longer mentions r07b. `closingIssuesReferences` is `[]`, as intended: the body only uses "Part of #120" and "Refs".

All findings from earlier rounds are resolved.

PR #158, opened 2026-09-18
A04 slice 1a: model references — bridge, allInstances, closed lookup

kreneskyp:
## Summary

Part of #120 (A04). Adds three of the five originally-scoped items for model references in complete-V1 expressions:

- `EffectiveId` → `ObjectReference` bridge (`src/value/model_query.rs`)
- `allInstances<T>(p)` and `lookup<T>(p, r) absent m` as complete-V1 source forms (new `Expression`/`NodeKind` variants, typer dispatch, evaluator `apply()` arms)
- Accounting charges exactly per FR-153 (two independent meters: `AdmissionMeter` for binding admission, `Meter`/`ScalarMeter` for query evaluation)

Backed by FR-153 (closed model lookup) and FR-143 (reference identity triple: `(universe, most-specific-type, object-identity)` — a `Value::Reference`'s `object_type` always carries the object's own actual most-specific type, never the queried static type `T`).

### Rows moved

- V1-TYPE-023 (closed lookup) — **partial**, see below.

V1-TYPE-021 and V1-TYPE-022, listed as moved in an earlier revision of this PR, are not: V1-TYPE-021 maps to FR-011, which this slice never exercises; V1-TYPE-022 depends on TC-198 L08 (upcast equality *operator* semantics), deferred below.

**V1-TYPE-023 partial coverage:**
- Covered: AC-1 (closed population), AC-2 (per-mode presence), AC-3 (accounting, the exact work/result totals, and the foreign-universe/type-mismatch refusal cases at the expression layer — `foreign-type` is admission-layer only, `tests/model_population.rs` L05, pre-existing and unaffected by this slice), AC-4 (typed Outputs), AC-5 (bounded typed result), AC-6 (upcast reference-identity equality — `b1`'s reference is identical whether selected under `M::A` or queried directly under `M::B`, `l12_...`).
- Missing: TC-198 L08 (upcast equality *operator* semantics) and AC-7 (`pre()`), plus `lookup<T>(p, r)`'s check-time refusal when `r`'s declared type does not conform to `T` — all three deferred, see below.

### Tests

`tests/model_reference_queries.rs`, 17 tests, exercised through the real `CheckedPackage::check_expression`/`evaluate` boundary (no textual parser in this language):

- `l12_all_instances_expression_selects_subtype_population_once` (AC-1, AC-5, AC-6)
- `l13_all_instances_expression_incomplete_under_a_low_work_limit` (AC-3) — pins the exact 5 work-unit / 4 result-unit totals and the low-limit denial's exact charge point (`CollectionResultRetain`, `consumed: 4`, `next_charge: 1`), matching `tests/model_population.rs`'s L01 at the model layer.
- `l14_lookup_expression_undefined_mode` (AC-2, AC-4) — asserts `Undefined::AbsentKey` specifically.
- `l15_lookup_expression_empty_mode` (AC-2, AC-4)
- `l16_lookup_expression_refused_mode` (AC-2, AC-4)
- `population_refused_as_equality_operand`, `population_refused_as_option_payload`, `population_refused_as_collection_element`, `population_refused_as_record_field`, `population_refused_as_object_attribute` — first-review finding 1.
- `all_instances_expression_target_declared_but_not_in_model_is_type_mismatch` (AC-3) — first-review finding 2.
- `lookup_expression_malformed_universe_is_foreign_universe_after_one_work_unit` (AC-3), `lookup_expression_malformed_identity_is_none_in_empty_mode` — first-review finding 3.
- `all_instances_refuses_non_population_operand_as_ineligible`, `lookup_refuses_non_population_operand_as_ineligible` — first-review finding 5.
- `lookup_expression_malformed_identity_never_aliases_a_lossy_decoded_member` — second-review finding 1: regression repro for a real member whose identity is `"\u{FFFD}\u{FFFD}"`, looked up via a malformed reference whose identity bytes are `[0xFF, 0xFE]`, pinning absence in all three modes.
- `lookup_expression_inside_a_set_literal_keeps_the_most_specific_element_type` — second-review finding 3: `set[lookup<M::A>(p, r) absent refused]` keeps a `B`-typed element inside `Set<Reference<M::A>>`.

These continue `tests/model_population.rs`'s existing L01–L08 TC-198 coverage.

### Real bugs found and fixed

- `lookup<T>(p, r) absent empty`'s present-branch Option-wrapping used `OptionValue::present`, whose `admits()` check is exact object-type equality with no model-conformance knowledge. It wrongly refused every genuine FR-153 upcast (a subtype object presented as its declared supertype `T`) even though `crate::model::population::lookup` had already verified conformance before returning. Fixed with a new `OptionValue::from_admitted` uncharged constructor (`src/value/composite.rs`), mirroring the existing `collection::from_admitted` pattern already used by `allInstances` for the same reason. Its doc comment states the exact soundness chain: admission guarantees the argument's runtime most-specific type `F` conforms to (not equals) its declared static type `S` — `S` may itself be the result of an earlier upcast lookup, e.g. `lookup<A>(p, lookup<A>(p, rb) absent refused) absent empty` has `S = A`, `F = B`; `lookup`'s own `type_conforms(S, T)` call then gives `S` conforms to `T`; therefore `F` conforms to `T` transitively.
- Second review (2026-09-18): the identity-bridging fallback this PR originally shipped (a lossy UTF-8 decode of a malformed reference's identity bytes, substituted and delegated to the real `lookup`) was unsound — `String::from_utf8_lossy(&[0xFF, 0xFE])` is `"\u{FFFD}\u{FFFD}"`, a value a population is free to admit as a real member's own identity, so a producer-supplied malformed reference carrying those same bytes was mistaken for that member, wrongly returning "present" in all three absence modes. Fixed by removing all substitution: `src/value/model_query.rs`'s `bridge_reference` now reports which identity-triple component (if any) cannot be losslessly bridged, and `evaluate_unresolvable_lookup` decides the outcome directly — `type_conforms(S, T)` first, one `lookup.key` charge, then `foreign-universe` (reporting the caller-supplied bytes, never a derived value) for a malformed universe, or the mode's own absence outcome for a malformed identity, with `empty` mode's extra `lookup.result-retain` charge before returning `none` — mirroring `crate::model::population::lookup`'s own ordering without ever constructing a key that could alias a real member. New regression test above.

### Deferred (documented, not silently dropped)

Tracked at #164 ("`value`'s `TypeEnvironment` island: no model generalization data at check time"), the root cause behind all three:

- TC-198 L08 (upcast equality *operator* semantics) — `ValueType::admits()` for `Reference` is exact structural match only; teaching it model conformance is a larger change than this slice. (FR-153-AC-6 itself, reference-identity equality under an upcast, is covered — see `l12_...` above; L08 is a distinct, still-deferred case.)
- Replacing `deref(r).f` display-name resolution with identity-based resolution.
- `lookup<T>(p, r)`'s check-time refusal when `r`'s declared static type does not conform to `T` (first-review finding 4) — the checker cannot decide this without the model's own generalization graph, which `TypeEnvironment` does not carry; today it is refused only at evaluation, inside `crate::model::population::lookup`'s own `type_conforms` call (or, for a malformed reference, the equivalent direct check in `crate::value::model_query::evaluate_unresolvable_lookup`).

### Explicitly out of scope for this PR

`spec/tests.md`, `spec/complete-v1/tests.md`, and `plan/Plan-013-complete-v1-delivery/plan.md` / `Task-050-model-graph-semantics.md` are shared planning artifacts being edited concurrently by other authors on the same task (#131 shares `Task-050`). Not touched here; the real per-AC evidence is the new `#[trace(...)]`-tagged tests plus `quire coverage`.

#157 (`task/120-conformance`) adds a required `PopulationMember::field_values` field. If #157 merges first, this branch will need a rebase adding `field_values: vec![]` to this file's `member()` test helper.

### Review response (PR review, 2026-09-18)

All ten findings from the first review fixed in this PR:

1. **`ValueType::Population` admitted as an ordinary type (HIGH)** — fixed. `TypeEnvironment::type_refusal` (`src/value/composite.rs`) now refuses `Population` everywhere except as the direct `allInstances`/`lookup` operand and a bare parameter's own declared type (`bind_parameters` bypasses the walk for exactly that case). `TypeEnvironment::check_equality` (`src/value/equality.rs`) refuses it as an equality operand directly, before the `l == r` plan schedule. Five new tests.
2. **`resolve_target` used `invariant()` for an unrecognized target (HIGH)** — fixed. `src/value/model_query.rs::resolve_target` now returns `ill_typed`/`type-mismatch` when a package-declared target type is absent from the binding's own model, the same cause `crate::model::population::all_instances`'s `is_object_type` gives. New test.
3. **`from_object_reference` converted before charging (MED)** — fixed in this PR's first response by bridging malformed identity components into deterministic, provably-foreign values and delegating to the real `crate::model::population::lookup`. The identity half of that fix turned out to be unsound; see "Real bugs found and fixed" and the second review response below for the actual fix shipped in this PR.
4. **`lookup<M::B>(p, ra)` accepted at check time (MED)** — deferred, tracked at #164 (see above); the checker cannot decide `r`'s static-type conformance without the model's generalization graph, which `TypeEnvironment` does not carry.
5. **Non-population operand refused as `type-mismatch` (MED)** — fixed. Both `all_instances`/`lookup` in `src/value/expression/check.rs` now refuse `operator-ineligible` per TC-198 L06. Two new tests.
6. **L13/L14 tests too loose (MED)** — fixed, see Tests above; also added the AC-3 foreign-universe/type-mismatch cases at the expression layer.
7. **`OptionValue::from_admitted` comment wrong; collection literals skip `admits()` (LOW)** — comment fixed (see "Real bugs found and fixed" above; corrected again in the second review response for the `F == S` vs. `F` conforms to `S` distinction). The collection-literal note is now covered by a pinning test (`lookup_expression_inside_a_set_literal_keeps_the_most_specific_element_type`), added in the second review response.
8. **`population.rs` module docs stale (LOW)** — fixed; docs now describe this PR's bridge instead of denying one exists.
9. **Hex round-trip `.expect()`; double linear catalog scan (LOW)** — fixed. New `NodeKey::from_bytes` (mirrors `EffectiveId::from_digest_bytes`) removes the panic surface; a new `reverse_catalog` builds the reverse type-catalog index once per query instead of scanning linearly per resolved type.
10. **Prose ids in test doc comments (LOW)** — fixed; bare `AC-4`/`AC-5`/`AC-6` fully qualified to `FR-153-AC-N`, bare `FR-143` mentions reworded.

### Second review response (PR re-review, 2026-09-18)

All four findings from the re-review fixed in this PR:

1. **Lossy UTF-8 identity fallback unsound (HIGH)** — fixed. See "Real bugs found and fixed" above: `bridge_reference` takes no substitution for either identity-triple component; `evaluate_unresolvable_lookup` decides `type_conforms(S, T)` first, charges `lookup.key`, then dispatches directly to the mode's absence outcome (with `empty` mode's extra `lookup.result-retain` charge) for a malformed identity, and to `foreign-universe` (reporting the caller-supplied bytes, not a fabricated complement) for a malformed universe — the universe case is now handled the same direct way as the identity case, for consistency, rather than keeping its (already-sound) bit-complement substitution. New regression test `lookup_expression_malformed_identity_never_aliases_a_lossy_decoded_member` covers all three absence modes.
2. **`OptionValue::from_admitted` doc stated `F == S` (LOW)** — fixed. The doc now states `F` conforms to `S` (never equality), with the `lookup<A>(p, lookup<A>(p, rb) absent refused) absent empty` example (`S = A`, `F = B`) as the concrete case `F == S` would get wrong.
3. **No pinning test for the collection-literal upcast (LOW)** — fixed. New test `lookup_expression_inside_a_set_literal_keeps_the_most_specific_element_type` pins that `set[lookup<M::A>(p, r) absent refused]` produces a `B`-typed element inside `Set<Reference<M::A>>`, per FR-143/FR-153 and TC-198 L01.
4. **PR body claims (LOW)** — fixed. "Deferred: AC-6" corrected to "Deferred: TC-198 L08" (AC-6 itself is covered by `l12_...`); the AC-3 "foreign-type" claim corrected to say it is covered only at the admission layer (`tests/model_population.rs` L05), not by this slice's expression-layer tests (which cover `foreign-universe` and `type-mismatch`).

## Test plan

- [x] `cargo fmt --check`
- [x] `cargo clippy --workspace --all-targets --all-features -- -D warnings` — clean
- [x] `cargo test --workspace --all-features` — 0 failed
- [x] `make revendor-check` — 69 + 64 matched, 0 drifted, 0 stray
- [x] New/changed tests pass standalone: `cargo test --test model_reference_queries --all-features` — 17 passed
- [x] Reverted `bridge_reference`'s identity check back to the old lossy-decode substitution and reran the new regression test standalone: it fails (`Completed(Reference(...))` — the malformed reference wrongly matched the `"\u{FFFD}\u{FFFD}"` member), confirming the test pins the fix; reverted back to the fix and confirmed all 17 tests pass again


kreneskyp:
## Review of PR #158 at `5ea93b3` (single Opus reviewer: code review, Rust review, gap analysis)

**Verdict: FIX.**

The bridge, the charges and the `lookup` modes are correct on the happy path. A throwaway probe through `CheckedPackage::check_expression`/`evaluate` measured exact TC-198 accounting:

- L01: `allInstances<M::A>(p)` uses 5 work units and 4 result units, and completes at `work_units: 5`.
- L03 `undefined` present: 2 work units, 1 result unit.
- L03 `empty` present: 2 work units, 2 result units.
- L04 foreign 32-byte universe: `foreign_reference`/`foreign-universe` after one work unit.

Nothing is double-charged. But a type-checked program with valid arguments can reach `Refusal::CheckedInvariant`, which the code documents as "unreachable for an admitted program", through four routes. Two of the claimed rows are also not backed.

### Gates (all green)
- `cargo fmt --check`: 0
- `cargo clippy --locked --workspace --all-targets --all-features -- -D warnings`: 0
- `cargo test --locked --workspace --all-features`: 0 failed
- `make revendor-check`: native-v1 69 matched, complete-value 64 matched, 0 drifted, 0 stray

### Findings

1. **High: `src/value/composite.rs:67` (`ValueType::Population`), with `composite.rs:566/607` and `equality.rs:333`.**
   - **Defect:** `Population[N]` is admitted as an ordinary value type everywhere: equality operands, `Option`/collection payloads, record fields and object attributes. `check_equality` treats `l == r` as `EqualitySchedule::Plan`, and `plan_pairs` then returns `CheckedInvariant`.
   - **Probe:** `p == p` with `p: Population[3]` checks as `Boolean` and evaluates to `Refused(CheckedInvariant)`. `Set<Population[3]>` and `Option<Population[3]>` also pass `type_refusal`.
   - **Fix:** FR-153 names a population binding only as the operand of `allInstances`/`lookup`. Refuse `Population` everywhere except as a parameter type used as one of those operands, or at least in equality, nesting and declaration member types, with `ill_typed`/`operator-ineligible`.

2. **High: `src/value/model_query.rs:93-101` (`resolve_target`).**
   - **Defect:** the checked `T` (and lookup's `S`) is resolved against `binding.type_catalog()` at run time. The checker never ties the package's `TypeEnvironment` to the binding's model, so a `T` that the package declares but the binding's model does not is reachable.
   - **Probe:** package declares object type `X::Z`, then `allInstances<X::Z>(p)` evaluates to `Refused(CheckedInvariant)` with 0 work units.
   - **Fix:** FR-153 requires `T` to be a model object type of the binding's selection, otherwise `ill_typed`/`type-mismatch`. The model layer already emits that cause (`population.rs` `all_instances`, `is_object_type`), so the bridge should return it instead of `invariant()`.

3. **Medium: `src/value/model_query.rs:71-80,148` (`from_object_reference`).**
   - **Defect:** the conversion runs before `lookup` charges `lookup.key`. A universe that is not 32 bytes, or an object identity that is not UTF-8, returns `CheckedInvariant` with no charge. Both arguments pass parameter admission, because `UniverseIdentity::new`/`ObjectIdentity::new` accept any non-empty bytes.
   - **Probe:** `lookup<M::A>(p, r) absent empty` with a 5-byte universe gives `CheckedInvariant` with 0 work. Per FR-153 it should be `foreign_reference`/`foreign-universe` after `lookup.key` (1 work unit). With a non-UTF-8 identity it also gives `CheckedInvariant`; that key can never be a member, so `empty` mode should return `none` after 2 work units.

4. **Medium: `src/value/expression/check.rs:1500-1520` (`lookup`).**
   - **Defect:** FR-153 says `r: Reference<S>` with `S` not conforming to `T` is `ill_typed`/`type-mismatch` (TC-198 L03, last case: "before any charge"). The checker admits it; the refusal only appears at run time.
   - **Probe:** `lookup<M::B>(p, ra) absent empty` checks OK as `Option<Reference<B>>` and evaluates to `Refused(Model{IllTyped, "type-mismatch"})` with 0 work. The code and cause match, but the disposition is an evaluation outcome, not a `CheckRefusal`. The PR does not list this among its deferrals, and no expression-layer test covers it.
   - **Fix:** either list it as a third "island" deferral, or have the typer refuse it.

5. **Medium: `src/value/expression/check.rs:1479-1480`.** `allInstances<T>(x)` with a non-population `x` gives `type-mismatch`. FR-153 says "requires `p` to be a population binding with a declared maximum `N`, otherwise `ill_typed`/`operator-ineligible`" (TC-198 L06, third case). The same applies to `lookup`'s population operand at `check.rs:1513`.

6. **Medium: `tests/model_reference_queries.rs:422` (L13) and `:485` (L14), test strength.**
   - L13 accepts any `Outcome::Incomplete(_)` at `work_units: 4`. A double charge in the bridge (for example `collection::from_admitted` charging `collection.result-retain` again) would still pass. No expression-layer test pins L01's 5/4 totals or the `collection.result-retain` charge point, so the PR's claim "accounting charges exactly per FR-153" is unbacked at the layer this PR adds (the probe shows it is currently true).
   - L14 accepts any `Undefined(_)`, not `Undefined::AbsentKey`.
   - AC-3 is tagged, but its foreign-universe and foreign-type halves have no expression-layer test; the path in finding 3 is where they break.

7. **Low: `src/value/model_query.rs:158-172`, the `OptionValue::from_admitted` justification.**
   - **Soundness:** sound today. It is `pub(crate)`, its only caller is `evaluate_lookup`, and the `Some` branch runs only after `LookupOutcome::Completed(Some(_))`.
   - **Inaccurate comment:** the comment says `lookup` "proved `found` conforms to `T` via `type_conforms`". It did not. `population::lookup` checks only `type_conforms(S, T)` on the caller-declared static type. The found key is `r`'s own key (via `contains_key`), so conformance of the payload to `T` follows from "runtime type of `r` ≤ `S`", an invariant held by exact-type parameter admission and by the `allInstances`/`lookup` results. No caller can currently reach it with an unverified value.
   - **Fix:** correct the comment to name that invariant.
   - **Related (not a defect of this PR):** a collection literal also skips `admits()`. The probe showed `set[lookup<M::A>(p, rb) absent refused]` completing with a `B`-typed element in `Set<Reference<A>>`. Upcast values now flow freely, and any constructor that does call `admits()` (`OptionValue::present`, and possibly function-argument binding) will refuse them. That is the same island seen from the other side.

8. **Low: `src/model/population.rs:45-60`.** The module docs still say "this crate defines no mapping from `crate::model`'s `EffectiveId`/`ProducerKey` identities into that byte space anywhere". This PR adds exactly that mapping (`model_query.rs`), so the doc is now stale.

9. **Low (Rust idiom): `src/value/model_query.rs:59-62` and `:93-101`.**
   - `to_object_reference` round-trips the 32 bytes through `hex()`/`NodeKey::from_hex` with an `expect` in library code. A `pub(crate) NodeKey::from_bytes([u8; 32])` would mirror `EffectiveId::from_digest_bytes` and remove the panic surface.
   - `resolve_target` scans the whole catalog linearly, twice per `lookup`. A reverse `EffectiveId -> ProducerKey` map built at admission is O(log n).

10. **Low: `tests/model_reference_queries.rs` doc comments.** `quire coverage` reports prose ids in these doc comments as `unmatched_tags` (`AC-5`/`AC-6` at :328-329, `AC-4` at :430 and :492, `FR-143` at :559). Write them as `FR-153-AC-5` or rephrase.

### Overlap with #157 (`task/120-conformance`, head `50bf2ba`)
- **Textual:** a trial `git merge-tree` of both heads shows **no conflict markers**. Both touch `population.rs`, `value/mod.rs` and `value/expression/mod.rs`, in disjoint hunks. #157's subsetting walk sits just above #158's `type_catalog: type_lookup` line in the `PopulationBinding` literal.
- **Semantic break:** #157 adds a required public field `PopulationMember::field_values`. The `member()` helper in #158's `tests/model_reference_queries.rs` builds `PopulationMember { object, type_identity }`, so whichever PR merges second fails to compile its tests until it adds `field_values: vec![]`.
- `key.rs` and `normalize.rs`: no overlap. #158 does not touch `normalize.rs`, and #157 does not touch `key.rs`.

### Deferrals
- **FR-153-AC-6 / L08 upcast equality.**
  - **Real:** yes. `ObjectTypeDeclaration`/`TypeEnvironment` carry no generalization data, so `check_equality` can only compare `Reference<A>` and `Reference<B>` structurally. The probe confirms `(lookup<M::A>(p, rb) absent refused) = rb` is refused at check time as `ill_typed`/`type-mismatch`.
  - **AC-6 itself is backed:** its statement (one reference key whose type component is the most-specific type) is backed by L12's triple comparison. The deferral is really of TC-198 L08, the FR-149 upcast.
  - **Row dependency:** V1-TYPE-022 is "explicit typed object references and **identity equality**" (QSpec `docs/v1-capability-inventory.md`, plan row FR-009/TC-231). It depends on exactly this, and no FR-009 AC (unequal identities, pre/post, foreign universe) is exercised here.
  - **Ruling:** deferral acceptable, but **claiming V1-TYPE-022 is an overclaim**. Fixing it is a type-system change (teaching `TypeEnvironment` generalizations), which is out of scope for this slice.
- **`deref(r).f` display-name resolution.**
  - **Real:** yes. `ObjectEnvironment::attribute` looks up the runtime type's declaration and resolves the attribute by name. With upcast values now producible, a `Reference<A>` that holds a `B` resolves `x` in `B`'s declaration, which has no inherited members.
  - **Row dependency:** none of V1-TYPE-021/022/023 depends on it (it is TC-198 L11, FR-043 graph territory).
  - **Ruling:** deferral acceptable, not needed in this PR.
- **Undeclared deferral:** the island also causes findings 2 and 4 (the checker cannot see the binding's model or conformance). Map findings 2 and 3 to their named refusals in this PR; they are cheap. Finding 4 can be listed as a deferral.

### Row claims
- **V1-TYPE-023 (FR-153/TC-198):** partly moved.
  - Done: AC-1, 2, 4 and 5 at the expression layer.
  - Weak: AC-3 (finding 6).
  - Not done: AC-7 `pre()`, L08, and the L03 check-time type-mismatch.
- **V1-TYPE-021:** not moved. It maps to FR-011 (validate populations: `duplicate_object`, `dangling_reference`, `incomplete_population`, ...), which no test in this PR exercises.
- **V1-TYPE-022:** not moved; see the L08 ruling.

### Trace binding
`quire coverage --scope . --json` (rc 0; stderr carries only module warnings: `DuplicateArchetype`, `inline-data-schema`). All 15 `#[trace("TC-198", "FR-153-AC-N")]` tags in the new file are parsed, and all land in `untracked_symbols`, the same as the existing `tests/model_population.rs` L01-L08. QSL declares no FR-153/TC-198 rows (tracked as #160), so this is not held against this PR. The only new noise is finding 10.

**Required before merge:**
- findings 1, 2 and 3;
- finding 5;
- pin the charges in L13 and the `AbsentKey` reason in L14 (finding 6);
- drop V1-TYPE-021/022 from "rows moved" and state V1-TYPE-023 as partial.

Findings 7-10 are small and should be fixed in the same cycle.

kreneskyp:
## Re-review of PR #158 at `fc8cb8366e08c617c09a9bf5bd08efc17874db5d`

**Verdict: FIX.** One High finding: the finding-3 fix substitutes a lossy UTF-8 conversion, and a probe shows it returns a *different* object as a present lookup result. Everything else from the first review is fixed or correctly deferred.

### Gates (all green at `fc8cb83`)
- `cargo fmt --check`: 0
- `cargo clippy --locked --workspace --all-targets --all-features -- -D warnings`: 0
- `cargo test --locked --workspace --all-features`: 0 failed (108 ok result lines)
- `make revendor-check`: native-v1 69 matched, complete-value 64 matched, 0 drifted, 0 stray

### Revert-probes (each fix reverted in turn, then `--test model_reference_queries` run)
| Fix reverted | Result |
| --- | --- |
| 1: `type_refusal` `Population` arm | 4 failed (option, collection, record, object attribute) |
| 1: `check_equality` `Population` arm | 1 failed (`population_refused_as_equality_operand`) |
| 2: `resolve_target` returns `invariant()` again | 1 failed (`all_instances_expression_target_declared_but_not_in_model_is_type_mismatch`) |
| 5: `ineligible` changed back to `mismatch` (both sites) | 2 failed |
| 6: `Undefined::AbsentKey` changed to `NoneValue` | 1 failed (L14) |
| 6: `collection.result-retain` occurrence count +1 | 1 failed (L13) |

Every new test fails when the behaviour it covers is removed. None passes by construction.

### Findings

1. **High: `src/value/model_query.rs` `bridged_object` (lossy UTF-8 fallback) is unsound.**
   - **Probe:** a population with one `model.A` member whose identity is the valid string `"\u{FFFD}\u{FFFD}"` (bytes `EF BF BD EF BF BD`), and `r: Reference<M::A>` with identity bytes `[FF FE]`. `r` is a different object that is admitted in the object environment. `lookup<M::A>(p, r)` returns:
     - `absent empty`: `some(ObjectReference{identity: [EF BF BD EF BF BD]})` after 2 work units;
     - `absent undefined`: the same reference as a bare `Reference<A>`;
     - `absent refused`: the same reference as a bare `Reference<A>`.
   - **Why it is wrong:** FR-153 requires `none`, `undefined absent-key` and `refused absent-key` respectively. Instead the query returns a *different object's* reference as if `r` were present. This is an identity confusion (FR-009/FR-143), not a rare hash collision. U+FFFD is an ordinary JSON-string character, so a producer can create such a member deliberately.
   - **Fix:** don't substitute. `r`'s identity bytes cannot name any member, because member identities are `String`s. So charge `lookup.key` (after the `type_conforms(S, T)` check, preserving FR-153's order) and then take the absent branch directly for the mode:
     - `empty`: `lookup.result-retain` with 1 occurrence, then `none`;
     - `undefined`: `absent-key`;
     - `refused`: `invalid_runtime_input`/`absent-key`.

     The cleanest place for this is a `population::lookup` entry point that takes a key known to be a non-member. Alternatively, make `ReferenceKey.object` a byte string. Add the collision case above as a regression test.
   - The universe bit-complement is sound, because the complement can never equal the binding's universe. It has two cosmetic costs:
     - the `foreign-universe` detail reports the fabricated complement as the "supplied" universe instead of the real 5 bytes (TC-198 L04 names the supplied universe);
     - it duplicates the same "substitute and let `lookup` decide" pattern.

     Fixing the refusal detail can happen in the same change.

2. **Low: `src/value/composite.rs` `OptionValue::from_admitted` doc comment.** It now states the invariant as "exact-match parameter admission guarantees `F == S`". That is false once an upcast value is used as the key.
   - **Probe:** `lookup<M::A>(p, lookup<M::A>(p, rb) absent refused) absent empty` has static `S = A` and runtime `F = B`. It correctly returns `some(b1 as B)`.
   - Soundness still holds, but through the invariant **every `Value::Reference`'s runtime type conforms to its static type** (`F ≤ S`), maintained at every entry point:
     - parameter admission: exact match;
     - `ObjectEnvironment` attributes: exact match;
     - `allInstances`: `type_conforms`;
     - `lookup`: inherits it by transitivity.
   - `F ≤ S` and `S ≤ T` give `F ≤ T`. Restate the doc comment as `F ≤ S`.

3. **Low: collection-literal `admits()` skip (the coordinator's question).**
   - **Reachable:** yes. `set[lookup<M::A>(p, rb) absent refused]` against `Set<Reference<M::A>>[0,3]` completes with a `B`-typed element.
   - **Wrong:** no. FR-143/FR-153 require a reference to carry its most-specific type whatever its static type. TC-198 L01 itself expects `allInstances<M::A>(p)` to be a `Set<Reference<M::A>>` containing `b1` as `(U, B, "b1")`. A literal that holds the same upcast value is the same shape.
   - What would be wrong is a non-conforming element. The static check (element type must equal `Reference<A>`) together with the `F ≤ S` invariant from finding 2 rules that out.
   - The PR body calls this "out of scope, no test". Since the invariant is what makes it sound, add one test that pins it: the literal above completes, and its element is `(U, B, "b1")`. Fold it into the same cycle as finding 1.

4. **Low: PR body wording.**
   - It lists AC-6 as "Missing" while L12 is tagged with, and does back, FR-153-AC-6 (one key whose type is the most-specific type). What is actually deferred is TC-198 **L08** (FR-149 upcast equality).
   - It lists AC-3 "foreign-type" as covered, but the new test covers `ill_typed`/`type-mismatch` for a target outside the model. `foreign-type` is a binding-admission refusal that stays covered at the model layer (`tests/model_population.rs` L05).

### Deferral and issue check
- **#164** exists (OPEN): "value's TypeEnvironment island: no model generalization data at check time". It is scoped correctly to the three items: L08 upcast equality, `deref(r).f` inherited-member resolution, and the check-time `S`-vs-`T` refusal (finding 4). It names the root cause and the required type-system change. Finding 4's deferral to #164 is accepted. The `check.rs` `lookup` doc and the `model_query.rs` module docs both link it.

### Row claims
- The PR body now lists only V1-TYPE-023, as **partial**, with V1-TYPE-021 (FR-011) and V1-TYPE-022 (depends on L08) explicitly removed. Correct, apart from the AC-6/L08 wording in finding 4.

### Earlier findings: status
- **Fixed and verified:**
  - finding 1: `Population` refused outside the operand/parameter positions;
  - finding 2: named `type-mismatch`;
  - finding 5: `operator-ineligible`;
  - finding 6: L13 pins 5 work / 4 result units and the `CollectionResultRetain` charge point with `consumed: 4`/`next_charge: 1`, and L14 pins `AbsentKey`;
  - finding 8: population.rs docs;
  - finding 9: `NodeKey::from_bytes` and the reverse catalog;
  - finding 10: prose ids.
- **Finding 3:** the universe half is fixed; the identity half is replaced by new finding 1.
- **Finding 7:** the comment is rewritten but states the wrong invariant (new finding 2).
- **Finding 4:** deferred to #164, accepted.
- **#157 overlap:** unchanged; still no textual conflict. The `field_values` rebase note in the PR body is accurate.

**Required before merge:**
- new finding 1: no substitution for a non-UTF-8 identity, plus a collision regression test.

**Same cycle:**
- findings 2–4.

kreneskyp:
## Round-3 review of PR #158 at `ba8e12912800635773345372f34bb0a8b72d154b`

**Verdict: FIX.** One Medium finding: the new "second copy" of `lookup`'s decision order drops `lookup`'s universe check when the identity is the malformed component. The rest of the round-3 delta is correct, and every charge matches. This review ran in a fresh https clone, not the author's worktree.

### Gates (all green at `ba8e129`)
- `cargo fmt --check`: 0
- `cargo clippy --locked --workspace --all-targets --all-features -- -D warnings`: 0
- `cargo test --locked --workspace --all-features`: 0 failed (108 ok result lines)
- `make revendor-check`: native-v1 69 matched, complete-value 64 matched, 0 drifted, 0 stray

### Revert-probe
I restored the lossy path (`String::from_utf8_lossy` in `bridge_reference`). `lookup_expression_malformed_identity_never_aliases_a_lossy_decoded_member` then **fails**, and 16 others pass. The regression test is a real gate.

### Charge comparison: unresolvable path vs the real `lookup()` path
Probe setup: `r: Reference<M::A>`, a well-formed universe, and identity `"zz"` (absent) vs `[FF FE]` (non-UTF-8).

| Case | Mode | Outcome | Work units | Value occurrences | Result units |
| --- | --- | --- | --- | --- | --- |
| well-formed absent | empty | `none` | 2 | 1 | 1 |
| non-UTF-8 identity | empty | `none` | 2 | 1 | 1 |
| well-formed absent | undefined | `AbsentKey` | 1 | 1 | 0 |
| non-UTF-8 identity | undefined | `AbsentKey` | 1 | 1 | 0 |
| well-formed absent | refused | `absent-key` | 1 | 1 | 0 |
| non-UTF-8 identity | refused | `absent-key` | 1 | 1 | 0 |

At `work_units` 0 and 1, both references stop with identical `Incomplete` records: `LookupKey` at 0/0, and `LookupResultRetain` at 1/1 with next charge 1. The order is also the same as `population::lookup`: `type_conforms(S,T)` before any charge, then `lookup.key` (1 occurrence), then `lookup.result-retain` (1/1) for `empty`. A malformed identity costs exactly what an absent well-formed one costs. A short universe gives `foreign-universe` after 1 work unit, and the detail now shows the real supplied bytes.

### Findings

1. **Medium: `src/value/model_query.rs` `evaluate_unresolvable_lookup`, the `Unbridgeable::Identity` arm.** A 32-byte universe that is *foreign*, combined with a non-UTF-8 identity, gets the absence outcome instead of `foreign-universe`.
   - **Why:** `bridge_reference` returns `Identity` without comparing the (well-formed) universe to the binding's. The `Identity` arm then goes straight to the mode's absent branch. `population::lookup` checks `r.key.universe != binding.universe()` right after `lookup.key` and before membership. FR-153 says the same: "After `lookup.key`, a key whose universe differs from the universe of `T` is `foreign_reference`/`foreign-universe`".
   - **Probe** (universe `[07; 32]`):

     | Identity | Mode | Outcome | Work units | Result units |
     | --- | --- | --- | --- | --- |
     | `"zz"` | any | `foreign-universe` | 1 | 0 |
     | `[FF FE]` | `empty` | `none` | 2 | 1 |
     | `[FF FE]` | `undefined` | `Undefined(AbsentKey)` | 1 | 0 |
     | `[FF FE]` | `refused` | `invalid_runtime_input`/`absent-key` | 1 | 0 |

   - **Effect:** a reference from another universe is silently answered as "not a member of this population". In `empty` mode it becomes a completed `none` where a refusal is required.
   - **Fix:** in the `Identity` arm, compare the (32-byte) universe with `binding.universe()` first and refuse `foreign-universe` on mismatch. Simplest: have `bridge_reference` return `Identity { universe: EffectiveId }`, or check `Universe` for any failing reference. Then add the foreign-32-byte + non-UTF-8 case to the regression test.
   - **Recommended:** move the unresolvable-key decision into `population::lookup` itself, for example as a `LookupKey` whose `object` is an enum `Member(String) | NonMember`. That leaves one owner of the order, and `pub(crate) conforms` is no longer needed. This finding is exactly the drift that a second copy invites.

2. **Pass: `population::conforms` visibility.**
   - It is `pub(crate)` over `type_conforms`, which is already `pub(super)`. It is read-only and exposes nothing beyond what `lookup` already decides, and its only caller is `model_query`. That is a minimal, sound widening.
   - If finding 1 is fixed by moving the decision into `population::lookup`, the wrapper can be deleted.

3. **Pass: `OptionValue::from_admitted` doc.** It now states `F ≤ S`, and gives the chained-upcast example.

4. **Pass: set-literal test.** `lookup_expression_inside_a_set_literal_keeps_the_most_specific_element_type` asserts that the element is exactly `(U, B, "b1")` inside `Set<Reference<A>>`. That pins the most-specific-type invariant.

5. **Pass: PR body.**
   - Only V1-TYPE-023 is claimed, marked partial, and 021/022 are removed with reasons.
   - AC-6 is credited to L12, and L08 is the deferred item.
   - AC-3 `foreign-type` is credited to the admission layer (`model_population.rs` L05).
   - #164 is linked for the three deferrals.

6. **Style note (Low, no scenario):** `#[allow(clippy::too_many_arguments)]` on `evaluate_unresolvable_lookup` has no reason comment. Repo integrity rules want one. This goes away with the recommended fix in finding 1.

#157 has merged, so the rebase (adding `field_values: vec![]`) will be delta-checked separately once it is pushed.

**Required before merge:** finding 1, plus its regression case.

kreneskyp:
## Round 4 review: MERGE at `0747873fa8788b63f916e2e3ac7c0e95c73bab64`

Delta review of `ba8e129` → `0747873` (rebased onto `0f1112a`), in a fresh https clone.

### Rebase check: nothing lost
- `git range-diff` shows the three pre-rebase commits unchanged: `5ea93b3 = fd3c2c1`, `fc8cb83 = 5ce79ec`, `ba8e129 = f5c4c63`.
- `21749dd` only adds `field_values: Vec::new()` to the `member()` test helper, following #157's pattern.
- The PR's `src/model/population.rs` delta against `0f1112a` is the same as before: `type_catalog`, the doc update and `conforms()`. #157's `field_values`/FR-151 code is untouched.
- Current main (`8558318`, #156 and #162) merges cleanly. `src/value/mod.rs` changed on both sides with no conflict.

### Order in `evaluate_unresolvable_lookup` (`src/value/model_query.rs` ~330-385)
`conforms(S,T)` → `lookup.key` charge → `Universe` / `Identity{universe} if universe != binding.universe()` → absence dispatch. This matches `population::lookup`, and the refusal detail text is the same, byte for byte.

### Charge table: 6 reference shapes × 3 modes × 3 budgets (temporary probe, deleted)
- Outcomes and meters are identical between each malformed shape and its well-formed counterpart in every cell:
  - own universe + non-UTF-8 identity matches well-formed absent;
  - `[07;32]` + non-UTF-8 matches `[07;32]` + UTF-8.
- Unlimited budget:
  - foreign universe: `foreign_reference/foreign-universe`, w=1 vo=1 res=0, in every mode;
  - absent key: undefined 1/1/0, refused 1/1/0, empty 2/1/1.
- `work_units=0`: every shape stops at `LookupKey` with 0 consumed and next=1.
- `work_units=1`: foreign-universe rows refuse after w=1 in all modes; empty-absent stops at `LookupResultRetain` (lim 1, cons 1, next 1).
- Short universe, with either identity: foreign-universe after w=1.

### Revert probe
With the new guard arm removed, `lookup_expression_malformed_identity_in_a_foreign_universe_is_refused_not_absent` fails at `tests/model_reference_queries.rs:1129`, and the other 17 tests pass. The test guards the fix.

### Gates (at `0747873`)
- fmt: 0
- clippy `--locked --workspace --all-targets --all-features -D warnings`: 0
- `test --locked --workspace --all-features`: 0
- `make revendor-check`: 0 (native-v1 69/0/0, complete-value 64/0/0)

### Closing keywords
None in the commit messages or the PR body.

### Duplicated check order: ticket, not a blocker
No existing test pins the two copies to the same order. Each current test pins one outcome of the unresolvable path:
- malformed universe gives foreign-universe after 1 work unit;
- malformed identity gives none;
- the new foreign-universe + non-UTF-8 test.

Nothing compares `population::lookup` and `evaluate_unresolvable_lookup` at budgets 0/1 or across all modes. The probe above shows they agree today. A future change to `population::lookup` (a new check, a charge amount, or check order) would still desync silently. **Recommendation:** open a follow-up ticket for either fix:
- (a) move the decision into `population::lookup` with a `Member(String)`/`NonMember` key, which removes the copy; or
- (b) add a differential table test over modes × {own-absent, foreign universe} × budgets {∞, 0, 1} that asserts well-formed and malformed shapes give equal outcome and meter.

(a) is the real fix; (b) is the cheap guard.

### Nits (non-blocking)
1. `src/model/population.rs` `type_catalog` field doc names `crate::value::expression::model_query`. The module is `crate::value::model_query`.
2. `src/value/model_query.rs` `bridge_reference` clones `universe` inside the `map_err` closure. A `match` on `String::from_utf8` would move it instead.
3. The `foreign-universe` `ModelRefusal` is now built in three places (two in `model_query.rs`, one in `population.rs`). A shared constructor would keep the detail text in step. This belongs with the ticket above.

Row claims are unchanged from round 3: V1-TYPE-023 partial; V1-TYPE-021/022 not moved. QSL does not own the TC rows (#160).

PR #162, opened 2026-09-18
QSL #152: document numeric ordering and arithmetic in value module docs

kreneskyp:
## Audit result

Searched README.md, docs/*.md and spec/**/*.md for any claim that numeric ordering or integer/rational arithmetic is absent from QSL. Found none. `src/value/numeric.rs`'s `order_numbers`, `evaluate_integer_arithmetic`, `evaluate_rational_arithmetic` and `evaluate_boolean` already ship and nothing in the repo's own documentation says otherwise. Because no false absence claim exists, no test was weakened on that basis, so none needed tightening.

The one adjacent gap: `src/value/mod.rs`'s module-doc capability list enumerated FR-144 collections and FR-307 library resolution but skipped `numeric.rs`'s exports entirely, understating what step 4 of the pipeline actually covers.

## Change

`src/value/mod.rs`: add a list item naming `order_numbers`/`OrderingOperator`, `evaluate_integer_arithmetic`/`IntegerArithmetic`, `evaluate_rational_arithmetic`/`RationalArithmetic` and `evaluate_boolean`, under `quire.value.accounting/v1`; renumber the two list items after it.

## Gates

Branch created fresh from `origin/main` at `c0901d5`.

- `cargo fmt --all -- --check` — clean
- `cargo doc --no-deps --all-features` — same 11 pre-existing unrelated private-item-link warnings as before this change, no new ones
- `cargo clippy --locked --workspace --lib --bins -- -D warnings` (default features) — clean
- `cargo clippy --locked --workspace --all-targets --all-features -- -D warnings` — clean
- `cargo test --locked --workspace --lib --bins` (default features) — 46+6+0+9+2 passed, 0 failed
- `cargo test --locked --workspace --all-features` — 107 test-result blocks, all `ok`, 0 failed

Closes #152

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #161, opened 2026-09-18
QSL #139: name the static authority interface_version pin as a constant

kreneskyp:
## What FR-049 requires

`StaticAuthority` carries no producer interface version. FR-049 ([spec/functional/FR-049-admit-composed-evaluation-inputs.md](spec/functional/FR-049-admit-composed-evaluation-inputs.md)) declares that the accepted native-v1 observation contract's producer-interface-version member does not apply to `StaticAuthority`, and states what is checked instead:

> `StaticAuthority`'s document, model, profile and configuration identity and digest are checked independently of that domain-package identity: each is a static identity paired with a `sha256` digest in the `filament-canonical-json-1` domain, which the evaluator's `valid_digest` check enforces. Moving these identities to the FR-056 domain package's `sha256-jcs` digests is tracked separately (refs #131).

## Changes

- `src/state/input.rs`: deleted `STATIC_AUTHORITY_INTERFACE_VERSION` and `StaticAuthority::interface_version`. `StaticAuthority`'s doc comment states what the struct holds and how each field is checked.
- `src/state/evaluation.rs`: deleted the `interface_version` comparison from the `AuthorityRevision` check.
- `tests/compiled_protocol_v2.rs`, `tests/composed_state_evaluation.rs`: dropped the field from `StaticAuthority` construction. `tc_136_pre_reads_the_exact_invocation_pre_observation_and_binding`'s doc comment (already traced to TC-136/FR-049-AC-2) states what the authority sets and that evaluation completes from it.
- `spec/functional/FR-049-admit-composed-evaluation-inputs.md`: states the current behavior above; no AC or TC-136 trace changes, since the checked behavior is unchanged.

## Follow-ups (not fixed here)

- Moving `StaticAuthority`'s identities to FR-056 domain-package `sha256-jcs` digests is QSL #131 (refs #131).
- `quire-protocol`'s `tests/support/v2_state.rs` still writes `interface_version: "1.2.0"` for this field. That is a follow-up for quire-protocol's owner to drop when quire-protocol re-pins QSL past this change. Not edited in this PR.

## Gates

- `cargo fmt --all -- --check` — clean
- `cargo clippy --locked --workspace --all-targets --all-features -- -D warnings` — clean
- `cargo test --locked --workspace --all-features` — all test-result blocks `ok`, 0 failed
- `make revendor-check` — 0 drifted/stray files, confirming `resources/` is untouched

Closes #139

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #163, opened 2026-09-18
QSL #141: type ModelRefusal.cause as ModelRefusalCause

### LINKED ISSUE #141: ModelRefusal.cause should be a typed enum; ModelRefusal is a second error surface parallel to Diagnostic
## Context

PR #140 (FR-150 model normalization, QSL #120) introduced `crate::model::normalize::ModelRefusal { code: Code, cause: &'static str, detail: String }`. Review finding F9 (Opus reviewer, PR #140 re-review) flagged two related defects, deferred out of that PR's scope:

## F9a: `cause: &'static str` should be an enum

`ModelRefusal::cause` is a string literal constructed independently at each call site (`"unknown-owner"`, `"digest-domain-mismatch"`, `"specialization-cycle"`, `"wrong-model-selection"`, `"unsupplied-producer-record"`, `"unsorted-derivation"`, `"duplicate-path"`, `"unsorted-view"`, `"generalization-depth-exceeded"`, ...). Per the rust-review idiom "the variant set is the API, not the message": each of these is a distinct condition a caller must distinguish, so each should be its own enum variant with typed fields, not a shared string-payload variant. Symptom: tests can only assert `refusal.cause == "some-string"` and a typo in a literal at a new call site would silently create a new, undocumented cause with no compiler check.

## F9b: `ModelRefusal` is a second error surface parallel to `Diagnostic`

`crate::model::normalize::normalize`/`normalize_with_meter` return `ModelRefusal` directly rather than converting into (or wrapping) this crate's existing `Diagnostic` error type used elsewhere. Nothing currently converts a `ModelRefusal` into a `Diagnostic`, so a caller integrating FR-150 normalization into a larger diagnostic-reporting pipeline has two incompatible refusal shapes to handle. Decide whether `ModelRefusal` should implement `From<ModelRefusal> for Diagnostic` (or similar), or whether `Diagnostic` itself should grow a variant carrying `ModelRefusal`.

## Related, smaller cleanup noted during the same review (fix here if convenient, otherwise as a follow-up)

- **F11**: `src/model/normalize.rs`, `charge_all`'s `record_keys.sort()` sorts the bundle's producer keys but the sorted *order* is never used — only `record_keys.len()`/the running `enumerate` position matters for the `normalize.record` charge sequence. This is dead work (a `Vec<ProducerKey>` clone + `O(n log n)` sort to throw away the order), left over from the pre-#140 shape and not resolved by the F1 rework.
- **P2**: `Code::ResourceExhausted` is reused for `generalization-depth-exceeded"` in `ancestor_paths`, and `MAX_GENERALIZATION_DEPTH` (`src/model/normalize.rs`) duplicates the same "128" bound as `MAX_CHECKING_DEPTH` (`src/value/expression/check.rs:32`) as an independent constant. Consider whether these two depth bounds are actually the same normative limit (in which case they should be one named constant) or genuinely independent (in which case a doc comment should say why they happen to share a value).

None of these are correctness defects — normalization behaves correctly today — they are API-shape and duplication cleanup.

kreneskyp:
## F9a: ModelRefusalCause enum

`ModelRefusal.cause` was `&'static str`; the actual variant set lived in scattered string literals across `normalize.rs`, `dispatch.rs`, `population.rs`, `systems.rs` and `conformance.rs`, undiscoverable from the type alone. Per rust-review's "the variant set is the API, not the message" idiom, added `ModelRefusalCause`, one variant per distinct cause (56 total, after the three #157-driven causes below), each with a doc comment and an `as_str()` mapping back to the original kebab-case string. `ModelRefusal.cause` and every helper that produces one (`SubtypeRefusal`, `UnclosedMethodSet`, `AxisFailure`, `RedefinitionTargetOutcome::Refused`, `ConditionFailure`, and the parameterized `dangling`/`missing_member` helpers) now carry the enum instead of the literal.

`EffectiveDeclarationPreimage::validate_derivation` in `key.rs` (called only from tests, never touches `ModelRefusal`) also returns `ModelRefusalCause` now (via its own `Result<(), (ModelRefusalCause, String)>`), for consistency with the rest of the cause vocabulary. F9a's own text already names `"unsorted-derivation"`/`"duplicate-path"` correctly as belonging to this second type, not `ModelRefusal`.

### Review follow-up (4 findings)

1. **No test named the wire spellings.** Added an exhaustive spelling test in `src/model/refusal.rs`: `expected_tag(&ModelRefusalCause) -> &'static str` is a `match` over all 56 variants with **no wildcard arm**, so an unhandled new variant fails to compile rather than silently passing; a `Vec<ModelRefusalCause>` (one instance per variant) asserts `cause.as_str() == expected_tag(cause)` for each. This replaces an earlier hand-counted `assert_eq!(cases.len(), 53, ...)`, which could drift from the enum without anyone noticing.
2. **Merge coordination with #157/#158.** Rebased onto `origin/main` at `8445da9` (contains #156, #157, #158, #162). Added the three #157-driven string causes as typed variants: `MalformedDeclaration`, `DuplicateMember { object: String, field: ProducerKey }`, `SubsettingViolation { object: String, record: ProducerKey, subsetting: ProducerKey, subsetted: ProducerKey }`. #158 added `src/value/model_query.rs`, built before this PR's typing landed on main — see the new "#158 follow-up" section below for what that required.
3. **Variants carried no typed fields.** Every variant whose `detail` text names an identity/path/name, where *every* construction site for that variant supplies the same shape of data, now carries that data as a typed field, using the crate's own identity types rather than `String` wherever the site has one: `ProducerKey` (e.g. `UnknownOwner { member: ProducerKey, owner: ProducerKey }`), `EffectiveId` (e.g. `ConflictingIdentity`, `ForeignUniverse`), and `Multiplicity` (`MultiplicityNarrowing { from: Multiplicity, to: Multiplicity }`). `UnknownEndpoint` was restructured from a two-`String` shape to `{ end: &'static str, relationship: ProducerKey, missing: ProducerKey }` to match what its one construction site actually has in scope. A variant stays a unit variant only when it has no such data at all (`WrongExport`, `TypeMismatch`, `UnsuppliedProducerRecord`, `UnprovedRefinement`, `OperatorIneligible`, `VarianceResult`, `NoApplicable`, `MultipleUndominated`, `UnclosedMethodSet`, `RedefinitionTarget`, `MalformedDeclaration` — 11 total) or when its own construction sites genuinely disagree in shape. `as_str()`/`Display` still return the unchanged tag, and every `detail` string is byte-identical to before. `ModelRefusalCause` necessarily drops `Copy` (and the now-unused `Hash`) since most variants carry owned data; callers that read `.cause` by value now `.clone()` it.

   Two tests (`model_normalization.rs`'s `r01`/`r01b`) assert `matches!(refusal.cause, ModelRefusalCause::SpecializationCycle { .. })` rather than an exact-field comparison — deliberately: both tests' own docstrings record that which type's generalization walk detects the cycle first is legitimately nondeterministic, so an exact-field assertion would be flaky, not stricter.
4. **Module cycle (`key.rs` importing from `normalize.rs`, which imports from `key.rs`).** Moved `ModelRefusalCause` into its own lower-level module, `src/model/refusal.rs`, which now imports `ProducerKey`/`EffectiveId` from `key.rs` — `key.rs` already imported `ModelRefusalCause` from (what was) `normalize.rs`, so this is a reciprocal `use` cycle between two files in the same crate, which Rust permits; correct types matter more than avoiding it. `normalize.rs` re-exports it (`pub use crate::model::refusal::ModelRefusalCause;`), so `crate::model::normalize::ModelRefusalCause` is still a valid path — no public rename.

### Size-lint fix: `ModelRefusalCause`/`ModelRefusal` growth

Typing ~45 of 56 variants with `ProducerKey` (152 bytes) and `EffectiveId` (32 bytes) fields grew `ModelRefusalCause` to 480 bytes and `ModelRefusal` (which embeds `cause` directly) to 512 bytes, tripping clippy's default `large_enum_variant`/`result_large_err` lints at 26 sites across `conformance.rs`, `normalize.rs`, `population.rs`, `systems.rs`, `dispatch.rs`, `key.rs` — everywhere `ModelRefusal` appears in a `Result` or an `Outcome::Refused` variant.

Fixed with a module-level `#![allow(clippy::large_enum_variant, reason = "...")]`/`#![allow(clippy::result_large_err, reason = "...")]` in each affected file, matching `src/state/evaluation.rs`'s existing precedent for exactly this class of problem (a closed typed-failure enum retaining complete identities without heap allocation) — not a new pattern for this codebase. Considered and rejected boxing `ModelRefusal.cause` into `Box<ModelRefusalCause>` instead: it would fix the lint in one place, but at the cost of `*`-dereferencing roughly 90 sites this same PR round already wrote (every `assert_eq!`, `.clone()`, and construction site), for no behavioral benefit and against the repo's own stated preference (per `evaluation.rs`'s `reason` text) for avoiding heap allocation in this exact situation.

### #158 follow-up: `src/value/model_query.rs`

#158 merged to main after this PR's typing round was first written, and built `model_query.rs`'s `foreign-universe`/`type-mismatch`/`absent-key` refusals as raw `&'static str` literals (the only shape available before this PR). After rebasing:

- `ModelQueryRefusal.cause` (`src/value/outcome.rs`) stays `&'static str` at its own crate boundary, so the one site that builds it from a `ModelRefusal` now calls `refusal.cause.as_str()`.
- `model_query.rs`'s four raw-string construction sites now build the typed variants (`ModelRefusalCause::TypeMismatch`, `::ForeignUniverse`, `::AbsentKey`) instead.
- `ForeignUniverse.actual` (was `EffectiveId`) and `AbsentKey.key` (was `String`) are widened to `Vec<u8>`. Both fields were sound as originally typed for `population.rs`'s own well-formed construction site, but `model_query.rs`'s malformed-reference path constructs the *same* cause tags for a reference whose universe is not 32 bytes, or whose identity is not valid UTF-8 — exactly the case those two paths exist to refuse without a substituted or truncated value, per the module's own doc comment ("reporting the bytes the caller actually supplied"). `Vec<u8>` represents both the well-formed and malformed cases honestly; `population.rs`'s site and the spelling test's fixture were updated to match. `detail` string wire output is unaffected.
- #158 also added `tests/model_reference_queries.rs` using the `test-support` fixture constructors (`ProducerKey::fixture`, `ModelSelection::fixture`) without adding its `required-features = ["test-support"]` entry to `Cargo.toml` (the mechanism #156 added for exactly this). That gap broke `make ci-default-features` — `cargo clippy --all-targets` under default features failed to compile that test file — as soon as this branch rebased onto main; added the missing `[[test]]` entry alongside the other five.
- `population.rs`'s own `conforms()` (also new in #158) returns `Result<bool, ModelRefusal>`, tripping the same `result_large_err` lint the rest of this PR already fixed elsewhere; added the same module-level allow there too.

## F9b: no `From<ModelRefusal> for Diagnostic`

Not added. No existing refusal type in the crate (`LibraryRefusal`, `PackageRefusal`, `ModelSourceCause`) converts into `Diagnostic` anywhere in the codebase, and `Diagnostic`'s required fields (`phase`, `code`, `source`, `path`, `span`) have no natural values drawn from `ModelRefusal`'s domain — `ModelRefusal` already carries its own `Code` via the enclosing type. Adding the conversion now would invent an API with no caller.

## F11: dead sort

Removed the dead `record_keys.sort()` + `Vec<ProducerKey>` clone in **both** `charge_all` and `unsupported_capability_refusals` in `normalize.rs`. The issue named one site; the second is the identical defect duplicated — the sorted clone was never read after being built in either. Both now iterate `bundle.records` directly by index.

## P2: MAX_GENERALIZATION_DEPTH vs MAX_CHECKING_DEPTH

Both equal 128, but they bound independent recursions: `MAX_GENERALIZATION_DEPTH` (`normalize.rs`) bounds FR-150's generalization ancestor-path walk; `MAX_CHECKING_DEPTH` (`value/expression/check.rs`) bounds Complete-V1's expression-checking recursion. No spec text links them. Added cross-referencing doc comments on both constants stating they are coincidentally equal, not the same normative limit.

No public item outside `ModelRefusalCause`'s own introduction, its move to `src/model/refusal.rs`, and the F9a/F11 fixes above was renamed.

## Gates

Rebased onto `origin/main` at `8445da9` (contains #156, #157, #158, #162).

- `make ci-default-features` (`cargo fmt --all -- --check`; `cargo clippy --locked --workspace --all-targets -- -D warnings`; `cargo test --locked --workspace`) — clean
- `make ci-all-features` (`cargo clippy --locked --workspace --all-targets --all-features -- -D warnings`; `cargo test --locked --workspace --all-features`) — clean
- `make ci-clean-build` (from-clean `--no-default-features` build; fixture-audit self-test; parse example) — clean
- `cargo test --locked --workspace --all-features` — all test-result blocks `ok`, 0 failed, including the new exhaustive `as_str`/`expected_tag` coverage test

Closes #141


kreneskyp:
## Opus review: PR #163 at `0bd53c405d9ce12047c13c45f9ed37d7a74c776a`

**Verdict: FIX.** One small fix is needed (finding 1). The refactor itself is correct.

### Gates (run in a fresh clone)
- `cargo fmt --all -- --check`: clean
- `cargo clippy --all-features --all-targets -- -D warnings`: clean
- `cargo test --all-features`: 103 result blocks, all ok, 0 failed
- `make revendor-check`: native-v1 69 matched, complete-value 64 matched, 0 drifted, 0 stray

### Checks
1. **Wire strings are unchanged.** I paired every removed cause literal with its replacement `ModelRefusalCause::X` in src/ and tests/: 141 substitutions. Mapped through `as_str()`, all 141 match the old literal exactly. The table has 53 entries, and every variant is built at least once in src/. No variant is unreachable, and no old string is left without a variant.
2. **Scope.** The key.rs `validate_derivation` conversion is inside #141. The issue lists `"unsorted-derivation"` and `"duplicate-path"` in F9a by name. The PR body's "correction to the issue's own text" is wrong on that point, but the code is fine. No existing public item was renamed. The only public signature changes are the ones #141 asks for: `cause` fields and `validate_derivation`'s error type now use the enum.
3. **F9b.** #141 asks for a decision ("Decide whether..."), not for the `From` impl. The decision is justified: nothing in the crate implements `From<_> for Diagnostic`, and no caller needs one.
4. **F11 / P2.** Loop count and charge sequence are the same as before. The doc cross-links resolve (`crate::value::MAX_CHECKING_DEPTH` is re-exported).
5. **Derive set, `as_str`, `Display`.** The derive set `Clone, Copy, Debug, Eq, Hash, PartialEq` matches the sibling cause enums (`CompleteCause`, `LibraryCause`, `CompoundUnitCause`). `Display` delegates to `as_str`. Both are fine.

### Findings
1. **Medium: `src/model/normalize.rs:236` (`as_str`). The wire spellings no longer have any test.** Before this PR, every refusal test asserted `cause == "kebab-string"`. That was the only guard on the tag spelling. The PR changes every assertion to compare variants, and nothing in src/ or tests/ calls `ModelRefusalCause::as_str()`. Failure scenario: someone edits one arm, for example `UnknownOwner => "unknown_owner"` under the underscore id ruling. Every test still passes, and the changed tag reaches consumers. #158's `Refusal::cause()` would expose it. **Fix:** add one test that asserts `as_str()` for all 53 variants against the literal spellings.
2. **Merge coordination, not a defect in this PR.**
   - **#158 (`5ea93b3`):** merges without textual conflicts, but I expect it to stop compiling. `ModelQueryRefusal { cause: &'static str }` is filled from `refusal.cause` in `model_refusal()`, and after #163 that is a `ModelRefusalCause`. The fix on rebase is `refusal.cause.as_str()`. This comes from reading the code. I did not compile the merged tree.
   - **#157 (`50bf2ba`):**
     - Textual conflicts in `src/model/conformance.rs` (3 hunks: the import, the `RedefinitionTargetOutcome::Refused` field or doc, and the refinement arm) and `tests/model_conformance.rs` (2 hunks: the r07 `Refused { cause, candidates, valid_targets }` asserts).
     - A semantic conflict: #157 adds a new literal cause `"subsetting-violation"` in `population.rs` and its test. After #163 this needs a new `SubsettingViolation` variant (plus its `as_str` arm and the test from finding 1).
   - Whichever PR lands second has to absorb these.
3. **Low, for Peter to decide: variants carry no typed fields.** #141 F9a says "its own enum variant with typed fields". The variants are unit variants, and the offending keys are still only available inside the `detail: String` text. The discriminant is now typed, which is the main point of F9a. Typed payloads can be a follow-up if a caller needs them.
4. **Low, style: `src/model/key.rs:17` creates a module cycle.** `key.rs` now imports from `normalize.rs`, which already imports from `key.rs`. The enum is a shared vocabulary for five modules plus `key`, so it could live in its own lower-level module. No behavior impact.
5. **Info, out of #141 scope:** `ConditionFailure.condition` (`systems.rs:376`) and `AxisFailure.axis` (`conformance.rs:62`) are still `&'static str` discriminants. No action in this PR.

kreneskyp:
## Opus re-review: PR #163 at `caf94ff48f1e1ad9273d287b2361b0e29ca9d5c2`

**Verdict: FIX.** The wire output is unchanged and the fix for my last round's finding 1 is correct. But the new typed fields have no tests, and several variants left without data have one construction site, so "sites disagree in shape" cannot be the reason.

### Gates (run in a fresh clone)
- fmt: clean
- `clippy --all-features --all-targets -D warnings`: clean
- `test --all-features`: all ok
- `make revendor-check`: native-v1 69 matched, complete-value 64 matched, 0 drifted, 0 stray

### Checks
- **Wire tags are unchanged.** I paired each changed site's removed literal with its new variant and mapped the variant through `as_str()`. All 195 pairs match origin/main exactly.
- **The `as_str` test's strings come from origin/main.** Its 53 expected strings are exactly the set of cause literals the diff removes from origin/main's src/, so the test does not pass by construction. One doc nit: the comment says the strings come "from `as_str`'s match arms on `origin/main`", but origin/main has no `as_str`. They come from the literals at the call sites.
- **`detail` text is unchanged.** No `detail:` or `format!` line changed anywhere in src/. Only the `cause:` lines moved.
- **Public paths.** `normalize::ModelRefusalCause` is re-exported, and every test still imports it from the old path. Nothing public was renamed.
- **Size.** `refusal.rs` is +677 lines: about 340 of enum and docs, and about 330 for the spelling table after rustfmt spreads out the `String::new()` placeholders. The rest of the diff is in proportion. No churn.

### Findings
1. **Medium: no test checks any typed field.** Every cause assertion in tests/ is `matches!(cause, Variant { .. })`: 8 in model_population, 8 in model_normalization, 3 in model_conformance, 1 in model_systems. The rust-review idiom names this as the symptom to look for. Failure scenario: a site swaps two fields, for example `UnknownOwner { member: owner, owner: member }`, `ConflictingIdentity { existing_type, declared_type }` in reverse, or `ForeignUniverse { actual, expected }` reversed. Every test passes. **Fix:** at least the existing refusal tests should assert the field values they already check through `detail.contains(..)`.
2. **Medium: some unit variants should carry data.** The doc says a variant stays a unit variant when its construction sites "disagree in shape", but these don't:
   - `DerivationConflict` has 1 site (normalize.rs:1057): type, member, the undominated redefiners.
   - `AboveMaximum` has 1 site (population.rs:733): `selected` and `maximum` counts. `VarianceParameter` already carries a `usize`, so counts are not excluded.
   - `UnsortedDerivation` has 1 site (key.rs:338): original, position, ordinal.
   - `DuplicatePath` has 1 site (key.rs:350): original, earlier, later.
   - `UnknownValueType` has 2 sites (normalize.rs:601/617). Both have an operation and a value type. The parameter can be an `Option`.
   - `MultiplicityNarrowing` has 5 sites. All of them format two `Multiplicity` values (the checked one and the one it must conform to). The parameter index and end label can be `Option`s.

   Also, the PR says 38 variants have fields and 15 do not. The file has 37 and 16.
3. **Low-Medium: field types.**
   - Every field is a `String`, including `EffectiveId`s stored as hex (`UnsortedView.at`, `ConflictingIdentity.*_type`, `ForeignUniverse.*`). A caller has to format its own id to compare it.
   - `UnknownEndpoint.item` holds prose: `format!("{} end of {}", label, relationship.key.identity)` at systems.rs:275.
   - `UnclosedSubtypes.type_name` uses the sentinel `"<no members>"` where the idiom is `Option<String>`.
   - This probably follows from the module-cycle fix: `refusal.rs` cannot import `key` types because `key` imports `refusal`. That finding was only Low style. It should not force string payloads. Accepting the module cycle, or giving `validate_derivation` its own small enum, would let the fields be `ProducerKey` or `EffectiveId`.
4. **Low: the spelling test is not exhaustive.** `cases.len() == 53` is a hand count. Add a variant to the enum and `as_str` but not to `cases`, and the test still passes. #157 will add three. A `match` with no wildcard arm inside the test would make the compiler enforce every row.

### Conflicts once #157 and #158 land
- **#157 (`847e1e1`):**
  - Textual conflicts in `src/model/conformance.rs` (3 hunks) and `tests/model_conformance.rs` (2 hunks).
  - Three new string causes will need variants: `"malformed-declaration"` (conformance.rs:700), `"duplicate-member"` (population.rs:656) and `"subsetting-violation"` (population.rs:687). The spelling test then goes from 53 to 56 cases.
- **#158 (`ba8e129`):** no textual conflict. I expect a compile break: `ModelQueryRefusal { cause: &'static str }` is filled from `refusal.cause` and needs `.as_str()`. I found this by reading the code; I did not compile the merged tree.

kreneskyp:
## Opus re-review 3: PR #163 at `395d63b213c9f08bc7455338b37742779affa27b` (rebased on `8445da9`)

**Verdict: FIX.** One item from my last round is still open: most refusal tests still don't check the typed field values (finding 1). Everything else below is either resolved or a non-blocking ruling.

### Gates (run in a fresh clone)
- `make ci` (default-features, all-features and clean-build lanes): **pass**
- `make revendor-check`: native-v1 69 matched, complete-value 64 matched, 0 drifted, 0 stray
- **main `8445da9` fails `make ci-default-features` on its own.** There are 11 `E0599` errors: `ProducerKey::fixture` is not found in `tests/model_reference_queries.rs`, because #158 left out its `[[test]] required-features = ["test-support"]` entry. `.github/workflows/ci.yml`'s `--no-default-features` clippy and test steps would hit the same error. **Main's CI is red today.** The Cargo.toml entry this PR adds is correct and matches its sibling entries.

### Output is unchanged from main
- I paired every changed site's old literal with its new variant through `as_str()`: 271 pairs, 0 mismatches, 56 variants.
- No `detail:` or `format!` line changed anywhere in src/, including the #158 malformed-byte messages in `model_query.rs`. Only the `cause:` lines changed.

### My earlier findings
1. **Typed fields in tests: partly fixed. This one blocks.** 13 new assertions check field values, covering 9 variants, all of them newly given data. The 20 `matches!(cause, Variant { .. })` wildcards from last round are unchanged, and they include exactly the field pairs that can be swapped: `UnknownOwner`, `ConflictingIdentity`, `ForeignUniverse`, `ForeignModelSelection`, `PortDirection`, `SpecializationCycle`, `DigestDomainMismatch` and `UnknownMember`. The PR says a "field-swap guard probe" was run, but a probe that isn't committed doesn't guard anything. **Fix:** turn those 20 into `assert_eq!` on the full cause. Each test already has the expected keys or ids in scope for its `detail.contains(..)` check.
2. **The six variants that could carry data: fixed.** `DerivationConflict`, `AboveMaximum`, `UnsortedDerivation`, `DuplicatePath`, `UnknownValueType { parameter: Option<_> }` and `MultiplicityNarrowing` now have fields. The 11 unit variants that remain have no single shape of data across their sites. That split is now credible.
3. **Field types: fixed.** Fields use `ProducerKey`, `EffectiveId` and `Multiplicity`. `UnclosedSubtypes.type_name` is an `Option<ProducerKey>`. `UnknownEndpoint` no longer stores prose.
4. **Spelling test: fixed.** `expected_tag` is a `match` with no wildcard arm, the `cases` list covers all 56 variants, and the comment no longer cites an `as_str` on origin/main.

### Ruling: the module-level allows
- **What they hide today.** With all 11 `#![allow]`s removed, clippy reports 27 sites: 19 `result_large_err` on functions, 1 on a closure and 7 `large_enum_variant`. All 27 are the same cause: `ModelRefusal` is 504 to 512 bytes. Nothing unrelated is hidden right now.
- **Is the pattern acceptable.** It is the same pattern as `src/state/evaluation.rs`, and the lint is a performance heuristic on a refusal path that is rarely taken. It is **acceptable to merge**.
- **But the `reason` text is false.** It says "without heap allocation", but `ProducerKey` holds `String`s (`authority`, `identity`) and `ModelRefusal.detail` is a `String`. Every refusal already allocates, so heap avoidance is not the trade-off. **Action:** fix the `reason` in all 6 files to the true one, for example "cold refusal path; `ModelRefusalCause` carries `ProducerKey`s inline". Put `key.rs`'s allow on `validate_derivation` itself rather than the whole module, since that function is its only site.
- **Boxing.** If the outcome enums ever sit on a hot path, `cause: Box<ModelRefusalCause>` is the follow-up. It brings `ModelRefusal` to about 40 bytes and clears all 27 sites. It is not needed for this PR.

### Ruling: `Vec<u8>` in `ForeignUniverse.actual` and `AbsentKey.key`
- **Acceptable.** It is lossless for both sources: #158's malformed reference bytes (a universe that isn't 32 bytes, an identity that isn't UTF-8) and the well-formed population sites. #158's own module docs say the crate defines no mapping from model identities into the byte space `ObjectReference` uses, so reusing `EffectiveId` or `ObjectIdentity` for malformed bytes would invent a mapping.
- **The cost.** In `population.rs` a well-formed `EffectiveId` or `String` is flattened to bytes and loses its type.
- A precise shape would be a two-arm enum, a `Model(EffectiveId)` case and a `Reference(Box<[u8]>)` case. That is optional polish and does not block.

### Notes
- **Closing keywords:** the PR body has only `Closes #141`, and GitHub's `closingIssuesReferences` is `[141]`. But commit `395d63b`'s subject says "fix #158 gate gap", and `fix #158` is a GitHub closing keyword. #158 is an already-merged PR, so it would have no effect. Still, reword it in the squash message so the only closing keyword is `Closes #141`.
- **Conflicts:** none. The PR is rebased on `8445da9`, and the three #157 causes (`MalformedDeclaration`, `DuplicateMember`, `SubsettingViolation`) are present.

kreneskyp:
## Opus delta review: PR #163 at `c3f6958983ffb54860b65bb676eeb77a5f9761c7` (range `395d63b..c3f6958`)

**Verdict: FIX.** One variant is still not checked by any test, and its payload loses the data that distinguishes the refusal. The fix is small. Everything else checks out.

### Checks
1. **No wildcard `matches!` on refusal causes is left** in tests/model_{normalization,population,systems,conformance,dispatch,reference_queries}.rs. The `matches!` calls left in `tests/model_source.rs` are on `ModelSourceCause`, a different type outside this PR.
2. **Swap probes.** On the merge result, I swapped two fields in src, one variant at a time, and ran the matching test binary:

   | Variant | Swap in src | Result |
   |---|---|---|
   | `SpecializationCycle` | `ancestor` ↔ `via` | **fails** (r01, r01b) |
   | `UnknownOwner` | `member` ↔ `owner` | **fails** |
   | `ConflictingIdentity` | `existing_type` ↔ `declared_type` | **fails** (l05_conflicting_identity) |
   | `ForeignModelSelection`, view site (population.rs:520) | `actual` ↔ `expected` | **passes: not guarded** |

   The other new `assert_eq!` values match their src sites field by field, and each pair uses distinct values.
3. **Allow reasons.** All are corrected ("cold refusal path; ModelRefusalCause carries ProducerKeys inline"). `key.rs`'s allow is now an item-level `#[allow]` on `validate_derivation`. Resolved.
4. **Merge onto main `af10b0b` (#161).** It merges with no conflicts. On the merge commit, `make ci` passes and `make revendor-check` is clean (69 and 64 matched, 0 drifted, 0 stray).
5. **Closing keywords.** The body has only `Closes #141`. You are setting the squash message yourself.

### Finding
**Medium: `ForeignModelSelection` at the view site stores the export identity strings, which cannot express a mismatch that differs only in revision** (src/model/population.rs:520–523, tests/model_population.rs `l05_view_from_a_different_bundle_revision_refuses_at_admission`).

That test's whole point is a revision-only mismatch: the same `bundle.n01` identity at revisions "1" and "2". So the cause it now asserts is `ForeignModelSelection { actual: "bundle.n01", expected: "bundle.n01" }`. The typed payload of a "foreign selection" refusal says the two selections are equal, and swapping `actual`/`expected` in src passes every test. No test exercises the view site with two different identities. The `detail` text has the same blind spot, but that text predates this PR and must stay as it is. The typed field is new here and should not copy the blind spot.

**Fix:** keep the full `ModelSelection` at the view site. For example:
- `ForeignModelSelection { actual: OfferedSelection, expected: ModelSelection }`
- `enum OfferedSelection { View(ModelSelection), Document(String) }`, because the document site only has `modelIdentity`.

Then make l05 assert revision "1" against "2". `as_str` and `detail` stay unchanged.

kreneskyp:
## Opus delta review: PR #163 at `05a154623ffde2691446c1cafcfedde5064796d3` (range `c3f6958..05a1546`)

**Verdict: MERGE.**

1. **The payload is correct at both sites.**
   - The view site carries `OfferedSelection::View(view.model_selection.clone())` against `expected: bundle.model_selection.clone()`.
   - The document site carries `OfferedSelection::Document(document.model_identity.clone())` against the full bundle selection.
   - l05 (view site) now compares the revision-"1" view selection against the revision-"2" bundle selection. l05 (document site) asserts `Document("bundle.other")`.
2. **The swap probe goes red.** I swapped `actual` and `expected` at population.rs:523. `l05_view_from_a_different_bundle_revision_refuses_at_admission` FAILED (19 passed, 1 failed). I reverted the swap afterwards.
3. **Output is byte-identical to main.** In the full diff from main, the only removed src lines that aren't comments or imports are cause literals, the old `validate_derivation` signature and the dead sort code from F11. No `detail`/`format!` line changed. All 271 changed sites still give the same `as_str` tag as the literal they replaced, across 56 variants.
4. **Merge onto main `af10b0b` is clean.** On the merge commit, `make ci` passes (default-features, all-features and clean-build lanes) and `make revendor-check` is clean (69 and 64 matched, 0 drifted, 0 stray).

All findings from my earlier rounds are resolved. This PR also fixes main's red `ci-default-features` gate (the missing `[[test]]` entry for `model_reference_queries`).

Issue #139, opened 2026-09-17
Hardcoded interface_version == 1.2.0 check in state/evaluation.rs and state/input.rs


kreneskyp:
Found during #131's review and deliberately **not** fixed there, because it is a different contract and fixing it would have been scope creep. Filed so it is not lost.

## What is there

`src/state/evaluation.rs` and `src/state/input.rs` hardcode an `interface_version == "1.2.0"` check, with doc comments naming Producer 1.2.

## Why it is not #131

#131 deleted the Producer 1.2 / Filament correspondence **adapter** — `agent-ix-baseline-producer`, `producer.rs`, `AdmittedProducerModel`, `ModelSelection.producer`, `PRODUCER_CONTRACT_REVISION` and the `_with_producers` API family. That work is done and merged as #137.

This check belongs to a **different** contract: FR-049 runtime authority and evidence. It shares the Producer 1.2 name and version string, which is exactly why it looks like leftovers from the deletion and is not. #131's removal list does not name it, and the reviewer was right to flag rather than fix it.

## Why it still deserves a look

The standing ruling is that this repo is prerelease and we do not keep compatibility with protocol versions nobody uses. A bare string equality against `"1.2.0"` raises the question of what happens to a document that names anything else, and whether the check is a **refusal control** (which stays, and should emit a named typed refusal) or a **compatibility gate** (which does not).

Those are different things and the code should say which it is:

- If it is a refusal control, it stays, but a non-matching version should produce a named refusal carrying the version it saw, not a bare boolean mismatch.
- If it is pinning FR-049's runtime authority contract to a version that is simply the current one, the literal should be a named constant with the contract it belongs to, so it is obvious that bumping it is a contract change.
- If the version is vestigial, it goes.

## Scope

Read FR-049, decide which of the three it is, and make the code say so. Update the Producer-1.2-named doc comments either way — they currently point a reader at a deleted adapter.

Do not reintroduce version dispatch. One contract at a time.

kreneskyp:
No downstream consumer — this can be decided on FR-049's merits alone.

Agent E checked contract-ir on `origin/main` at `cd58577`, across `crates/`, `src/`, `tests/` and `spec/`:

- zero hits for `interface_version`
- zero hits for the literal `1.2.0`
- zero hits for `PRODUCER_CONTRACT_REVISION` or `AdmittedProducerModel`

[...]

PR #167, opened 2026-09-18
Fix QSL #145: bound and exactly meter phase-4 dominance resolution

### LINKED ISSUE #145: Phase 4 dominance is O(|edges|^2) and dominates() budgets unmetered across calls
PR #144 review finding #4 (fix-in-a-follow-up, not in that PR).

## What

`src/model/normalize.rs`'s phase 4 (explicit field redefinition) resolves
conflicting redefiners by proper-descendant dominance. Two defects, found
during PR #144 review:

1. **`O(|edges|^2)` dominance resolution.** Phase 4's conflict-resolution
   loop (around `src/model/normalize.rs:971-987`) calls `dominates` for
   every pair of contesting redefinition edges targeting the same member,
   which is quadratic in the number of contesting edges.

2. **`dominates` (`src/model/normalize.rs:1073-1086` prior to the PR #144
   doc fix) re-runs `ancestor_paths` under a budget that starts fresh at
   zero facts consumed on every call** (`remaining_fact_budget(limits, 0)`),
   unlike `build`'s own phase-3 walk, which threads its real
   `facts_so_far` through. Each individual `dominates` call is bounded, but
   the `O(|edges|^2)` loop over it is not metered cumulatively the way
   `build`'s own walk is — a bundle with many contesting redefinition edges
   over the same member could do much more work than the `ModelNormalizationLimitsV1`
   counters would suggest.

The doc comment on `dominates` was corrected in PR #144 to describe this
honestly instead of falsely claiming it is "bounded the same way" as
`build`'s metered path; the underlying performance/metering defect itself
is unfixed.

## Also: consolidate the three "proper descendant" implementations

Three near-identical proper-descendant/type-conformance graph walks exist
independently in this crate:

- `src/model/conformance.rs`'s `type_conforms`
- `src/model/normalize.rs`'s `dominates` (via `ancestor_paths`)
- `src/model/dispatch.rs`'s `dominates` (delegates to `conformance::type_conforms`,
  but has its own dominance-pair bookkeeping on top)

These should be reviewed for consolidation into one bounded, consistently
metered graph-walk primitive.

## Scope

- Fix or document-accept the `O(|edges|^2)` phase-4 loop.
- Thread real cumulative `facts_so_far` through `dominates`'s budget, or
  otherwise account for cross-call cost within `ModelNormalizationLimitsV1`.
- Evaluate consolidating `type_conforms`/`normalize::dominates`/`dispatch::dominates`.

Filed per PR #144 review finding #4 ("file as a ticket, do not fix here").

kreneskyp:
## Summary

Part of #120. Closes #145.

This PR is the #145 work only. V1-TYPE-028 is unchanged by it — dispatch's grammar/IR/check/link/evaluate wiring is separate work, tracked in its own PR under #120.

QSL #145 named two defects in `src/model/normalize.rs` phase 4's redefinition-conflict resolution: an O(|edges|^2) dominance-walk loop, and a per-call budget that reset to zero instead of tracking real cumulative cost. Three rounds of review have gone into this revision; this revision fixes every finding from all three rounds.

## Current design

- `normalize.redefinition-check` charges once per redefinition record in the entire bundle — field-member and operation-member redefinitions alike — sorted once, ascending by the record's own producer key. `r` is that record's 0-based position in that global ascending order; `m` is the record's owning type's effective member count (field members plus operation members, each counted once whether direct or inherited). This sequence is computed once in `build()`, before the per-type phase-4 loop runs, matching `value-accounting.md:455`'s record-level, bundle-wide charge.
- `normalize.conflict-check` charges `Σ (c − 1) × f(o)` per contested target reached by `c >= 2` redefinition records, where `f(o)` is owner `o`'s own count of type derivation facts. A group with `c == 1` charges nothing. This applies uniformly to field and operation targets: a contested operation-redefinition group (for example, two operation members both redefining the same inherited operation) is charged the same as a contested field group, even though — see below — nothing resolves which operation redefiner wins.
- Owner ancestor sets used by phase 4's dominance check are derived from each owner's own `type_paths` (already computed in phase 3), not from a second, separately-bounded walk. This removes `MAX_CONFORMANCE_DEPTH` from phase 4's cost model entirely: phase 4's cost now scales with each owner's own ancestor count, not a depth ceiling that a type's *breadth* of direct generalizations could exceed even with no live conflict.
- **Current behavior:** when phase 2/3's own fact budget is already exhausted going into phase 4, phase 4's resolution loop does not run at all — the outcome is `Incomplete`, never a phase-4 refusal (`derivation-conflict` or `redefinition-unreachable`) built from a truncated phase-2/3 path set. This matches FR-150's "exhaustion ends checking": `charge_all`'s own replay of the phase-2/3 facts that triggered the truncation runs, in charge order, before phase 4's charges are ever reached, so it is always that replay — never phase 4 — that reports the `Incomplete`.
- Refusal precedence: phase 2/3 runs to completion (or refusal) as its own pass before phase 4 begins. A phase-3 refusal (for example, `specialization-cycle`) is therefore always found and returned before any phase-4 charge or refusal (for example, `derivation-conflict`) is even computed, when a bundle has both defects.
- Operation-member redefinitions are charged in the same `normalize.redefinition-check` and `normalize.conflict-check` passes as field-member redefinitions, per `value-accounting.md`'s record-level charges, which do not filter by member kind. Their *resolution* (which redefinition wins a contested target) stays out of this module's scope; field-target resolution is handled here, and operation-target resolution is handled by `crate::model::conformance`'s own field-scoped tools only for the *single*-redefiner case — see "Known gap" below for the contested case.
- Work-unit amounts are computed with `length_amount` throughout, never a bare `as u64` cast; `m + r` and `Σ(c−1)×f(o)` both saturate.

## Known gap (not fixed here, filed separately)

Nothing in this crate detects or resolves *which* operation redefiner wins when two or more operation-redefinition records contest the same inherited operation (`c >= 2`). The field case has a dominance search for exactly this (`apply_redefinitions`'s winner search, `derivation-conflict`/`redefinition-target` refusals); the operation case has no counterpart anywhere in `src/`. The contention is now priced (see `normalize.conflict-check` above), but never detected as a defect or resolved to a winner. This predates this PR and is outside QSL #145's own scope — filed as QSL #173, referenced as `Remaining work: #173` in both `src/model/normalize.rs`'s and `src/model/conformance.rs`'s module docs.

## Tests

- N06's redefinition-check math is pinned to the record's global charge sequence: 3 charges totaling 11 (redef.B `2+0`, redef.C `2+1`, redef.D `4+2`); the exact-bound test's total `work_units` is 100, with its floor at 54 denying exactly at `normalize.conflict-check` with `next_charge == 18`.
- Wide-ancestry coverage: an owner with 128 (and 200) direct generalizations and a single uncontested redefiner completes; the same shape with two contesting redefiners of the same target (`c == 2`, one dominates) also completes and resolves correctly, instead of exceeding `MAX_CONFORMANCE_DEPTH` on breadth alone.
- A new test pins that an operation-member redefinition is charged identically to a field-member redefinition.
- A new test pins that a phase-3 refusal (`specialization-cycle`) wins over a phase-4 refusal (`derivation-conflict`) when a bundle has both defects — this is current behavior, since phase 2/3 always completes as its own pass before phase 4 starts.
- A new test pins that `fixture_n06_conflict` under a tight `derivation_facts` budget (10, then 0) is `Incomplete` at `normalize.fact`, never the `derivation-conflict` or `redefinition-unreachable` refusals the same input produces under an unlimited budget.
- A new test pins that a contested operation-redefinition group (`B` redefining `A.op` with both `B.op2` and `B.op3`) charges exactly one `normalize.conflict-check` of work `4` (`(c-1) * (f(B)+f(B)) = 1 * 4`).

## Out of scope

- Phase-4 redefine facts (`RULE_REDEFINE`) are never charged as `normalize.fact`. This predates this PR and is tracked separately as QSL #169.
- Detecting and resolving a contested operation-redefinition target (QSL #173, see "Known gap" above).

## Verification

- Revert-probes confirmed each fix: reverting the bundle-wide, once-per-record redefinition-check charge back to a per-type computation with `r` reset per type reproduces the exact over-charge this PR fixes (5 charges totaling 15, instead of 3 totaling 11). Reverting the `type_paths`-derived ancestor sets back to a separately-bounded walk reproduces the exact `conformance-depth` refusal on both wide-ancestry tests. Reverting operation members out of the `m` computation drops the operation-redefinition test's total `work_units` from 18 to 16. Removing the `fact_budget_exceeded` gate around phase 4's resolution loop reproduces the exact pre-fix `derivation-conflict` refusal under a tight fact budget. Removing the new operation-group conflict-check charging loop drops its `NormalizeConflictCheck` count to `0`. All confirmed by hand: reintroducing each regression locally reproduces the failure; restoring the fix returns the suite to green.
- Gates, all green: `cargo fmt --check`; `cargo clippy --locked --workspace --all-targets --all-features -- -D warnings`; `cargo test --locked --workspace --all-features` (0 failures across every suite); `make revendor-check` (native-v1: 69 matched/0 drifted/0 stray; complete-value: 64 matched/0 drifted/0 stray).


kreneskyp:
## Review verdict: FIX (head `774eef7b5610c6b4675f69d01d99bfd9ceaa56c2`)

This review covers code, Rust idioms (`agent-skills:rust-review`) and gaps against #145, scoped to this diff. Fix everything below inside #167.

**Gates on head, all green:** `cargo fmt --check` ok · `cargo clippy --locked --workspace --all-targets --all-features -- -D warnings` ok · `cargo test --locked --workspace --all-features` 0 failures · `make revendor-check` native-v1 69/0/0, complete-value 64/0/0.

**Main vs head on the existing fixtures: identical.** I dumped the full outcome Debug plus the admitted charge sequence for `fixture_n06_conflict`, `fixture_n06_resolved`, `fixture_r07_same_owner_contending_redefiners` and `fixture_r07_dominated_owner_takes_no_part_in_the_same_owner_test`, under UNLIMITED. The output was byte-identical. I also swept `derivation_facts` 0..79 and `work_units` 0..79 over the same four fixtures (640 runs), and those were identical too. The equivalence claim still fails outside these fixtures (finding 1).

### Findings (most severe first)

**1. High: a regression. A valid bundle with wide ancestry now refuses, even with no contest.**
- **Where:** `src/model/normalize.rs:1035-1038`, `src/model/conformance.rs:305-315`.
- **Why:** `owner_dominance_closures` runs for every target group, including groups with a single edge. The old `dominates` was never called when `edges.len() == 1`. `ancestor_closure` also counts *distinct nodes visited* against `MAX_CONFORMANCE_DEPTH` (128), not path depth.
- **Scenario:** type `O` with 128 direct generalizations `G000..G127` (depth 1), and one uncontested redefinition `O.z redefines G000.x`.
  - main: `Completed(132 decls)`.
  - head: `Refused(ResourceExhausted, "conformance-depth", "ancestor closure of model.O exceeded 128 generalization steps")`.
  - With n=200: main completes, head refuses.
  - Phase 3 handles this graph fine, because its path depth is 1 < `MAX_GENERALIZATION_DEPTH`.
- **Spec conflict:** FR-150 says normalization exhaustion ends as `resource_exhausted`/`insufficient-next-charge`. A breadth count labelled "depth" is not a `ModelNormalizationLimitsV1` limit.
- **Fix:**
  - Skip the closure computation when `edges.len() < 2`.
  - Stop charging breadth against a depth constant. Preferably, derive each owner's ancestor set from phase 3's already-walked `paths` for `type_key`; every contesting owner is on those paths. That needs no new walk at all. Otherwise, bound the closure walk by path depth the way `ancestor_paths` does.
  - Add a regression test for the 128-wide shape.

**2. High: the core of #145 is not done. The pairwise dominance walk is still uncharged, and the new docs say it should be.**
- **Where:** `src/model/normalize.rs:1292-1294` (`charge_all`), plus the docs at `normalize.rs:66-90` and `normalize.rs:1204-1206`.
- **What the new doc says:** "a pure boolean dominance query has no derivation facts to charge against that budget and no legitimate claim on it."
- **What the spec says:** the vendored `proposals/quire-v1/definitions/value-accounting.md:456` states:
  > each `normalize.conflict-check` | phase 4, after its last `normalize.fact`: each (effective type, redefined member) reached by `c >= 2` redefinition records, ascending by effective member key, before its redefining owners are compared pairwise | no size-counter change; `work_units += Σ (c − 1) × f(o)` instead of one, summed over the `c` redefining owners `o`, where `f(o)` is the number of type derivation facts (qualify and inherit) of `o` that each proper-descendant test of `o` against another owner scans
- **What the code does:** it charges a flat 1 for every group, including groups where `c == 1`.
- **Scenario:** `D` descends from sibling redefiners `B001..B00k`, and each redefines `A.x`.
  - With c=2, two conflict-checks are charged; one of them is `B001`'s single-edge group, which the spec does not charge.
  - Work units are 51, 219 and 891 for c = 2, 8 and 32, the same on main and head.
  - So the O(c²) comparisons are still metered as O(groups). This is exactly the cumulative cost #145 scope item 2 asks to account for.
- **Fix:**
  - Record per group `c` and `Σ f(o)` in `Built`. Here `f(o) = 1 + |phase-3 type inherit facts of o|`.
  - In `charge_all`, charge `normalize.conflict-check` only when `c >= 2`, with `.work(Σ (c−1)×f(o))`.
  - Delete the "no legitimate claim" prose.
- **Related, same phase:** `normalize.redefinition-check` (`:1289-1291`) is also flat 1, where line 455 prices it at `m + r`. Fix it in the same change, or name it explicitly as `Remaining work: #N` in #145's closing note.

**3. Medium: no test backs #145's acceptance.**
- The mutation probe in the PR body only shows that n06/r07 exercise the new code path. No test fails if cumulative accounting regresses, and no test covers finding 1.
- **Add, TC-195 FR-150-AC-8 style:**
  - (a) A many-contender fixture (c ≥ 3, distinct owners). It completes at exactly the hand-computed `work_units`, and the one-less run is `Incomplete` at `normalize.conflict-check`.
  - (b) A single-redefiner group admits no `normalize.conflict-check`.
  - (c) The 128-wide single-redefinition bundle from finding 1 completes.

**4. Medium: the consolidation #145 asked for is not done. There are now two copies of the same walk.**
- **Where:** `src/model/conformance.rs:235-269` (`type_conforms`) and `:293-323` (`ancestor_closure`).
- **What is duplicated:** the stack, visited set, steps counter, ceiling check and refusal construction are copied line for line.
- **Why the stated reason fails:** early exit is a real difference, but it does not need two walkers. One `walk_ancestors(generals, s, |general| -> ControlFlow<()>)` covers both: `type_conforms` breaks on a match, and `ancestor_closure` collects.
- **Scenario:** the next change to the ceiling or refusal shape must be made twice. The detail strings already differ ("conformance check from" vs "ancestor closure of").
- **Fix:** extract one bounded walker and have both functions call it.

**5. Low: the memo covers one target group, not the whole build.**
- **Where:** `src/model/normalize.rs:1026-1038`, and the doc at `:84-88`.
- **What happens:** closures are rebuilt for each (`type_key`, target) group. The same owner is re-walked for every type and every target that reaches it, so the walk count is O(types × groups × distinct owners), not O(distinct owners) as the doc claims.
- **Fix:** hoist the cache to `build()`, or use the phase-3-derived sets from finding 1. Then correct the doc.

**6. Low: stale docs.**
- `normalize.rs:76-81` says "It now delegates to `type_conforms`". It delegates to `ancestor_closure`.
- `normalize.rs:1200-1203` calls it "the same primitive `type_conforms` and dispatch use". It is a second copy of that walk.
- **Fix:** make the docs describe the code.

**7. Low: the PR title and body.**
- The title says "consolidate dispatch dominance walks", but `dispatch.rs` is untouched.
- The body should say plainly that this PR is the #145 work only and that V1-TYPE-028 is unchanged. It currently says V1-TYPE-028 "does not move to complete", but spends a section on wiring and scope outside #145.
- **Fix:** retitle, for example "Phase-4 dominance: shared bounded closure (#145)", and add both statements.
- Closing references are fine: `closingIssuesReferences` = [#145] only, and the commit message has no closing keyword.

**8. Style note (Rust idiom).**
- `conformance.rs:293-297` spells out `std::collections::HashSet` three times; `use` it.
- `visited` duplicates `closure` apart from the root, because pushes are already gated on `closure.insert`.

### Checks with no finding
- **Determinism:** `closures` is used only for lookups. Owners are sorted and deduplicated before walking, so the first owner to exceed the ceiling is stable. No HashSet or HashMap iteration reaches refusal order.
- **Pair loop:** it is O(c²) set lookups per group, as claimed.
- **`Index::build`:** moving it onto `conformance::generals_by_specific` is correct and removes a duplicate.

kreneskyp:
## Re-review verdict: FIX (head `98271c10291470c9a835d2a36192e8f469bb411b`, range `774eef7..98271c1`)

This re-review covers code, Rust idioms and gaps against #145, scoped to the delta. Fix everything below inside #167.

**Gates on head, all green:** `cargo fmt --check` ok · `cargo clippy --locked --workspace --all-targets --all-features -- -D warnings` ok · `cargo test --locked --workspace --all-features` 0 failures · `make revendor-check` native-v1 69/0/0, complete-value 64/0/0.

**Closing references:** `closingIssuesReferences` = [#145] only. Neither commit contains a closing keyword.

**Revert-probes (run locally):**
- Guard disabled (`edges.len() < 1`): all three new tests fail.
- `normalize.conflict-check` back to a flat charge: `n06_conflict_check_charges_exactly_sigma_c_minus_1_times_f_o` fails.
- `normalize.redefinition-check` back to a flat charge: the N06 exact-bound test and the single-redefiner test both fail.

**Existing corpora, main vs head:** f1, f2, n06_conflict, n06_resolved, r07_same_owner and r07_dominated give a byte-identical outcome and non-phase-4 charge sequence. The phase-4 charges differ, by design.

**Resolved from round 1:**
- #3 is resolved; the tests exist and are killed by the probes.
- #4 is resolved; `walk_ancestors` is one walker.
- #5 is resolved; the memo is build-wide.
- #6 and #7 are resolved; the docs, title and body are fixed.
- #8 is resolved.
- #2 is resolved for conflict-check. For redefinition-check it is not (finding 1).
- #1 is only partly resolved (finding 2).

### Findings (most severe first)

**1. High: `normalize.redefinition-check` is charged once per (effective type, record), with `r` reset for each type. The spec charges once per record, with `r` counted globally.**
- **Where:** `src/model/normalize.rs:1116-1130`. It is called from the per-type loop at `:956`.
- **The spec, `value-accounting.md:455`:**
  > each `normalize.redefinition-check` | phase 4, before its first `normalize.fact`: each redefinition record ascending by producer key, before testing that its target is a member inherited by its owning type and that no record checked before it names the same redefining feature with another target | … `work_units += m + r` instead of one, where `m` is the number of effective members of the owning type and `r` the number of redefinition records checked before it
- **Why the spec means once per record:** the unit is the *record*, tested against *its owning type*. Contrast `:456`, which says "each (effective type, redefined member)" when it means per effective type.
- **Scenario, `fixture_n06_resolved`:**
  - Head charges five times: 2 (redef.B at B), 2 (redef.C at C), then 2, 3, 6 at D. Total 15.
  - The spec charges three times: redef.B `2+0`, redef.C `2+1`, redef.D `4+2`. Total 11.
  - The new test pins the wrong count at `tests/model_normalization.rs:1368` (`5`, with a message that rationalises the per-type reading).
  - It also pins `work_units == 104`. By the spec that is 100, and the exact bound is `54`, still denying at `normalize.conflict-check` with `next_charge == 18`.
  - More generally, a record reached by many subtypes is charged once per subtype, not once.
- **Fix:**
  - Compute the redefinition-check sequence once in `build()`, before the per-type phase-4 loop.
  - Sort every redefinition record ascending by key, set `r = index`, and set `m = |effective members of record.owner|`.
  - Drop `redefinition_check_work` from `apply_redefinitions`.
  - Update the test to 3 charges, 100 total, and the 54 bound.

**2. High: the finding-1 regression persists for contested groups.**
- **Where:** `src/model/normalize.rs:1191` (`ensure_owner_closures`) and `src/model/conformance.rs:341` (`ancestor_closure`, which still counts breadth against `MAX_CONFORMANCE_DEPTH`).
- **Scenario:** owner `O` has 128 direct generals `G000..G127`. `O.z` redefines `G000.x`, and `G000.w` also redefines `G000.x` (c = 2, O dominates).
  - main: `Completed(134 decls)`.
  - head: `Refused(ResourceExhausted, "conformance-depth", "ancestor closure of model.O exceeded 128 generalization steps")`.
  - The same happens at n = 200.
  - The `edges.len() < 2` guard only fixes the c = 1 case. The breadth-vs-depth defect named in round 1 is still there.
- **Fix:**
  - Build each owner's closure from the phase-3 paths the build now retains: `type_paths[owner]`, whose `ancestor_key`s are exactly the proper ancestors. That needs no second walk and has no breadth ceiling, and `ancestor_closure` / the `walk_ancestors` closure variant can go.
  - Those paths can be truncated under a tight budget. Skip phase-4 *resolution* when `fact_budget_exceeded(limits, facts_so_far)`, so the result is the correct `Incomplete` that `charge_all` reports at the phase-2/3 facts, never a spurious `derivation-conflict`.
  - Extend `n06_wide_ancestry_*` with the contested 128- and 200-parent shape.

**3. Medium: "exact" pricing covers field redefinitions only. Operation-member redefinition records, and operation members in `m`, are uncharged.**
- **Where:** `src/model/normalize.rs:1098` skips every redefinition record whose redefining or redefined key is not a field member. `member_counts_by_owner` (`:920`) counts field preimages only.
- **What the spec charges:** `:455` charges *each redefinition record*, and `m` counts *effective members*, which includes operations.
- **Scenario:** add one operation member to `B` in N06. The spec's `m(B)` rises by 1, and head charges the same as before. Add an operation redefinition record, and head charges nothing for it; the spec charges `m + r` and shifts `r` for every later record.
- **Fix:** include operation redefinition records in the global finding-1 sequence, and count operation members (direct plus inherited along the phase-3 paths) in `m`. If this rung deliberately excludes them, record that as a scope decision in the module docs, the way the other scope choices are recorded, and drop "exact" from the doc and PR body.

**4. Low: an undeclared change in which refusal wins. It matches the spec, but has no test.**
- **Where:** moving phase 4 to its own pass (`normalize.rs:938-966`) changes which refusal is reported when a phase-4 conflict sorts before a phase-3 refusal.
- **Scenario:** the N06 conflict diamond (A..D) plus a `Y <-> Z` cycle.
  - main: `derivation-conflict` at D.
  - head: `specialization-cycle [model.Y, model.Z]`.
- **Spec:** head is correct per FR-150 ("each normalization phase … reports every refusal it exposes in charge order; a phase that reports a refusal ends checking").
- **Fix:** add this shape as a test pinning "a phase-3 refusal wins over a phase-4 refusal", and state the change in the PR body.

**5. Style notes.**
- `normalize.rs:924` and `:1129` use bare `as u64` for lengths where `charge_all` uses the repo's `length_amount`.
- `m + r` is unchecked while `Σ(c−1)×f` saturates; pick one.
- The 5 s wall-clock assertion in `n06_wide_ancestry_*` (`tests/model_normalization.rs:1485`) belongs in a benchmark lane per rust-review §1. The repo already has one at `:928`, so this is a note, not a gate.

### Checks with no finding
- **The guard is sound for c == 1:** the old winner search with one edge always picks index 0 with no dominance query, which is exactly what the guard does. No refusal path is lost.
- **`f(o)`** = `type_preimages[o].derivation.len()` (1 qualify + one inherit per phase-3 path), summed once per edge. This matches `:456`.
- **Conflict-check** is charged only for c ≥ 2, per (effective type, target), ascending by type then target key.
- **`walk_ancestors`** keeps `type_conforms`'s early-exit order and refusal wording.

### Out of scope, not a finding of this diff
Phase-4 redefine facts (`RULE_REDEFINE`, `normalize.rs:1333/1346`) are never charged as `normalize.fact`, although `value-accounting.md:453` charges phase 4's facts. This predates the PR; it needs its own ticket.

kreneskyp:
## Round-3 verdict: FIX (head `e5e4ad4abe82547883d5275fd1a72a51598b6140`, range `98271c1..e5e4ad4`)

This round covers the fix commit `e5e4ad4`. The merge `5386c0b` of main `af10b0b` does not touch `src/model`. Fix both findings below inside #167.

**Gates on head, all green:** `cargo fmt --check` ok · `cargo clippy --locked --workspace --all-targets --all-features -- -D warnings` ok · `cargo test --locked --workspace --all-features` 0 failures · `make revendor-check` native-v1 69/0/0, complete-value 64/0/0.

**Closing references:** `closingIssuesReferences` = [#145] only. The "Closes #139" in the range comes from main's own `af10b0b` (#161), which is already merged, and #139 is already closed.

**Numbers against `value-accounting.md:455-456`, on N06 resolved:**
- `normalize.redefinition-check` is charged once per record, bundle-wide, ascending by key: redef.B `2+0`, redef.C `2+1`, redef.D `4+2` = 11.
- `normalize.conflict-check` is `2×(2+2+5)` = 18.
- Total 100; the floor of 54 denies at conflict-check with `next_charge` 18. This matches the spec.

**Revert-probes:**
- `r` forced to 0: `n06_conflict_check_charges_exactly_sigma_c_minus_1_times_f_o` fails.
- Operation members dropped from `m`: `operation_redefinition_is_charged_like_a_field_redefinition` fails.
- Round-2 `src` (`98271c1`) run under the head tests: three tests fail, including `n06_wide_ancestry_with_two_contesting_redefiners_completes`.
- `fact_budget_exceeded` gate removed: **all tests still pass** (finding 1).

**Existing corpora, main vs head:** f1, f2, n06_conflict, n06_resolved, r07_same and r07_dom under UNLIMITED give an identical outcome and identical charges outside phase 4. Under tight `derivation_facts` the outcome changes on purpose (finding 1).

**Resolved from round 2:**
- #1 is resolved.
- #2 is resolved: ancestor sets come from phase-3 `type_paths`, the contested 128/200 case completes, and `ancestor_closure` is deleted.
- #4 is resolved.
- #5 is resolved.
- #3 is partly resolved (finding 2).

### Findings

**1. Medium: the `fact_budget_exceeded` gate works, but no test pins it, and it changes outcomes from main without saying so.**
- **Where:** `src/model/normalize.rs:1003` (`if !fact_budget_exceeded(limits, facts_so_far)`).
- **What the gate does:** it turns false refusals into the correct `Incomplete`. Sweep of `derivation_facts` 0..39:
  - `n06_conflict` df 0-9: main (and head with the gate removed) gives `Refused(DanglingReference, "redefinition-unreachable", "redefinition target model.A.x is not an effective member of model.D")`. Head gives `Incomplete`.
  - `n06_conflict` df 10-15: main gives `Refused(derivation-conflict)`. Head gives `Incomplete`.
  - `n06_resolved` df 0-13, `r07_same` df 0-3, `r07_dom` df 0-8: the same pattern.
- **The gap:** removing the gate leaves the whole suite green, so nothing stops it being deleted.
- **Fix:**
  - Add a test: `fixture_n06_conflict()` at `derivation_facts = 10` is `Incomplete` at `normalize.fact` (not `derivation-conflict`), and at `derivation_facts = 0` it is `Incomplete` (not `redefinition-unreachable`).
  - Add one line to the PR body stating that exhaustion in phase 2/3 now wins over any phase-4 refusal (FR-150: "Exhaustion ends checking").

**2. Medium: an operation redefinition group with c ≥ 2 is charged no `normalize.conflict-check`, and its contention is detected nowhere.**
- **Where:** `src/model/normalize.rs:1210-1216` (`apply_redefinitions` skips operation records, so no conflict group forms for them).
- **The spec:** `:456` prices "each (effective type, redefined member) reached by `c >= 2` redefinition records". An operation member is a redefined member.
- **Scenario:** `B <= A`, operations `A.op`, `B.op2`, `B.op3`, with `redef.op2` and `redef.op3` both redefining `A.op`. Result: `Completed`, 0 conflict-check charges. The spec charges `(2−1)×(f(B)+f(B)) = 4` work units at `(B, A.op)`.
- **Also:** TC-196 R07's same-owner contention for operations is refused nowhere.
  - `conformance.rs:52-59` says that shape "is instead detected where a model actually normalizes through it, `normalize`'s own phase 4". Phase 4 is field-only.
  - That detection gap predates this PR and is outside #145.
- **Fix:**
  - Price conflict-check for operation groups too: group every redefinition record reachable at `type_key` by target, and charge `Σ(c−1)×f(o)` for any group with c ≥ 2. Keep *resolution* field-only.
  - Add a test with the scenario above asserting one `normalize.conflict-check` charge of work 4.
  - File the operation-contention detection gap as its own ticket and reference it as `Remaining work: #N` in the module doc at `normalize.rs:20-33` and in `conformance.rs:52-59`, which currently claims normalize detects it.

### Checks with no finding
- **The split is otherwise sound.** Operation redefinition *resolution* stays in `model::conformance` (`check_operation_redefinition`), as the scope decision recorded in the module docs says. The operation `m` count is direct operations plus those inherited along phase-3 paths, deduplicated.
- **The gate logic is correct.** `ancestor_paths` truncates only at `room + 1`, which always makes `facts_so_far` exceed the limit. Because `charge_all` replays the phase-2/3 facts first, a skipped phase 4 is never reached in charge order.
- **Style is fixed:** `length_amount` is used, and `m + r` saturates.

kreneskyp:
## Round-4 verdict: FIX, one Low finding (head `d80d6f1014c5a7fd6bee831e1666a47ac67af374`, range `e5e4ad4..d80d6f1`)

This round covers `f85f2ed` (the round-3 fixes) and the merge `d80d6f1` of main `bd29e7b` (#163). The finding below is small; fix it inside #167.

**Gates, in a fresh clone with its own target dir and debug info off:** `make ci` exit 0 (default-features, all-features and clean-build lanes; 210 test binaries ok, 0 failures) · `make revendor-check` native-v1 69/0/0, complete-value 64/0/0.

**Closing references:** `closingIssuesReferences` = [#145] only. The "Closes #141" in the range is main's own `bd29e7b` (#163).

**Round-3 fixes, verified by revert-probes:**
- Replacing the `fact_budget_exceeded` gate with `if true`: `n06_conflict_under_a_tight_fact_budget_is_incomplete_not_a_phase4_refusal` fails.
- Operation groups no longer charged: `operation_redefinition_group_with_two_or_more_redefiners_is_charged_a_conflict_check` fails.
- `Remaining work: #173` is in both module docs, and the PR body has the "Current behavior" line.

**Merge resolution:**
- `conformance.rs`: main's inline walk in `type_conforms` differed from #167's `walk_ancestors` only by its typed cause. The merged `walk_ancestors` keeps `ModelRefusalCause::ConformanceDepth { from: s.clone() }` and the detail `"conformance check from {s} exceeded 128 generalization steps"`, identical to main. Nothing was lost.
- `normalize.rs`: the `RedefinitionTarget` and `DerivationConflict { type_, member, redefiners }` constructions match main's code and detail format exactly.
- `src/model/refusal.rs` is unchanged against main, so no new variants were minted.
- Main's `tests/model_normalization.rs` assertions are kept verbatim; only one import line differs.

**Main vs head on the existing corpora:** f1, f2, deep_parallel_generalization_chain, n06_conflict, n06_resolved, r07_same and r07_dom under UNLIMITED give a byte-identical outcome Debug, `cause.as_str()`, detail string, typed cause and non-phase-4 charge sequence. The tags are `derivation-conflict`, `redefinition-target` and `redefinition-target`, the same as main.

### Finding

**1. Low: within one type, operation conflict-checks are charged after every field conflict-check, not interleaved by key.**
- **Where:** `src/model/normalize.rs:1335` (field target loop), then `:1576-1598` (operation target loop), inside the same `apply_redefinitions` call.
- **The spec:** `value-accounting.md:456` orders conflict-checks "ascending by effective member key".
- **Scenario (probed):**
  - Setup: `C <= B <= A`. Field `A.z` is redefined by `B.z1` and `C.z2`, which charges `(2−1)×(f(B)+f(C)) = 5`. Operation `A.a` is redefined by `C.a2` and `C.a3`, which charges `(2−1)×(f(C)+f(C)) = 6`.
  - Because `A.a < A.z`, the spec charges 6 then 5.
  - Head charges 5 then 6: at `work_units = 58` it denies with `next_charge = 5`, where the spec gives 6.
  - This only shows at an exact bound (FR-150-AC-8), and only for a type with both contested fields and contested operations.
- **Fix:**
  - Collect the per-type `(target_key, work)` conflict-check amounts from both loops into one list, sort it by target key, then push it to `conflict_check_work`.
  - The field loop may still `return Err` on a refusal first, which is fine.
  - Add the scenario above as a test pinning `next_charge == 6` at the bound.

### Style note
`normalize.rs:1353` and `:1585` use `edges.len() as u64`; the rest of the file uses `length_amount`.

kreneskyp:
## Round 5 verdict: MERGE at `ba903ba188734d742b15a998a83291c15496f96e`

Scope: delta `d80d6f1..ba903ba`, which is fix commit `93b8599` plus a merge of main `bd59852` (#168).

### Round-4 Low finding (conflict-check order): fixed
- `src/model/normalize.rs:1346`: both loops now collect `(target_key, amount)` into `conflict_charges`. The combined list is sorted once and pushed at `:1616-1619`. This is `value-accounting.md:456`'s single ascending pass by effective member key, with field and operation targets interleaved.
- The sort key matches the spec. Within one effective type, the redefined member's producer key is the effective member key. The outer per-type loop keeps its order, as accepted in earlier rounds.
- Ties can't happen. Field groups and operation groups are each keyed by their own distinct target, and no record is both a field and an operation redefinition (`:1296-1305`). Whether `sort_by` is stable therefore doesn't affect the result.
- The total is unchanged, since the same amounts are pushed in a new order. Refusal behaviour is also unchanged. A field-loop `return Err` still wins, and because `build()` exits through `?` (`:1130-1138`), the whole `Built` is dropped. Buffered charges never reach `charge_all`.
- `length_amount(edges.len())` is now used at `:1366` and `:1599`. The style note is resolved.

### New test
- `tests/model_normalization.rs:2014`, `conflict_check_charges_interleave_field_and_operation_groups_by_target_key`. Its numbers match my round-4 probe: a total of 89, and at wu=58 it is Incomplete at `NormalizeConflictCheck` with next_charge=6.
- Revert-probe: I commented out the `sort_by` so the old field-then-operation order came back. The test fails at `:2059` with `left: 5, right: 6`. I then restored the source.

### Merge of main `bd59852` (#168)
- `git diff bd59852 ba903ba` touches only the PR's own 3 files: `normalize.rs`, `conformance.rs` and `tests/model_normalization.rs`. #168 changed none of those 3 files, so nothing from #168 is dropped or changed.

### Gates at the head
- `make ci`: exit 0. That covers fmt, clippy default and all-features with `-D warnings`, the full test suite, and a no-default-features build.
- `make revendor-check`: native-v1 69/0/0, complete-value 64/0/0.
- `closingIssuesReferences`: only #145.

### Nit (does not block merging)
- The finding labels in comments are wrong. The code comment at `normalize.rs:1345` says "review finding #1", and the test doc at `tests/model_normalization.rs:1977` says "round 5, finding #1". It was the round-4 Low finding. This does not affect behaviour.

No open findings.

Issue #166, opened 2026-09-18
Model lookup: one owner for the key → universe → absence order


kreneskyp:
Found in the review of PR #158 (round 4: https://github.com/agent-ix/quire-spec-language/pull/158#issuecomment-5734532582). Part of #120.

The FR-153 lookup order (S-conforms-to-T, then the `lookup.key` charge, then the universe check, then membership or absence) is implemented twice:
- `population::lookup` (src/model/population.rs);
- `evaluate_unresolvable_lookup` (src/value/model_query.rs), which handles references whose identity is not valid UTF-8.

They agree today. No test compares them, so a later change to one would drift silently.

Done when either:
- (a) the unresolvable-key decision moves into `population::lookup` itself (e.g. a lookup key whose object is `Member(String)` or `NonMember`), so `evaluate_unresolvable_lookup`, the `pub(crate) conforms` wrapper and its `#[allow(clippy::too_many_arguments)]` are deleted; or
- (b) a table test runs well-formed and malformed references side by side (the three absence modes × {absent in own universe, foreign universe} × budgets {unlimited, 0, 1}) and asserts identical outcome and meter.

(a) is preferred.

Also in scope:
- The `foreign-universe` refusal is built in three places. Give it one constructor.
- `bridge_reference` clones `universe` inside a `map_err` closure; a `match` would move it.
- The `type_catalog` field doc in src/model/population.rs points to `crate::value::expression::model_query`. The module is `crate::value::model_query`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #164, opened 2026-09-18
[QSL#164] value's TypeEnvironment island: no model generalization data at check time


kreneskyp:
Follow-up from [#120](<https://linear.app/agent-ix/issue/AGE-323/v1-a04-implement-model-graph-binding-closed-lookup-inheritance-and>) (A04 slice 1a, PR [#158](<https://github.com/agent-ix/quire-spec-language/issues/158>)'s review at [https://github.com/agent-ix/quire-spec-language/pull/158#issuecomment-5732931271](<https://github.com/agent-ix/quire-spec-language/pull/158#issuecomment-5732931271>)).

`crate::value`'s `TypeEnvironment` (`src/value/composite.rs`) admits record, tuple and object-type declarations, but carries no generalization/conformance graph — that data lives only in `crate::model::conformance`/`crate::model::bundle`, reachable from a `PopulationBinding` (a runtime value), never from the checked package's static types. Three FR-153/FR-149 obligations need exactly that data at check time and cannot get it without teaching `TypeEnvironment` the model's generalization graph:

1. **FR-153-AC-6 / TC-198 L08 upcast equality.** `check_equality` can only compare `Reference<A>` and `Reference<B>` structurally, so `(lookup<M::A>(p, rb) absent refused) = rb` (comparing a value now known to hold a `B` against another `Reference<B>`) is refused at check time as `ill_typed`/`type-mismatch`, even though both operands denote the same real object.
2. `deref(r).f` **display-name resolution** (TC-198 L11, FR-043 territory). `ObjectEnvironment::attribute` resolves fields by the reference's *runtime* most-specific type's declaration, with no inherited-member fallback, so a `Reference<A>` presently holding a `B` cannot resolve a field `A` declares but `B` doesn't override.
3. `lookup<T>(p, r)` **check-time** `S` **vs** `T` **conformance** (finding 4 of the [#158](<https://github.com/agent-ix/quire-spec-language/issues/158>) review). `r: Reference<S>` where `S` does not conform to `T` is `ill_typed`/`type-mismatch` per FR-153, but the checker admits it; the refusal only appears at evaluation, inside `crate::model::population::lookup`'s own `type_conforms` check.

[...]

Issue #147, opened 2026-09-18
[QSL#147] FR-151-AC-7 and FR-152-AC-1/3/5/7: dispatch preconditions, qualified-name binding and runtime navigation


kreneskyp:
Five acceptance criteria owned by FR-151 and FR-152 have no test, because the machinery each one asserts over does not exist yet in `crate::model`. Split out of #144, which backed FR-151-AC-1 and measured the rest.

This is a scope boundary, not a deferral of work that fits elsewhere: each criterion below names behaviour that must be built before a test can assert its content. Writing a test that passes without it would be the "criterion too weak to fail" defect.

## Unbacked criteria

| Criterion | Requires |
| --- | --- |
| FR-151-AC-7 | A false selected precondition yields `undefined` and never selects a less specific method |
| FR-152-AC-1 | References to every systems-model kind resolve with exact producer and effective-declaration identity |
| FR-152-AC-3 | Missing relationship closure or an ambiguous effective target refuses without ambient lookup |
| FR-152-AC-5 | Relationship navigation never crosses correspondence bundles or substitutes display-name equality for identity |
| FR-152-AC-7 | Runtime navigation returns the multiplicity-determined type in canonical reference-key order and charges exactly the `model.navigate` schedule |

## Missing machinery

Three pieces, each confirmed absent in the tree at `514c759`:

1. **A qualified-name binder** (`M::Sys::pump`-style) and an effective-declaration identity per systems kind. `src/model/systems.rs` states that `crate::model::normalize` does not fold components, endpoints or relationships into its `EffectiveView`.
2. **A bridge from FR-146 expression evaluation into `crate::model`.** `src/model/dispatch.rs` states precondition evaluation is out of scope for the module. `src/value/expression/` carries the evaluator (`CheckedPackage`, `Machine`, `Callable`) and nothing connects the two in either direction.
3. **A runtime object population to traverse**, which is FR-153 territory. `src/model/accounting.rs:125` states the module does not implement `systems.resolve` or `model.navigate`.

Consequent additions the criteria name directly:

- `ChargePoint` has 15 variants and none is `ModelNavigate`, `DispatchSelect`, `SystemsResolve`, `CollectionVisit`, `CollectionBound` or `CollectionResultRetain`. FR-152-AC-7 charges a schedule that has no charge point.
- `Code` has no `CardinalityOutOfBound` or `ForeignReference` variant.
[...]

PR #168, opened 2026-09-18
A04 slice 2: model-population anchors and operation frame (V1-EXPR-020/021/022)

kreneskyp:
## Gap (read before code: FR-150-153, TC-195/198, `src/model/population.rs`, `src/value/expression/*.rs`)

At base `8445da9` (#157 `PopulationMember.field_values`, #158 model references/`allInstances`/closed lookup), the Complete-V1 model layer had no concept of an invocation's pre state at all:

- `src/model/population.rs`: `PopulationBinding` had no `pre_anchor`/pre-state field (checked in full before this PR). No function admitted a pre+post document pair; nothing computed created/deleted identities or checked an `OperationEffect` frame against them.
- `src/value/expression/{syntax,ir,check,evaluate}.rs`: no `pre(..)` form anywhere — `Expression`/`NodeKind` had no variant for it, so `pre(allInstances(p))`/`pre(lookup(p, r) absent m)` (FR-153, TC-198 L07) could not be written or evaluated at all.

So for the three rows in scope:
- **V1-EXPR-020** (postconditions against exact pre/post anchors): no anchor mechanism existed — `allInstances`/`lookup` had exactly one population to read, always "now".
- **V1-EXPR-021** (captures/aliases without anchor drift): nothing to drift, since there was no anchor.
- **V1-EXPR-022** (operation frames, created/deleted identities): no frame enforcement existed; nothing computed a created/deleted diff.

## What this PR adds

- `src/model/population.rs`: `PopulationBinding` gets a private `pre_anchor: Option<Arc<PopulationBinding>>` field, a consuming `with_pre_anchor` builder, and a public `pre_anchor()` accessor. `admit_invocation(InvocationContext, pre_document, post_document, &InvocationDelta, pre_meter, post_meter) -> AdmissionOutcome` admits both documents via `admit_binding`, then runs `enforce_frame`, which:
  - classifies created/survivor/deleted **by bare object identity** (never by `ReferenceKey`, which also carries type) and separately refuses any object whose most-specific type differs pre to post — `Code::FrameViolation`/`unauthorized-change`, since no FR-151 grant (`fieldWrites`/`creates`/`deletes`) authorizes a retype, and FR-143 binds the most-specific type into a reference's own identity;
  - compares a surviving field's pre/post values by its declared `Multiplicity.ordered` (exact sequence vs. order-insensitive multiset), so reordering an unordered field's values is never a write;
  - decides field-write coverage by direct `fieldWrites` match **or** a bounded, cycle-safe walk along the *full* redefinition chain (`redefinition_reaches`) — `model-complete.md:56`'s "the redefining feature replaces *the one* inherited redefined feature" means a chain like `C.x → B.x → A.x` is legal and normal, with no direct `C.x → A.x` record, so a single hop is not enough; `redefinitionClosure: closed` (`model-complete.md:64`) means every edge is *listed* in the bundle, not that the chain is pre-flattened. `redefinition_reaches` is written generic and reusable (not tied to `OperationEffect`) so `crate::model::conformance`'s identical one-hop gap (`check_operation_redefinition`, `conformance.rs:583`) can reuse it — tracked, not fixed here: QSL #171;
  - finally checks FR-046's own delta agreement: the invocation's caller-supplied `declared_created`/`declared_deleted` (the new `InvocationDelta` struct) must be internally consistent (no duplicate, no identity in both lists) and equal the complete computed created/deleted sets — any disagreement is `Code::PopulationDeltaMismatch`/cause `delta-disagreement`.
- `src/value/expression/{syntax,ir}.rs`: `Expression::Pre`/`NodeKind::Pre`, an identity-typed wrapper.
- `src/value/expression/check.rs`: `Typer` now carries a `postcondition: bool` (threaded through every `Typer::new` call site). `pre(...)` refuses, in order:
  1. `wrong_snapshot`/`forbidden-pre-read` if not in a postcondition context (FR-208 applies FR-042 to invariants/preconditions and names this same cause explicitly — this is a checker-decided refusal over `pre(...)`'s own syntax, never the runtime-only `wrong-anchor` cause);
  2. the same cause if its operand's own syntax contains no eligible read (`allInstances`/`lookup`/nested `pre` — never a bare `Name`/literal on its own; `Call`'s own *arguments* are still reached through `Expression::children()`'s ordinary recursion, but a call itself is no longer direct-eligible, so `pre(F(p))` over a bare-parameter argument refuses while `pre(F(allInstances(p)))` stays legal, FR-042-AC-1's `pre(P(self.version, delta))` analogue) — via `contains_pre_eligible_read`;
  3. the same cause if the eligible read's population operand is a bare `Name` resolving to a `let`-bound alias captured *before* the `pre(...)`'s own boundary (`let q = p in pre(size(allInstances(q)))` refuses, FR-042-AC-3's `let s = self in pre(s.version)` analogue) — via the new `Typer::contains_captured_pre_alias`, which distinguishes a captured `let` alias from a plain parameter reference using the new `LocalKind` (`Parameter` vs `Bound`) tag on every `Typer::bind` call. Alias status carries transitively through a chain of `let`s introduced *inside* the same `pre(...)` operand (`let q = p in pre(let r = q in size(allInstances(r)))` also refuses: `r`'s own binding is itself a bare reference to the already-captured alias `q`, so `contains_captured_pre_alias` marks `r` as an alias too, not as a fresh, unrelated shadow) — a rebinding to a genuinely fresh, non-alias value (say `allInstances(other)`) still stays exempt inside its own body.

  `CheckedPackage::check_postcondition_expression` is the new public entry point for a real postcondition context; plain `check_expression` stays non-postcondition.
- `src/value/expression/evaluate.rs`: `NodeKind::Call` now pushes `Task::RestoreAnchor(self.anchor)` and resets to `Anchor::Post` before evaluating a callee's body, so `pre(F(p))`-shaped calls never leak the caller's pre anchor into the callee's own body. Separately, `select_anchor`'s missing-pre-anchor case (a population admitted through `admit_binding` directly, never through `admit_invocation`) is now the real, caller-input-reachable `Refusal::WrongSnapshot(WrongAnchor)`, not `Refusal::CheckedInvariant` — the checker cannot see, at checking time, which admission path a caller's runtime argument will take, so this is not a broken-evaluator invariant.
- `src/value/expression/refusal.rs`: new `WrongSnapshotCause { WrongAnchor, ForbiddenPreRead }` and `CheckCause::WrongSnapshot`, wired to `Code::WrongSnapshot`. The two causes split by *layer*, not by clause: every checking-time decision over `pre(...)`'s own syntax is `ForbiddenPreRead`; `WrongAnchor` is reserved for the one case the checker cannot see at all, `select_anchor`'s runtime binding check.
- `src/value/outcome.rs`: new `Refusal::WrongSnapshot(WrongSnapshotCause)`, reusing the checker's own closed cause set rather than inventing a separate runtime cause type.

## Backed and by which tests

- **V1-EXPR-020** (FR-153-AC-7, TC-198 L07; **partial** — see Honest scope boundaries): `tests/model_reference_queries.rs` `l07_pre_all_instances_reads_the_invocation_pre_population`, `l07_pre_lookup_reads_the_invocation_pre_population_and_a2_keeps_its_pre_type`, `pre_anchor_does_not_leak_into_a_sibling_post_anchored_query`, `nested_pre_is_idempotent`, `pre_refuses_outside_a_postcondition_context`, `pre_refuses_a_bare_population_parameter`, `pre_of_an_integer_literal_is_refused`, `pre_of_a_function_call_over_a_bare_parameter_argument_refuses_forbidden_pre_read`, `pre_of_a_function_call_over_an_eligible_read_argument_stays_legal`, `pre_of_a_binding_with_no_pre_anchor_refuses_wrong_anchor`. Model-layer half in `tests/model_population.rs`: `l07_invocation_admits_a_declared_delete_and_attaches_the_pre_anchor`.
- **V1-EXPR-021** (captures/aliases without anchor drift; **partial** — population-rooted aliases only, see Honest scope boundaries): `tests/model_reference_queries.rs` `pre_refuses_a_let_bound_query_result_capture_drift` (a `let`-bound *query result* re-tagged `pre`), `pre_refuses_a_let_bound_population_alias_capture_drift` (a `let`-bound *population alias*, `let q = p in pre(size(allInstances(q)))`), `pre_refuses_a_let_bound_alias_of_a_let_bound_population_alias` (an alias *of* an alias, `let q = p in pre(let r = q in size(allInstances(r)))`, transitively), and `pre_of_a_captured_post_reference_retains_its_post_observation` (QSpec FR-012-AC-4's positive retention half: `let v = size(allInstances(p)) in pre(size(allInstances(p)) != v)` -- `v`, captured at post before `pre(...)` runs and never itself a bare population alias, keeps observing the post value even though it is read inside `pre(...)`).
- **V1-EXPR-022** (FR-046-AC-3 frame enforcement + FR-046 delta agreement): `tests/model_population.rs` `l07_invocation_refuses_a_delete_outside_the_declared_frame`, `invocation_refuses_a_create_outside_the_declared_frame`, `invocation_field_write_outside_the_declared_frame_refuses_and_inside_it_admits`, `enforce_frame_admits_an_unordered_field_reorder_without_a_write`, `enforce_frame_refuses_an_ordered_field_reorder_as_a_write`, `enforce_frame_admits_a_field_write_that_reaches_a_declared_grant_through_redefinition`, `enforce_frame_admits_a_field_write_that_reaches_a_declared_grant_through_a_redefinition_chain` (the two-hop `C.x → B.x → A.x` case), `enforce_frame_refuses_an_object_that_changes_type_between_pre_and_post`, `invocation_admits_a_subtype_created_and_deleted_under_a_supertype_grant`, `invocation_refuses_a_declared_delta_that_disagrees_with_the_complete_populations`, `invocation_refuses_a_declared_delta_that_declares_the_same_identity_created_and_deleted` (now asserts the overlap check's own detail text, so disabling that branch specifically — not just any `delta_mismatch` branch — turns it red).

Every new guarding test carries its own mutation-probe description in its doc comment: the specific edit made, the observed red result, and the revert.

## FR-046 duplication ruling

`src/runtime/validation/frames.rs`'s `inspect_frames`/`object_frame`/`declared_deltas` already decide the identical two violations over the native runtime's own budget-metered `Validator`/`ValueId` arena/`ObjectIdentity` types. Those types are foreign to `crate::model` (`Validator` is `impl`-private and tightly coupled to its own budget/arena plumbing) — there is no call this PR's `enforce_frame`/`check_declared_delta` could make into it, only a parallel decision over `crate::model::population`'s own `PopulationDocument`/`PopulationMember`/`AdmissionMeter` types. What is shared, deliberately: every cause. This PR reuses `Code::FrameViolation`/`"unauthorized-change"` and `Code::PopulationDeltaMismatch`/`"delta-disagreement"` for the same violations `frames.rs` names them for, and the equivalent decision shape (declared-identity duplicate, declared-set intersection, computed-vs-declared mismatch), never a differently shaped one. Full reasoning is in `admit_invocation`'s own "Native duplication" doc.

## Honest scope boundaries (not silently dropped)

- **V1-EXPR-020 is partial: `result`**. `Expression` (`src/value/expression/syntax.rs:81`) has no `Self_`/`Result` variant, and nothing in this crate's value layer defines a checked "operation" declaration with a result-type binding contract to bind `result` to (`crate::model::bundle`'s `PostconditionClause` is a separate model-layer structure, never lowered through this `Expression`/`Typer`/`Node` pipeline). Tracked as its own issue, refs #170, not fixed here.
- **V1-EXPR-021 is partial: population-rooted aliases only; `self`-rooted aliases wait on #170**. This PR backs the refusal of a re-tagged capture and of a direct (or transitively chained) `let`-bound population alias, plus the positive retention half for a non-alias captured value, all rooted in the `p` parameter. QSpec FR-012's `self`-rooted shapes (`let v = self.version in pre(v)`, `let s = self in pre(s.version)`) need `self`, which this checker has no textual form for yet (see the `self`/`state-clause` bullet below); refs #170, not fixed here.
- **QSL #171** (not fixed here): `crate::model::conformance`'s `check_operation_redefinition` (`conformance.rs:583`) has the identical one-hop redefinition-chain gap this PR fixes in `field_write_covered`; `redefinition_reaches` is written generic so #171 can reuse it directly.
- Removed a stale test comment in `tests/model_population.rs` (from PR #148) claiming FR-153-AC-7 was "tracked on QSL #147" — #147 is actually about FR-151-AC-7/FR-152-AC-1/3/5/7, an unrelated gap. FR-153-AC-7 is backed here by the tests above. Remaining work: #147 stays open for its own scope.
- The TypeEnvironment island (#164), duplicated lookup order (#166), `dispatch.rs`/`normalize.rs` (slice 3): untouched, as before.
- `self` and the full textual `state-clause` grammar/parser: still out of scope — no textual parser exists for it in this checker.
- QSpec #111 (the FR-012-AC-7 vs FR-042-AC-4 upstream spec conflict): out of scope for this PR — this PR keeps following the vendored FR-042 as written, and does not resolve the conflict.

## Gates (all green, worktree-local `CARGO_TARGET_DIR`)

- `cargo fmt --check`: clean
- `cargo test --locked --workspace --all-features`: all green, 0 failed
- `make revendor-check`: clean

Part of #120


kreneskyp:
## Review verdict: FIX (head `dac1152ba48341c4ad64bc983c3c7b5eab4359a9`)

This is the single review for #168. It covers a code review, a Rust review (`rust-review`, with the repo's conventions taking priority) and a gap analysis against the vendored spec, scoped to this PR's diff.

### Gates (run in a fresh clone at dac1152)
- `cargo fmt --check`: clean
- `cargo test --locked --workspace --all-features`: all green, 0 failed
- `make revendor-check`: `native-v1: 69 matched, 0 drifted, 0 stray`; `complete-value: 64 matched, 0 drifted, 0 stray`

### Revert probes (re-run by the reviewer)
- `select_anchor` `Anchor::Pre => Ok(binding)`: both `l07_pre_*` tests and `pre_anchor_does_not_leak…` went red. Confirmed.
- Removed the `Task::RestoreAnchor` push: `pre_anchor_does_not_leak_into_a_sibling_post_anchored_query` went red. Confirmed.
- Field-write guard forced to `continue`: `invocation_field_write_outside_the_declared_frame_…` went red. Confirmed.

The existing tests are real gates. The findings below cover inputs that the tests never exercise.

### Behavioural probes (local-only tests, not pushed)
| Probe | Expression / input | Result at dac1152 | Spec says |
| --- | --- | --- | --- |
| A | `let q = pre(p) in allInstances<M::A>(q)` over L07 | `Completed`, 2 elements (the **post** population) | FR-042: `pre` on a bare parameter refuses |
| B | `let v = allInstances<M::A>(p) in pre(v)` | `Completed`, 2 elements (the post value, now presented as pre) | FR-042-AC-3: refuse instead of retagging a post capture |
| C | `pre(allInstances(p))` where `p` came from plain `admit_binding` | `Refused(CheckedInvariant)` (runtime_invariant) | a checker refusal, `wrong_snapshot`/`forbidden-pre-read` |
| D | `pre(1)` | `Completed(1)` | FR-042-AC-4: `pre` on a constant refuses |
| E | `pre(pre(allInstances(p)))` | 3 elements | correct (FR-042-AC-2 idempotent) |
| F | survivor `a1.x` `[a1,a2]` → `[a2,a1]` on an unordered field, empty effect | `FrameViolation` | no change, so it should admit |
| G | delete `b1: B` under `deletes: [model.A]` | Admitted | correct (FR-151: a subtype is covered by a supertype grant), but no test covers it |
| H | `a1: A` pre → `a1: B` post, `creates/deletes: [model.A]` | **Admitted** | the object changed type while keeping its identity; this must not pass as a delete plus a create |
| I | H with an empty effect | `FrameViolation` "creates object a1" | the refusal reason is wrong |

### Findings, most severe first

**1. High: `pre(e)` accepts operands the spec refuses, and silently returns post values as pre.** `src/value/expression/check.rs:554`, `src/value/expression/evaluate.rs:457`
- Failure: probes A, B and D. `let v = allInstances(p) in pre(v)` completes with the post population. This is exactly the anchor drift that V1-EXPR-021 exists to forbid, and FR-042-AC-3 requires a refusal. `pre(p)` on a bare population parameter is accepted and has no effect. `pre(1)` is accepted.
- Why: the typer's `Expression::Pre` arm only infers the operand's type. Nothing checks eligibility.
- Fix: in `Typer::infer_form`, refuse `pre(e)` with `wrong_snapshot`/`forbidden-pre-read` in these cases:
  - `e` contains no eligible state read (`AllInstances`/`Lookup`);
  - `e` is a bare `Name` or literal;
  - `e` reads a local whose binder sits **outside** the `pre(..)` and whose type is `Population`, `Reference` or contains one (FR-042-AC-3). A `let` bound inside the `pre` stays legal (FR-042-AC-2).
- Add one test per refusal shape: A, B, D and `let s = … in pre(s …)`.

**2. High: `pre` is admitted in every context, and the anchor is dynamically scoped across function calls.** `src/value/expression/evaluate.rs:334-345`, `evaluate.rs:671-683`
- (a) Context. Nothing restricts `pre(..)` to a postcondition body. FR-153 says "in a postcondition"; FR-042 refuses it in invariants and preconditions; shared-grammar.md:500 makes it unavailable inside a reusable predicate. Probe C shows the result: a checked program reaches `Refused(CheckedInvariant)` at runtime. The `select_anchor` doc says "the checker only admits `pre(..)` where the population operand carries one", but no such check exists. That doc describes behaviour the code does not have.
- (b) Calls. `NodeKind::Call` pushes a frame but never saves or resets `self.anchor`. So in `pre(F(p))`, where `F(q) = size(allInstances<M::A>(q))`, F's body reads the pre population. That hands a reusable body a hidden pre meaning, which shared-grammar.md:500-503 forbids.
- Fix:
  - thread a clause/anchor context into the checker, and refuse `pre` outside a postcondition with `wrong_snapshot`/`forbidden-pre-read`;
  - on `Call`, push `RestoreAnchor(self.anchor)` and set `Anchor::Post` for the callee body. If the argument should carry the pre instant, resolve it when the argument value is produced, not through ambient state;
  - correct the `select_anchor` doc.
- Add tests for (a) and (b).

**3. Medium: the PR body over-claims the rows.**
- V1-EXPR-020 is "postconditions against exact pre/post anchors **and result**". `result`/`self` are explicitly out of scope, and there is no postcondition context at all (finding 2a), yet the body lists EXPR-020 under "Backed".
- V1-EXPR-021 is "captures and aliases without anchor drift". Only sibling leakage is tested, and probe B shows capture drift is unguarded.
- V1-EXPR-022 includes "created identities and deleted identities". The recorded delta is skipped (finding 4).
- Fix: after 1, 2 and 4 land, restate the "Backed" section per row as partial where it is partial. At minimum, EXPR-020 must read "pre/post population anchors only; `result`/`self` anchors not implemented", and neither EXPR-020 nor EXPR-022 may be presented as fully backed.

**4. Medium: the reason given for skipping `delta-disagreement` is wrong.** `src/model/population.rs:808-813` and the PR body
- FR-153's Inputs define no delta, which is true. But the invocation-level spec does define one. Vendored native-v1 FR-046 says: "Complete pre/post populations determine actual create/delete/change inventories; **caller-supplied lists must match** and stay within the selected frame". FR-046-AC-3 requires "distinct typed causes" for wrong inventories versus out-of-frame changes, and native-diagnostics.md:94 retains "the declared delta/frame permission".
- The repo's own native runtime already implements this: `src/runtime/validation/frames.rs:137-178` covers duplicate and intersecting declared deltas, and `frames.rs:332-339` covers `population_delta_mismatch`.
- Failure: an invocation that declares `deleted: []` but actually deletes `a2` under a `deletes: [A]` frame is admitted by `admit_invocation`, where native refuses `delta-disagreement`.
- Fix: make `admit_invocation` take the invocation's declared created/deleted identity sets. Refuse `PopulationDeltaMismatch`/`delta-disagreement` when they differ from the computed post\pre and pre\post sets, or when they intersect. Test both cases.

**5. Medium: a survivor that changes type is admitted as a delete plus a create.** `src/model/population.rs:879-920` vs `:924-953`
- Why: created and deleted identities are keyed by `ReferenceKey`, which includes the most-specific type, while survivors are keyed by the object string alone. These are two different notions of identity.
- Failure: probe H admits `a1: A → a1: B`. The same object identity counts as both created and deleted, its fields are also compared as a survivor, and the refusal in probe I names the wrong reason ("creates object a1").
- FR-153 fixes a member key by its most-specific type, and a deleted object keeps its pre type. V1 has no retyping, and the native path treats created∩deleted as a delta error.
- Fix: compute created and deleted by `(universe, object)`. A survivor whose most-specific type differs between pre and post refuses; `invalid_runtime_input`/`conflicting-identity` is the catalogued cause for one identity carrying two types. Add a test.

**6. Medium: reordering an unordered field reads as an unauthorized write.** `src/model/population.rs:944`
- `values_of(pre) == values_of(post)` compares `Vec<String>` in order. Probe F refuses `[a1,a2] → [a2,a1]` on a field with `ordered: false` (F1's `model.A.x`), so an operation that changes nothing is refused `frame_violation`.
- Fix: look up the field's `Multiplicity`. When the field is unordered, compare values as a multiset, or as a set when `unique`. Add a test.

**7. Medium: `fieldWrites` matching ignores redefinition.** `src/model/population.rs:947`
- `write == field` is exact key equality. FR-151's `quire.model.conformance.effect/v1` says a written field is covered when it "reaches one through redefinition records", and adds "the same inclusion governs FR-013 frames under inheritance".
- Failure: a frame granting `model.A.x` refuses a `B` member's write to `model.B.x`, where `B.x` redefines `A.x`.
- Fix: resolve the written field through the bundle's redefinition records before matching. Add a test with a redefinition fixture.

**8. Low: test gaps and trace tags.** `tests/model_population.rs:1449,1479,1520`
- Missing tests:
  - a subtype create or delete covered by a supertype grant (probe G passes, but nothing pins it);
  - nested `pre(pre(..))` (probe E);
  - a `pre` whose anchor is absent.
- The runtime frame tests trace to bare `FR-151`. FR-151's effect rule is link-time inclusion (`effect-escape`), not runtime frame enforcement. Trace them to the criterion they actually back: FR-046-AC-3, or the row's central requirement.

**9. Low: docs and idioms.**
- The `src/model/population.rs` `//!` header does not mention invocation admission or `pre_anchor`, although the module now owns both.
- `evaluate.rs` does not say that anchor switching is uncharged. `quire.value.accounting/v1` defines no charge point for it, so say so, as `admit_invocation`'s docs already do for the frame check.
- `enforce_frame` deduplicates fields with a `Vec` and a linear `any` (O(F²)); use a `BTreeSet<&ProducerKey>`.
- The two `let mut allowed … for … break` loops can become `try_fold`, or a small helper returning `Result<bool, _>`.
- `Arc<PopulationBinding>` for the pre anchor with a private `with_pre_anchor` is fine: it gives cheap `Clone` and only `admit_invocation` can attach an anchor.

**Note (outside the Fix list): two frame engines.** `enforce_frame` is a second frame engine next to `src/runtime/validation/frames.rs`, and it differs on delta, intersection, State roots and charging. Findings 4 and 5 close the behavioural gap. Neither #164 nor #166 covers this.

### Spec-fidelity answers
- **`pre(e)` form:** this is the spec's own form, not an invention. It appears in shared-grammar.md:500 and :604, in FR-153's anchor paragraph ("`pre(allInstances<T>(p))` reads the invocation's pre population"), and in native-v1 FR-042. The meaning is only partly implemented; see findings 1 and 2.
- **`unauthorized-change`:** correct. native-diagnostics.md:94 pairs `frame_violation` with `unauthorized-change`.
- **Skipping `delta-disagreement`:** the justification does not hold; see finding 4.

### Issue refs
- **#147:** do not close it. #147 tracks FR-151-AC-7 and FR-152-AC-1/3/5/7 (dispatch preconditions, qualified names, runtime navigation). This PR touches none of them. The old `tests/model_population.rs` doc wrongly pointed FR-153-AC-7 at #147, and this PR removes that pointer, which is right.
- Keyword line for the body: `Part of #120`, with no `Closes`/`Fixes`. Optionally add a plain line: "Removes the stale FR-153-AC-7 → #147 pointer; #147 stays open."

All findings are fixed inside #168.

kreneskyp:
## Re-review verdict: FIX (head `95ccccef875d594b851146cd04526d3258755cab`)

This is the single re-review for #168, covering `dac1152..95cccce`. It includes a code review, a Rust review and a gap analysis. Most of the earlier findings are fixed properly. Three problems remain in the new code:
- `pre(F(p))` is accepted and returns the post value. A test currently enshrines that behaviour.
- The refusal cause for `pre` outside a postcondition is the wrong one.
- One redefinition hop does not cover chained redefinitions, which this normalizer accepts.

The PR body also still presents V1-EXPR-020 as backed.

### Gates (fresh clone at 95cccce)
- `cargo fmt --check`: clean
- `cargo test --locked --workspace --all-features`: 108 result blocks, 0 failed
- `make revendor-check`: `native-v1: 69 matched, 0 drifted, 0 stray`; `complete-value: 64 matched, 0 drifted, 0 stray`
- **Merge with main `af10b0b` (#161):** merges cleanly. The two sides touch no common file. On the merged tree, `model_population`, `model_reference_queries` and `composed_state_evaluation` all pass.
- **Closing keywords:** none. The body ends with `Part of #120`, which is correct. #147 stays open.

### Revert probes (re-run by the reviewer)
- `contains_pre_eligible_read` guard disabled: 3 tests red (bare parameter, integer literal, let-capture drift). Confirmed.
- `field_ordered` forced to `true`: `enforce_frame_admits_an_unordered_field_reorder_without_a_write` red. Confirmed.
- Overlap check disabled: **stays green**. The author disclosed this; see finding 6.

### Status of the earlier findings
| # | Status |
| --- | --- |
| 1 operand eligibility | Mostly fixed. `Call` is a hole (new finding 1), and a let-alias is not refused (new finding 4) |
| 2 clause context and call leak | Fixed. `NodeKind::Call` saves the anchor, resets it to `Post` and restores it, which matches shared-grammar.md:500-503. The cause is misassigned (new finding 2) |
| 3 row claims | Not fixed (new finding 5) |
| 4 delta | Fixed. Reuse is discussed under "Delta reuse" below |
| 5 type-change keying | Fixed. Identity is keyed by object, and a type change refuses `frame_violation` |
| 6 unordered reorder | Fixed. Multiset comparison for `ordered: false` |
| 7 redefinition | Partly fixed. One hop only (new finding 3) |
| 8 tests and traces | Tests added. Traces still off (new finding 7) |
| 9 docs and idioms | Fixed |

### Findings, most severe first

**1. High: `pre(F(p))` is accepted and silently returns the post value.** `src/value/expression/check.rs` (`contains_pre_eligible_read`, the `Expression::Call` arm); the test is at `tests/model_reference_queries.rs:1705`
- Why: `Call` counts as an eligible read regardless of its arguments. The callee then correctly runs at `Post`, so a `pre(...)` whose only content is a call over bare names does nothing. This is the same silent drift finding 1 of the first review was about, routed through a call.
- Failure: `pre_of_a_function_call_resets_the_callee_to_the_post_anchor` asserts that `pre(F(p))` completes with `2`, the post size. The test pins the wrong answer.
- Spec: FR-042-AC-1 legalises `pre(P(self.version, delta))` because an argument holds an eligible read. With bare-name arguments only, it is a `pre` over bare parameters, which FR-042's Behavior clause refuses.
- Fix:
  - drop `Expression::Call` from the direct `matches!`; the `children()` recursion already inspects the call's arguments;
  - `pre(F(p))` then refuses `forbidden-pre-read`, while `pre(F(allInstances(p)))` stays legal;
  - change the test to assert the refusal, and add a positive test for the argument-read form.
- The anchor reset on `Call` is still correct and should stay.

**2. Medium: `pre` outside a postcondition gets the wrong refusal cause.** `check.rs` `Expression::Pre` arm (`WrongAnchor`), `refusal.rs` `WrongSnapshotCause`
- Both causes are catalogued, not minted: FR-272 and native-diagnostics.md:88 list `wrong-anchor` and `forbidden-pre-read`. The mapping is the problem.
- QSpec's FR-208 applies FR-042 and names the cause explicitly: "`pre` in it is `wrong_snapshot`/`forbidden-pre-read` (FR-042)" for an invariant, and "`pre` in [a precondition] is `wrong_snapshot`/`forbidden-pre-read`".
- `wrong-anchor` is the "required vs supplied anchor selection" cause. It fits a different case, which the PR routes to the wrong code: a postcondition evaluated over a population with no attached pre anchor. That case currently returns `Refusal::CheckedInvariant`, which is `runtime_invariant` (`evaluate.rs` `select_anchor`). The new doc there admits this is "a real, reachable" case driven by caller input, and `runtime_invariant` is reserved for broken evaluator invariants.
- Fix:
  - `pre` outside a postcondition refuses `forbidden-pre-read`;
  - a missing pre anchor at evaluation refuses `wrong_snapshot`/`wrong-anchor`;
  - update `pre_refuses_outside_a_postcondition_context` and `pre_of_a_binding_with_no_pre_anchor_is_a_checked_invariant_refusal`.

**3. Medium: one redefinition hop misses chains.** `src/model/population.rs` `field_write_covered`
- Chains are legal. model-complete.md:56 says the redefining feature replaces "the one inherited redefined feature", so `C.x` must redefine the inherited `B.x`, not `A.x`, and there is no direct `C.x→A.x` record. FR-151 says a write is covered when it "reaches one through redefinition records".
- Failure (local probe): bundle `A ← B ← C`, with `B.x` redefines `A.x` and `C.x` redefines `B.x`, admits cleanly. A write to `c1.x` under `fieldWrites: [model.A.x]` is refused `frame_violation`.
- The doc's claim that one hop suffices because the records are "already FR-150-normalized/closed" is not what `redefinitionClosure: closed` says (model-complete.md:64). That header means every redefinition is listed, not that the records are transitively closed.
- Fix: walk `redefining → redefined` until a grant matches or no record continues the chain. Bound the walk by the record count, and refuse a cycle. Add the two-hop test.
- Ticket suggestion, outside this PR: `src/model/conformance.rs:583` has the same one-hop check for `effect-escape`, predating this PR.

**4. Medium: a let-alias of a population is re-anchored.**
- Failure (local probe): `let q = p in pre(size(allInstances<M::A>(q)))` completes with `3`, the pre population.
- FR-042-AC-3 refuses the direct analogue, `let s = self in pre(s.version)`. A state root captured outside `pre` must not be re-anchored inside it.
- Fix: when an eligible read under `pre` takes its population operand from a `let` local bound outside that `pre`, refuse `forbidden-pre-read`. A parameter operand stays legal, like `pre(self.version)`. The typer already tracks locals and their depth. Add a test.

**5. Medium: the PR body still presents V1-EXPR-020 as backed.**
- The row reads "postconditions against exact pre/post anchors **and result**". Its central requirement, QSpec FR-012, requires `result` in its Behavior clause ("result is post-only") and in FR-012-AC-3. The body's "Backed" section lists "V1-EXPR-020/EXPR-021", and the `result` gap appears only in the scope notes.
- Implementing `result` is not required in this PR. There is no operation declaration or result-type contract in the complete-V1 expression layer (`syntax.rs:81`), and the native layer's `ExprKind::ResultValue` (`src/checking/constraints.rs:225`) is a separate pipeline. It stays with row V1-EXPR-020 under #120.
- Fix: relabel the body, for example "V1-EXPR-020: **partial** — pre/post population anchors only; `result`/`self` not implemented (FR-012-AC-3), remaining under #120". Also say "V1-EXPR-021: partial" until finding 4 lands.

**6. Low: the overlap check is not pinned by any test.** `check_declared_delta` and `invocation_refuses_a_declared_delta_that_declares_the_same_identity_created_and_deleted`
- The final equality check catches every overlap anyway, since the computed sets are disjoint. So the only observable difference is the detail string.
- Fix: assert `refusal.detail.contains("both created and deleted")` so disabling the branch goes red. Otherwise delete the branch.

**7. Low: trace tags.**
- Five frame tests trace to bare `FR-151`: unordered and ordered reorder, redefinition, type change, and subtype grant. FR-151 is the link-time `effect-escape` inclusion rule. Runtime frame enforcement is FR-046-AC-3.
- Every frame and delta test also carries `FR-153-AC-7`, which is only about `pre` reads.
- `pre_refuses_outside_a_postcondition_context` and the bare-parameter and literal tests trace to FR-042-AC-3. Their criterion is FR-042-AC-4. `nested_pre_is_idempotent` is FR-042-AC-2.
- Fix: retag each test to the criterion it actually backs.

### Delta reuse
The "structurally impossible" claim holds for calling into native code. `declared_deltas` and `inspect_frames` are `Validator` methods interleaved with `budget.visit()` charges, location-path mutation and diagnostic accumulation (`self.related`). The shared part is about 10 lines of set logic, and extracting it would mean refactoring the metered native path. So a parallel implementation that reuses the same codes and causes is an acceptable reading of "reuse where possible", as long as finding 6 is resolved.

### Upstream spec conflict (not a #168 fix; ticket it in quire-specification)
QSpec FR-012-AC-7 (V1-EXPR-020's own central requirement) says "Applying pre to an immutable parameter in a postcondition preserves that parameter's value" (`expectedVersion = pre(expectedVersion)` → `true`). FR-042's Behavior clause and FR-042-AC-4 refuse `pre(delta)` on a bare parameter. #168 follows FR-042, the vendored rule, and that is defensible, but the two requirements contradict each other.

All findings are fixed inside #168.

kreneskyp:
## Delta re-review verdict: FIX (head `5cf83a231d0924dbd56470673ad07561dcd4e328`)

Covers the delta `7953455..5cf83a2`. The branch was rebased onto main `af10b0b`, and `git range-diff` shows the first two commits unchanged from `dac1152`/`95cccce`. Five of the seven round-2 findings are fixed and pinned by tests. What remains: one alias shape still drifts, and the V1-EXPR-021 row is claimed as backed when it is only partly backed.

### Gates (fresh clone at 5cf83a2)
- `cargo fmt --check`: clean
- `cargo test --locked --workspace --all-features`: 108 result blocks, 0 failed
- `make revendor-check`: `native-v1: 69 matched, 0 drifted, 0 stray`; `complete-value: 64 matched, 0 drifted, 0 stray`
- The branch sits directly on main `af10b0b`, so it merges trivially.
- No closing keywords. `Part of #120`, `refs #170`, and #147 stays open.

### Revert probes (re-run by the reviewer; every one went red)
- `contains_captured_pre_alias` guard disabled: `pre_refuses_a_let_bound_population_alias_capture_drift` red. This check is new code in 5cf83a2 and was not there before; at 95cccce this shape completed with `3`.
- `Expression::Call` re-added to the eligible forms: `pre_of_a_function_call_over_a_bare_parameter_argument_refuses_forbidden_pre_read` red.
- `redefinition_reaches` bound cut so it can no longer follow the chain: both redefinition tests red, including the new two-hop `C.x → B.x → A.x` test.
- Overlap branch disabled: `invocation_refuses_a_declared_delta_that_declares_the_same_identity_created_and_deleted` red. It is now pinned by the detail text.

### Round-2 findings
| # | Status |
| --- | --- |
| 1 `pre(F(p))` | Fixed. It refuses `forbidden-pre-read`, and `pre(F(allInstances(p)))` stays legal (FR-042-AC-1) |
| 2 cause swap | Fixed. Wrong clause refuses `forbidden-pre-read`. A missing pre anchor refuses `Refusal::WrongSnapshot(WrongAnchor)`, and the test asserts the exact variant. The remaining `CheckedInvariant` sites in this diff (the unreachable `NodeKind::Pre` arm in `Machine::apply`) cannot be reached from caller input |
| 3 redefinition chain | Fixed. The walk is bounded by `records.len()` and returns `false` when the bound is exceeded, so a cycle cannot hang it |
| 4 let-alias | Mostly fixed. One remaining hole, finding 1 below |
| 5 PR body, EXPR-020 | Fixed. Marked partial, `refs #170` |
| 6 overlap assertion | Fixed |
| 7 trace retags | Fixed. Frame and delta tests trace to FR-046-AC-3; `pre` tests to FR-042-AC-1/2/3/4 |

### Findings

**1. Medium: an alias of an alias escapes the let-alias rule.** `src/value/expression/check.rs`, `contains_captured_pre_alias`
- Failure (local probe, postcondition context): `let q = p in pre(let r = q in size(allInstances<M::A>(r)))` completes with `3`, the **pre** population. This is the same re-anchoring of a post capture that `pre_refuses_a_let_bound_population_alias_capture_drift` refuses, with one extra rename.
- Why: an inner `let r = q` pushes `r` onto `shadowed` without checking whether its initializer is itself a captured alias.
- Fix: in the `Expression::Let` arm of the walk, when `value` is a bare `Name` that is a captured alias (not shadowed and `captured_before`), treat `name` as a captured alias in `body` rather than as shadowed. Keep a separate `tainted` list, or pass the flag down. Add the test. `pre(let q = p in size(allInstances(q)))` must stay legal; the probe confirms it completes with `3` today, and FR-042-AC-2 allows it.

**2. Medium: the PR body claims V1-EXPR-021 as backed; it is partial.**
- The row reads "Evaluate immutable captures and aliases without anchor drift". Its central requirement, QSpec FR-012, has AC-4: "A captured post reference retains its observation inside pre". FR-042-AC-3's own shapes are `let v = self.version in pre(v)` and `let s = self in pre(s.version)`.
- Backed here: refusal of a re-tagged capture and of a direct population alias.
- Not backed:
  - the positive retention half, FR-012-AC-4. The behaviour is right: a local probe shows `let v = size(allInstances(p)) in pre(size(allInstances(p)) != v)` completes `true`, with `v` kept at post. But no test pins it;
  - every `self`-rooted alias, which needs `self` (#170);
  - the alias-of-alias hole (finding 1).
- Fix: mark EXPR-021 **partial** in the body, stating "population-rooted aliases only; `self`-rooted aliases wait on #170". Add a FR-012-AC-4 retention test.

**3. Low: `redefinition_reaches` is not yet reusable as-is for #171.** `src/model/population.rs:1032`
- It is private and takes `&Bundle`. `conformance.rs:583` works from `index.redefinitions` and compares by `.identity`, where this module uses `ProducerKey ==`.
- Reuse therefore needs a mechanical move to `pub(super)` taking `&[RedefinitionRecord]`, and #171 has to settle which equality it uses. Behaviour here would not change.
- Also, each hop rescans every record, so the walk is O(R²) per changed field. This does not block the PR.

All findings are fixed inside #168.

kreneskyp:
## Delta re-review verdict: FIX (head `924b7d7f00064b81cbdae88b0170b0529341b8ad`)

This review covers `5cf83a2..924b7d7`, a fast-forward that changes `check.rs` and `tests/model_reference_queries.rs`. Both round-3 fixes hold for the shapes they target. The capture rule still misses one shape: the captured alias reaches `allInstances`/`lookup` through an `if` or `let` *expression*, not a bare name.

### Gates (fresh clone at 924b7d7)
- `cargo fmt --check`: clean
- `cargo test --locked --workspace --all-features`: 108 result blocks, 0 failed
- `make revendor-check`: `native-v1: 69 matched, 0 drifted, 0 stray`; `complete-value: 64 matched, 0 drifted, 0 stray`
- No closing keywords (`closingIssuesReferences` is empty). The body has `Part of #120` and `refs #170`.

### Round-3 findings
| # | Status |
| --- | --- |
| 1 alias of an alias | Fixed for **bare-name** chains. Revert probe (always push `false` in the `Let` arm): `pre_refuses_a_let_bound_alias_of_a_let_bound_population_alias` red |
| 2 EXPR-021 row claim + FR-012-AC-4 retention | Fixed. The body now marks EXPR-021 partial ("population-rooted aliases only; `self`-rooted wait on #170"). The retention test does tell pre from post (pre = 3, post = 2). If `v` were re-anchored, the check becomes `3 != 3` and the test goes red |

### Reviewer probes (postcondition context, `p: Population(3)`, pre = 3 members, post = 2)
| Shape | Result |
| --- | --- |
| `let q = p in let s = q in pre(let r = s in let u = r in size(allInstances(u)))` | refuses `forbidden-pre-read` ✓ |
| `let q = p in pre(F2(allInstances(q)))` (alias in a call argument) | refuses ✓ |
| `let q = p in pre(F(q))` | refuses ✓ |
| `let q = p in pre(let r = p in size(allInstances(r)))` (non-alias rebinding) | completes `3`, legal ✓ |
| `let q = p in pre(let r = q in size(allInstances(p)))` (alias bound but unused) | completes `3`, legal ✓ |
| Same-name shadowing (`let q = p in pre(let q = p in …)`) | the language refuses `AmbiguousName`, so shadowing cannot be expressed |
| **`let q = p in pre(size(allInstances(let s = q in s)))`** | **completes `3`** ✗ |
| **`let q = p in pre(size(allInstances(if true then q else q)))`** | **completes `3`** ✗ |
| **`let q = p in pre(let r = if true then q else q in size(allInstances(r)))`** | **completes `3`** ✗ |
| **`let q = p in pre(let r = (let s = q in s) in size(allInstances(r)))`** | **completes `3`** ✗ |

### Findings

**1. Medium: a captured population alias still escapes through an `if` or `let` expression.** `src/value/expression/check.rs`, `contains_captured_pre_alias` (`is_captured_alias` closure, and the `bindings.push` in the `Let` arm)
- Failure: the four ✗ rows above re-anchor the post capture `q` to the **pre** population, the same drift `pre_refuses_a_let_bound_population_alias_capture_drift` refuses.
- Cause: `is_captured_alias` only recognises a bare `Expression::Name`. Any population-valued expression that *evaluates to* the alias is treated as a fresh value: `If` branches, and a `Let` whose body is the alias. This applies both as the direct `allInstances`/`lookup` operand and as a `let` initializer.
- Fix: replace the closure with a recursive `resolves_to_captured_alias(expr, bindings)`:
  - `Name` looks up `bindings`, then falls back to `captured_before`;
  - `Let { name, value, body }` recurses into `body` with `(name, resolves(value))` pushed;
  - `If` returns `then || otherwise`;
  - anything else returns `false`.
  Use it for both the direct operand and the `bindings.push`. Add tests for at least the `if` operand and the `let`-expression operand shapes.
- Alternative: `check.rs:1957`'s own doc says FR-153 gives a population value "Outputs only as the direct operand of `allInstances`/`lookup`". Refusing any population-typed `If`/`Let` *result* would also close this. Either fix is fine; the recursive resolver is the smaller diff.
- Once fixed, the body's "direct (or transitively chained)" claim for EXPR-021 becomes accurate. Today it holds only for bare-name chains.

**2. Low: trace tag on the retention test.** `tests/model_reference_queries.rs`, `pre_of_a_captured_post_reference_retains_its_post_observation` is traced `FR-042-AC-3`. That AC is the *refusal* criterion, but this test pins the positive retention half. Tag it with the criterion it backs: FR-042-AC-2, or FR-012-AC-4 if the repo traces QSpec central IDs. Without that, it inflates AC-3's backing.

All findings are fixed inside #168.

kreneskyp:
## Delta re-review verdict: FIX (head `f24db4948b0cc417d0a259331aa1f57e2586b61a`), one Low finding

This review covers `924b7d7..f24db49`, a fast-forward: merge `e225473` of main `bd29e7b` (#163), plus round-4 fix `f24db49`. Everything I was asked to check holds. What is left is one untested refusal branch, a single test to add.

### Gates (fresh clone at f24db49)
- `make ci` (fmt, clippy `-D warnings` default and all-features, workspace tests default and all-features, from-clean no-default build, fixture-audit self-test, parse example): exit 0
- `make revendor-check`: `native-v1: 69 matched, 0 drifted, 0 stray`; `complete-value: 64 matched, 0 drifted, 0 stray`
- No closing keywords (`closingIssuesReferences` is empty). The body has `Part of #120` and `refs #170`.

### Merge `e225473` (#163 typed `ModelRefusalCause`)
- **Conflict resolution:** both sides are kept. The branch's "Invocation admission" module doc sits above main's two `#![allow]`s. Every main-side site in `population.rs` (`foreign-model-selection`, `incomplete-scope`, `unclosed-subtypes`, `foreign-type`, `conflicting-identity`, `duplicate-member`, `subsetting-violation`, `operator-ineligible`, `type-mismatch`, `above-maximum`, `foreign-universe`, `absent-key`) carries main's typed variant unchanged.
- **Tags:** `unauthorized-change` and `delta-disagreement` are both catalogued causes, for `frame_violation` and `population_delta_mismatch` respectively. Source: `native-diagnostics.md` (native-v1:89, complete-value:94). No new tag strings.
- **Style:** there are seven variants, one per condition, matching the enum's convention. They carry typed data: `ProducerKey` for types and fields, `object: String` like the existing `ConflictingIdentity`, and `BTreeSet<String>` for the delta sets. `as_str` groups each family onto its catalogued tag.
- **Spelling test:** `as_str_covers_every_variant_with_its_original_tag` lists all seven, and its `expected_tag` match is exhaustive, so a new variant cannot compile untested.
- **Tests assert full causes:** yes, `assert_eq!` on the typed variant, for six of the seven. `DuplicateDeclaredIdentity` is not reached by any test; see finding 1.

### Round-4 fix `f24db49`
- **Spec reading:** agreed. Neither vendored FR-042 nor FR-153 limits the alias-capture rule to bare identifiers. `bind_parameters`' "direct operand" doc governs where the `Population` *type* may appear, not this syntactic check. The recursive resolver is the right route.
- **Can anything else produce a population value?** No. The grammar has no `match`. Function results (`mod.rs:218`), `Option` payloads, record/tuple/object members and collection elements all go through `check_declared_type`, which refuses `Population`. So a population-typed expression can only be a `Name`, `Let`, `If` or `pre`, and the resolver covers the first three. `pre` over a bare name or alias already refuses as having no eligible read.
- **Revert probe:** with the resolver's `If` and `Let` arms set to `false`, all four new tests go red.
- **Retag:** the retention test is now `FR-042-AC-2` ✓

### Reviewer probes beyond the four tested shapes (postcondition, `p: Population(3)`)
| Shape | Result |
| --- | --- |
| `let q = p in pre(size(allInstances(let a = (if true then (let b = q in b) else p) in if true then p else a)))` | refuses ✓ |
| `let q = p in pre(size(allInstances(if true then p else q)))` (alias in the else branch only) | refuses ✓ |
| `let q = p in pre(size(allInstances(pre(if true then q else q))))` | refuses ✓ |
| `let q = p in pre(pre(size(allInstances(let s = q in s))))` | refuses ✓ |
| `let q = p in let s = q in pre(let r = s in let u = r in size(allInstances(u)))` | refuses ✓ |
| `let q = p in pre(F2(allInstances(q)))` / `pre(F(q))` | refuse ✓ |
| `let q = p in pre(size(allInstances(let t = q in p)))`: no false positive | completes `3` ✓ |
| `let q = p in pre(size(allInstances(if true then p else p)))`: no false positive | completes `3` ✓ |

### Finding

**1. Low: the `DuplicateDeclaredIdentity` refusal branch has no test.** `src/model/population.rs:1290-1298` (`check_declared_delta`)
- V1-EXPR-022 claims FR-046 delta agreement as backed. Of `check_declared_delta`'s three conditions, overlap and mismatch are pinned by full-cause asserts. Nothing reaches the duplicate branch.
- Failure: delete the `if !identities.insert(...)` refusal and every test stays green. A delta declaring `["o1", "o1"]` as created then collapses into the set and is admitted when `o1` really was created. That breaks the "declared-identity duplicate" shape the PR body's own FR-046 duplication ruling says it shares with `frames.rs`.
- Fix: add `invocation_refuses_a_declared_delta_that_declares_the_same_identity_twice` to `tests/model_population.rs`. Assert `Code::PopulationDeltaMismatch` and `ModelRefusalCause::DuplicateDeclaredIdentity { identity: "o1".into() }`, traced like its siblings, with a revert-probe note.

All findings are fixed inside #168. Once this test lands, the head is MERGE from my side.

PR #172, opened 2026-09-18
FR-151: dispatched-call typing, bridge, runtime and termination (TC-196 D06-D08)

kreneskyp:
## Summary

Adds the checked-layer/runtime half of FR-151 dispatched `receiver.member(args)` calls that `tests/model_dispatch.rs`'s D01-D05 (link-time linking) explicitly leave uncovered: TC-196 D06 (`dispatch.select` runtime evaluation with effective-precondition decision), D07 (the `ClauseKind` restriction admitting a dispatch call only inside an invariant/precondition/postcondition), and D08 (an FR-146 call-graph cycle through a dispatch edge refused `definition-cycle` before measure-decrease logic runs).

Five commits, each building on the last:

1. **IR node** — `Expression::Dispatch`/`NodeKind::Dispatch` and the checked-layer `DispatchTable`/`DispatchCandidate`/`DispatchOperation` types. The checker types a dispatch call unconditionally for now; the evaluator refuses one as a checker invariant (both completed later in the series).
2. **Clause-kind + D07** — `ClauseKind` (Invariant/Precondition/Postcondition/Body) as the context every `FunctionDeclaration` body and dispatch call site checks under; the D07 restriction guard.
3. **Bridge** — `crate::model::checked_dispatch`: a pure model↔checker bridge that types every linked dispatch candidate's effective precondition and body against the value checker, using `link_dispatch`'s own linked table, and returns a checked `PackageDeclarations` ready for the evaluator. `OperationClauses` is the caller-supplied side table of each candidate's clause `Expression`/signature, since `Bundle` carries no `Expression` payload. The new end-to-end bridge test does not yet pass at this commit (evaluator runtime lands next).
4. **`dispatch.select` runtime + D06** — the evaluator resolves a dispatched call's receiver against its `DispatchTable`, charges `dispatch.select` sized by the table's distinct-candidate count, and either runs the linked candidate body directly or evaluates its effective precondition first (a new `DispatchGuard` task) — `true` continues to the body, `false` undefines the call as `Undefined::PreconditionFalse`. `ChargePoint::DispatchSelect` backs the charge; `tests/exact_decimals.rs`'s golden charge-point table now carves `dispatch.select` out of the otherwise-still-deferred `dispatch` family. The bridge test now passes end to end.
5. **Termination + D08** — extends the FR-146 call-graph with dispatch edges (`CallSite::kind`: `EdgeKind::Ordinary`/`Dispatch`); a strongly connected component containing a dispatch edge is refused `CheckCause::DefinitionCycle` (`invalid_package`/`definition-cycle`) ahead of, and instead of, the ordinary measure-decrease obligation. An all-ordinary component keeps exactly its existing refusals, unaffected (verified via revert-probe).

## Scope / limitations documented in the bridge test

The bridge's own end-to-end test (`bridge_links_a_real_family_and_evaluates_through_the_built_table`) links a real two-candidate family and evaluates `self.size()` for one receiver type only — not also for the redefining subtype through the same checked call site. `ValueType::admits()` requires a `Value::Reference`'s exact `object_type()` to equal a checked parameter's declared `Reference<T>`, with no subtype allowance (the same "`TypeEnvironment` island" boundary `tests/model_reference_queries.rs` documents for `allInstances`/`lookup`). Resolving a dispatch call's receiver against its instance's most-specific type end to end needs that same identity bridge or an equivalent — FR-152/FR-153/#131 territory, not this bridge's.

## Test plan

- [x] `cargo fmt --check`
- [x] `cargo clippy --locked --workspace --all-targets --all-features -- -D warnings`
- [x] `cargo test --locked --workspace --all-features` (full workspace, 0 failures)
- [x] `make revendor-check`
- [x] All 3 new guarding tests (D06/D07/D08) revert-probed: disabling each guard makes its own test fail with the expected symptom, then restored
- [x] 5-commit reconstruction verified byte-identical (aside from formatting) to the originally fully-tested working tree

Part of #120


kreneskyp:
## Review of #172 at `f87e5eefc0aaaf62af09d017cb9002b20976307e`: **FIX**

This review covers code review, `/rust-review` and gap analysis against FR-151, TC-196 D06–D08 and V1-TYPE-028, scoped to this diff.

### Gates (fresh worktree, separate `CARGO_TARGET_DIR`, debug info off)
| Gate | Result |
| --- | --- |
| `cargo fmt --check` | pass |
| `cargo clippy --locked --workspace --all-targets --all-features -- -D warnings` | pass |
| `cargo test --locked --workspace --all-features` | pass |
| `make revendor-check` | pass (native-v1 69/0/0, complete-value 64/0/0) |
| Merge onto current `origin/main` (af10b0b) | clean |
| Closing keywords in commit messages | none. The messages say "part of #120". |
| Revert probes (disable each guard, run `tests/dispatch_calls.rs`) | D06 `if !holds` → `d06_..._false_is_undefined` red. D07 clause-kind guard → `d07_...` red. D08 `dispatch_cycle_refusal` → `d08_...` red. |
| `dispatch.select` charge vs `value-accounting.md:132` | matches (`value_occurrences=c`, `work_units += c` instead of 1, charged after receiver/args, before `function.call`) |
| Cause strings | `definition-cycle` and `precondition-false` are both already catalogued. No new cause string is minted. |

The guards themselves work. The findings below come from bridge-level probes: throwaway integration tests run against this head and then deleted.

### Findings (most severe first; all fix inside this PR)

**1. High: the effective precondition drops absent clauses, which gets TC-196 D06's second half wrong.** `src/model/checked_dispatch.rs:278-310` (the `if let Some(expression) = clauses.own_precondition.get(&current)` at :289)
FR-151 says "An absent precondition is `true`", and the effective precondition is own ∨ redefined. TC-196 D06 ("then without `PB`") expects `true` through `B.size`, "whose effective precondition is its absent (true) clause disjoined with `PA`". The bridge skips absent clauses, so `B.size`'s effective precondition becomes `PA` alone.
Probe: the bridge family A.size/B.size, `PA = false` on `A.size`, no precondition on `B.size`, static receiver `B`, evaluate `self.size()` for `b1`. It returns **`Undefined(PreconditionFalse)`**; the spec requires `Completed(2)`. The doc comment at :272-276 cites D08 for the wrong rule. A member with no own precondition makes the whole effective precondition `true`. Only the call-graph edges still reach the ancestors' clauses (finding 3).

**2. High: D06 / FR-151-AC-7 / AC-8 are not delivered by the tests. The D06 tests cannot fail on finding 1.** `tests/dispatch_calls.rs:203-291`
Both `d06_*` tests use a hand-built single-candidate table with a literal `true`/`false` precondition. Nothing runs the TC-196 D06 fixture: two candidates, `PA`/`PB` both false giving `precondition-false` "without selecting `A.size`", then without `PB` giving `true` through `B.size`. So AC-7's "never selects a less specific method" and the absent-is-true rule are untested.
The charge assertion `meter.consumed(WorkUnits) >= 2` (:248) is too weak to fail. The false case never asserts "no `function.call`".
This can be done today. Build it through `checked_dispatch_operation` with `DispatchRoot.receiver_type = B` and `self: Reference<B>`; the probe above does exactly this. Then assert the exact meter: `dispatch.select` `value_occurrences` 2 and work 2; for the false run, zero `function.call`.

**3. High: D08's call graph uses the wrong nodes. The D08 test asserts only the variant.** `src/model/checked_dispatch.rs:202-214`, `src/value/expression/termination.rs:141-175`, `tests/dispatch_calls.rs:311-317`
FR-151 draws dispatch edges to "every precondition clause of every candidate's effective precondition". TC-196 D08 expects the SCC `{PS}` listing exactly `PS -> PS`. The bridge instead copies every ancestor clause into a synthesized `<candidate>.precondition` function per candidate.
Probe: D08 through the bridge (`PS = self.size() >= 0` on `A.size`). It refuses with component `{model.A.size.precondition, model.B.size.precondition}` and **four** edges: A→A, A→B, B→A, B→B. The spec wants one edge, `PS -> PS`, and the listing should follow "source declaration order", not bridge-assigned package indices.
The D08 test is `matches!(.., DefinitionCycle { .. })` with no `edges` or location assertion, so it passes either way.
Fix: make one checked function per authored precondition clause, shared by every candidate that inherits it. Build the effective precondition from those. Add a bridge-driven D08 test that asserts `edges == [("PS","PS")]` and the location.

**4. Medium: dispatch table and candidate indices are never validated. An out-of-range table silently disables D08.** `src/value/expression/facts.rs:893-897` (`.unwrap_or_default()`), `check.rs:1345-1393`, `evaluate.rs` `NodeKind::Dispatch` arm
`PackageDeclarations`, `DispatchOperation` and `DispatchTable::new` are `pub`. Probe: `one_candidate_package` with a self-dispatching precondition and `dispatch_operations[0].table = 7`. `check()` returns **Ok**: no edges, so no `definition-cycle`. Evaluation then fails only as a runtime invariant.
There are more unchecked cases. Nothing checks `DispatchCandidate::{body, precondition}` against `functions.len()`. Nothing checks the candidate function's arity or parameter types against the `DispatchOperation` signature, yet the evaluator binds `[receiver, ..args]` into whatever frame the callable has.
Fix: validate all of these in `PackageDeclarations::check` and refuse there. A missing table should never mean "no edges".

**5. Medium: the effective-precondition walk is order-dependent, truncates silently and ignores parameter renames.** `src/model/checked_dispatch.rs:283-304`
- It keeps only the first matching redefinition record (`.iter().find(..)`, :296-299). An operation that redefines several members (the FR-151 wording is "the members it redefines", plural) loses the others. The result also depends on record order, which AC-5 forbids.
- `MAX_ANCESTOR_DEPTH` hits `break` (:292-294). The effective precondition is silently truncated instead of refused.
- Ancestor clause `Expression`s are checked under the candidate's own parameter names and receiver type. Redefinition correspondence is positional, so an `A.size(x)` clause that names `x` fails to resolve inside `B.size(y)`. An ancestor clause that dispatches on `self` resolves only if the redefining candidate declares `self: Reference<ancestor>`: `dispatch_operations` matches `receiver_type` exactly.

**6. Medium: D07 can be bypassed through named calls.** `src/value/expression/syntax.rs:370`, `src/value/expression/mod.rs:223,240`, `check.rs` `fn call` (≈1264)
The candidate bodies and synthesized preconditions are ordinary named package functions. `Typer::call` resolves any function by name, so a `ClauseKind::Body` function can call `model.A.size.precondition(r)` and run a dispatched call from a body. It can also call `model.A.size(r)` statically and skip dispatch altogether. `clause_kind` is also a `pub` field any caller can set.
Separately, the `decreases` measure is checked under `function.clause_kind` (mod.rs:240). A `Precondition`-kind declaration's measure would admit a dispatch call, contradicting the `ClauseKind::Body` doc ("…or a `decreases` measure").

**7. Medium: the trace tags point at the wrong ACs.** `tests/dispatch_calls.rs:158,207,254,299,435`
- D06 is tagged `FR-151-AC-2` (inheritance cycle/variance). It should be AC-7, plus AC-8 for the charges.
- D08 is tagged `FR-151-AC-4` (variance independence). It should be AC-3, which is where "a call-graph cycle through a dispatch edge refuses `definition-cycle`" lives.
- D07 is tagged AC-2, but no AC covers the clause restriction; tag it `TC-196` only.
- The bridge test claims AC-1/AC-2, neither of which it exercises.

As tagged, the matrix would count AC-2/AC-4 as backed by unrelated tests and leave AC-3's cycle clause and AC-7 unbacked.

**8. Low: `Undefined::PreconditionFalse` is a unit variant.** `src/value/outcome.rs:79`
`native-diagnostics.md:123` says `precondition-false` carries "the called effective operation, the selected method's effective identity, the receiver reference and the call locus". The variant carries only the locus (through `Evaluation`).

**9. Low: `/rust-review` idioms.**
- `DispatchBridgeRefusal::MissingClauseData { field: &'static str }` puts the discriminant in prose, and `missing(subtype, "object_keys")` stores a type key in a field named `operation` (`checked_dispatch.rs:108-121,231-234`). Each missing input should be its own variant.
- `distinct.len() as u64` (:241) should use `u64::try_from`. This is style only; it widens on every supported target.
- The D07 test only goes through `check_clause_expression(.., Body, ..)`. It never uses TC-196's own shape, a package `function f` whose body contains `r.size()`.

**10. Low: argument typing admits more than reference upcasts.** `check.rs` `dispatch_call` → `check_as` → `coerce`
This admits `Integer → Int[..]` coercion, but FR-151 says "reference upcasts only". No test passes a dispatch argument.

### Out of scope: ticket suggestions
- The FR-151 query-only restriction (result present, empty effect set) is not enforced; the bridge ignores `OperationMemberRecord.effect` and `result`.
- There is no inherited exposure through ancestor static types, and subtype references are not admitted at `Reference<T>` parameters. The PR already documents both (#131 / FR-152 / FR-153).
- No test backs TC-196 D01's runtime half: evaluating `I` for `d1`/`c1` with exact work units, and `incomplete … charge_point: dispatch.select` at `work_units: 1`.

### Rebase notes (predictable, non-blocking)
- **#168**: textual conflicts in `src/value/expression/{check,evaluate,ir,mod,syntax}.rs` and `src/value/mod.rs`. There is also a semantic overlap: #168 adds `Typer.postcondition: bool` and `CheckedPackage::check_postcondition_expression`, which duplicate this PR's `ClauseKind` and `check_clause_expression`. Unify on `ClauseKind` when rebasing: `postcondition` becomes `clause_kind == ClauseKind::Postcondition`, and drop the second entry point. Both PRs add a `Task` variant (`RestoreAnchor` / `DispatchGuard`). Check that `DispatchGuard`'s frame pop and `RestoreAnchor` nest correctly when a dispatched call sits inside a `pre(...)` anchor.
- **#163**: merges cleanly on top of main + #172.
- **#167**: touches no files in this PR.

kreneskyp:
## Re-review of #172: delta `f87e5ee` → `7569d3d1914e98fedb48648dcbb875ecc39f208a`: **FIX**

The delta contains three commits: `9aae22f` (merge of main bd29e7b, #163), `4260bcb` (the fixes) and `7569d3d` (merge of #168).

### Gates at `7569d3d` (fresh worktree, separate `CARGO_TARGET_DIR`)
| Gate | Result |
| --- | --- |
| `cargo fmt --check` | pass |
| `cargo clippy --locked --workspace --all-targets --all-features -D warnings` | pass |
| `cargo test --locked --workspace --all-features` | pass (1249 passed, 0 failed) |
| `make revendor-check` | pass (69/0/0 native-v1, 64/0/0 complete-value) |
| `make ci` | pass |
| `origin/main` (bd59852) is an ancestor of head | yes |
| `9aae22f` | identical to a clean auto-merge of #163 |
| Closing keywords in this PR's own commits and body | none. `Closes #139`/`#141` appear only in main commits that have already landed. #147, #174 and #175 are untouched. |

### The #168 merge in `7569d3d`
- **Every `pre(...)` refusal still fires in the contexts #168 had.** The `Expression::Pre` arm matches #168 byte for byte except that `!self.postcondition` became `self.clause_kind != ClauseKind::Postcondition`.
  - Ordinary functions (`FunctionDeclaration::new`) are `Body`.
  - `check_expression` is `Body`.
  - `check_postcondition_expression` is `Postcondition`.
  - Measures are hard `Body`.
- **No #168 test was weakened.** `git diff bd59852 7569d3d -- tests/model_reference_queries.rs tests/model_population.rs` shows only the two `FunctionDeclaration { .. }` → `FunctionDeclaration::new(..)` rewrites and one comment. Every assertion is unchanged, and all of them pass.
- **`substitute_names` over `Pre` cannot open an alias escape.** It only rewrites clauses that the bridge compiles as `ClauseKind::Precondition`. There, `pre(...)` is refused `wrong_snapshot`/`forbidden-pre-read` before `contains_pre_eligible_read` or `contains_captured_pre_alias` ever run. See L4 for the one context that did widen.

### Prior findings: status
| # | Status |
| --- | --- |
| 1 absent precondition = true | **Closed.** `effective_terms` absorbs correctly. A probe with PA=true and PB=false on a B receiver returns `2`. |
| 2 D06/AC-7/AC-8 test gap | **Partially closed.** See M1. |
| 3 D08 nodes/edges | **Code closed. Test gap remains** (M2). Probe through the bridge with `PS = self.size() >= 0` on A.size gives exactly one refusal with edges `[("model.A.size.precondition","model.A.size.precondition")]`. |
| 4 unvalidated indices | **Closed.** `validate_dispatch_function` handles it, and `facts.rs` now refuses instead of `unwrap_or_default`. Revert probe: disabling the validation loop turns 2 tests red. |
| 5 ancestor walk | **Closed** for order and truncation. The parameter-rename part has a new defect (L3). |
| 6 D07 bypass / measure kind | **Closed for `Expression::Call`.** Revert probe: removing the `callable_by_name` filter turns `synthesized_dispatch_candidate_is_not_callable_by_name` red. The measure is now always `Body`. The `CheckedPackage::call` residue is in L4. |
| 7 trace tags | **Closed.** D06 → AC-7/AC-8, D08 → AC-3, D07 and the bridge test → `TC-196` only. |
| 8 payload | **Closed.** `PreconditionFailure` exists; a naming nit is in L6. |
| 9 idioms | **Closed.** `MissingClauseField`, `MissingObjectKey` and `u64::try_from` are in. See L5 for the new stringly-typed variant. |
| 10 coercion | **Closed.** Revert probe: neutering `check_dispatch_argument`'s equality check turns its test red. |

### Findings (most severe first; fix inside this PR)

**M1. Medium: the core of AC-7 and the whole combinator path have no test.** `tests/dispatch_calls.rs:461-559`, `src/model/checked_dispatch.rs:690-715`
FR-151-AC-7 says "a false selected precondition yields `undefined` … and **never selects a less specific method**". TC-196 D06's first half is PA=false and PB=false on a `B`/`D` receiver.
Neither new D06 test covers it:
- `d06_bridge_absent_…` has PB absent, so there is no false precondition to fall back from.
- `d06_bridge_false_…` uses an **A** receiver, which has no less-specific method to fall back to.

No test ever builds a `*.precondition.effective` combinator: `terms.len() > 1`, the Or-reduce and `rename_parameters` have zero coverage.
The behaviour is correct today. My probe with PA=false and PB=false on `b1` gives `Undefined(PreconditionFailure { operation: "size", selected: "model.B.size", .. })` with work 3. PA=true and PB=false gives `Completed(2)`.
Add both as tests, asserting the payload, zero `function.call` and the exact work units.

**M2. Medium: D08 is still tested only on the hand-built one-candidate package, so the static-ancestry edge fix is unguarded.** `tests/dispatch_calls.rs:561-597`, `src/value/expression/ir.rs:207`
Revert probe: I changed `callees.extend(candidate.precondition_clauses…)` back to the old `callees.extend(candidate.precondition)`. **All 10 tests in `dispatch_calls.rs` stay green.**
Add the TC-196 D08 fixture through `checked_dispatch_operation`: `PS = self.size() >= 0` on A.size, then assert `edges == [(PS, PS)]` and the location. Add a second case with an own `PB` on B.size. That case forces the combinator. With the reverted `callees`, the SCC then wrongly contains `model.B.size.precondition.effective`; with the fix, it is `{PS}` only.

**L3. Low: `rename_parameters` is not capture-avoiding, so a valid ancestor clause makes the package refuse.** `src/model/checked_dispatch.rs:358-374, 401-405`
Probe setup:
- A.size has parameter `a`, with `PA = let b = 5 in a = a`.
- B.size has parameter `b`, with `PB = false`.

The combinator is spliced as `false or (let b = 5 in b = b)`. `check()` then refuses with `AmbiguousName { name: "b" }` at `model.B.size.precondition.effective`, even though PA checks fine in A's own scope. `shadow` guards only the *source* names. A binder equal to a *target* name captures the renamed reference.
Fix: alpha-rename a colliding binder, or build the combinator from calls to the shared authored-clause functions instead of inlining. Add a test.

**L4. Low: the #168 "no package function admits `pre(...)`" invariant now depends on a public constructor.** `src/value/expression/syntax.rs:422`, `src/value/expression/mod.rs:324, 535-540`
#168 hard-coded `postcondition: false` for every package function body. The merged code uses `function.clause_kind`, and the public `FunctionDeclaration::clause(.., ClauseKind::Postcondition)` now yields a package function where `pre(...)` checks.
Separately, `CheckedPackage::call` resolves through `fn function()`, which ignores `callable_by_name`. Synthesized dispatch bodies and clauses therefore stay directly invokable by name from the public API.
Fix: make `clause` `pub(crate)` or reject `Postcondition`, and filter `callable_by_name` in `function()`.

**L5. Low: `CheckCause::InvalidDispatchDeclaration { detail: String }` puts the discriminant in prose and carries no index path.** `src/value/expression/refusal.rs:245`, `src/value/expression/mod.rs:161-213`
Every refusal is located at `root(Origin::Expression)`. `native-diagnostics.md:82` requires `invalid_package`/`invalid-value` to "retain the actual field/index path". Split it into typed variants: out-of-range table, out-of-range function (with role and index), and arity/type/result mismatch. Each needs a real location.

**L6. Low: the payload and bound are imprecise.**
- `PreconditionFailure.operation` is the unqualified member name, taken from the first `dispatch_operations` entry whose `table` matches (`evaluate.rs:972`). native-diagnostics wants "the called effective operation". If two operations share a table, the wrong one can be reported.
- `MAX_ANCESTOR_DEPTH` counts visited nodes, not depth. It counts per closure in `ancestor_closure`, and **once for the whole bridge run** in `effective_terms` (`steps`, :680). A wide family with more than 128 precondition-bearing members would therefore be refused `DispatchFamilyDepth` although no chain is deep.

### Ticket suggestions (unchanged from the last review)
- The query-only restriction (FR-151) is not enforced.
- There is no inherited exposure through ancestor static types or subtype `Reference<T>` admission (#131 / FR-152 / FR-153).
- TC-196 D01's runtime half is still unbacked.

kreneskyp:
## Round-3 review (7569d3d → f68c0fe731596e4e83149f2d87783802c495aa15): FIX, Low findings only

**Gates on f68c0fe** (fresh worktree): `cargo fmt --check` 0 · `cargo clippy --locked --workspace --all-targets --all-features -- -D warnings` 0 · `cargo test --locked --workspace --all-features` 0 (109 suites ok) · `make revendor-check` 0.

**Merge 4730731:** identical to a clean auto-merge of main 2744e7b. `src/model/normalize.rs` has no diff from origin/main, and origin/main is an ancestor of the head. The PR body says "Part of #120", and `closingIssuesReferences` is empty, so #147, #174 and #175 stay open.

### Round-2 findings: all six fixed, each revert-probed
| Finding | Fix | Revert probe |
|---|---|---|
| M1 AC-7 combinator | `d06_bridge_own_and_ancestor_precondition_both_false_selects_b_never_a` asserts `selected: "model.B.size"` and 3 work units. `…own_false_ancestor_true_completes_through_combinator` asserts `Integer(2)` and 4 units. | both pass on the head |
| M2 D08 bridge | `d08_bridge_ancestor_cycle_survives_a_sibling_combinator` | revert `ir.rs:207` to `candidate.precondition` → **red** |
| L3 capture | `bind_name` alpha-renames binders | force `bind_name` to never rename → capture test **red** |
| L4 `call()` | `callable_by_name` filter at `mod.rs:644` | delete the filter → `checked_package_call_refuses_…` **red** |
| L4 `clause(Postcondition)` | `assert!` at `syntax.rs:443`, plus a `should_panic` test | (see L-3 below) |
| L5 typed cause | `InvalidDispatchDeclaration` enum; located at `Origin::Body` | Arity, ParameterType and ResultType location tests pass |
| L6a operation index | `NodeKind::Dispatch.operation` | revert `evaluate.rs:980` to a table search → shared-table test **red** |
| L6b `effective_terms` | per-walk `depth` | simulate a walk-wide count (`depth + memo.len()`) → wide-family test **red** |

### Remaining findings (all Low)
**L-1: 12 review-history comments added by this PR.** Code comments should describe the code, not the review that led to it. Once the PR merges, a reader cannot resolve "finding #172-5". Every one of these is new in `origin/main..f68c0fe`:
- `src/model/checked_dispatch.rs:169, :266, :307, :823, :868`
- `src/value/expression/check.rs:1655`
- `src/value/expression/mod.rs:359`
- `tests/dispatch_calls.rs:220, :254, :292, :352` (`L4(a)`), `:638`

Fix: delete the parenthetical tags, and keep the sentence around each one.

**L-2: the new per-entry depth in `ancestor_closure` is untested.** At `src/model/checked_dispatch.rs:287`, change `depth > MAX_ANCESTOR_DEPTH` back to the node count `closure.len() > MAX_ANCESTOR_DEPTH`. The lib tests, `dispatch_calls` and `model_dispatch` all stay green. The wide-family test only exercises `effective_terms`.
- Failure scenario: a candidate with more than 128 ancestors that fan out but stay shallow, e.g. multiple generalization where one member redefines many parents. The code would regress to refusing that candidate, and no test would catch it.
- Fix: add an `ancestor_closure` unit test with more than `MAX_ANCESTOR_DEPTH` direct redefined parents. It must return `Ok`, with closure length = parents + 1.

**L-3: `FunctionDeclaration::clause` panics on a public enum value (`src/value/expression/syntax.rs:443`).** It is a library constructor, and `ClauseKind::Postcondition` is a public, valid value of its parameter. The rest of this API reports malformed caller data as typed refusals. This same PR does exactly that with `InvalidDispatchDeclaration`.
- Failure scenario: a downstream crate that builds declarations from data can abort the process instead of getting a refusal.
- Fix (preferred): make the value unrepresentable, e.g. give `clause` a narrower kind enum without `Postcondition`.
- Fix (alternative): refuse it in `PackageDeclarations::check` with a typed cause.

**Nit: `src/value/expression/facts.rs:909-918`.** The comment says the member comes from the node's own index and is "never a placeholder". The code then does `.unwrap_or_default()`, which produces `""` if the index is missing. It also clones `member` on every dispatch-node walk, even when the table is valid. Fix: move the lookup into the `ok_or_else` closure, and report the missing operation index honestly.

### Ticket suggestions (out of scope)
- None new.

Issue #173, opened 2026-09-18
[QSL#173] Detect competing operation redefinitions (c >= 2) in one type


kreneskyp:
Found reviewing PR #167 (comment 5736464209). This predates that PR.

When one type redefines the same inherited operation more than once, nothing detects the competition. For example, `B` redefines `A.op` with both `B.op2` and `B.op3`, and the model is Completed. `src/model/conformance.rs:52-59` hands the competing-redefinition case to normalize's phase 4, but phase 4 in `src/model/normalize.rs` resolves only field redefinitions.

Scope:
- Detect competing operation redefinitions, applying the same dominance rule as fields (a winner if one dominates, otherwise `derivation-conflict`), using catalogued FR-272 causes only.
- Add tests for the resolved and conflicting cases.

#167 charges the conflict-check for these groups; this ticket adds the detection and resolution. Part of #120.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #177, opened 2026-09-19
QSL #171: walk the full redefinition chain in the effect-escape check

### LINKED ISSUE #171: effect-escape covers only one redefinition hop (conformance.rs:583)
Found reviewing PR #168 (comment 5735948798). This predates that PR.

`src/model/conformance.rs:583` (the `effect-escape` check) follows a single `BundleRecord::Redefinition` hop when it decides whether a field write reaches a declared grant. Redefinitions can chain. model-complete.md:56 says a redefining feature replaces "the one inherited redefined feature", so C.x→B.x→A.x is legal. FR-151 covers a write that "reaches" a grant.

Repro shape: a bundle with A←B←C, each redefining x. An operation declares `fieldWrites: [A.x]` and writes C.x. It is wrongly refused with effect-escape.

Fix: walk the redefinition chain with a bound and refuse a cycle. Share the walk with the runtime frame check that #168 adds (its `field_write_covered`).

Part of #120.

kreneskyp:
## Gap

`check_operation_redefinition`'s effect axis (`src/model/conformance.rs`) compared a redefining operation's field writes against the redefined operation's declared writes through at most one `BundleRecord::Redefinition` hop (a `direct` match plus one `via_redefinition` hop). `model-complete.md:56` says a redefining feature replaces "the *one* inherited redefined feature", so a chain such as `C.x` redefines `B.x`, `B.x` redefines `A.x` is legal and normal, with no direct `C.x -> A.x` record (`model-complete.md:64`). An operation declaring `fieldWrites: [A.x]` and a redefining operation writing `C.x` through that two-hop chain was wrongly refused with `EffectEscape`.

## Fix

Reuses `crate::model::population::redefinition_reaches` (added by #168 for the identical gap in the runtime frame check, `field_write_covered`) instead of writing a second chain walk. Per the #168 review's reuse note:

- `redefinition_reaches` now takes a `&[BundleRecord]` slice rather than a whole `&Bundle`, and is `pub(super)` so `conformance.rs` (a sibling module under `crate::model`) can call it.
- The effect axis's `direct`/`via_redefinition` two-branch check collapses into one `redefinition_reaches` call, since the walk already checks the field itself before hopping.
- **Equality change is one-sided.** Before this PR, the conformance effect axis compared `ProducerKey`s by `.identity` alone; that closure now compares the whole key (full derived `PartialEq` — `authority`, `identity`, `revision`, `digest`), matching `population.rs`'s `field_write_covered`, which already used full-key equality (via `effect.field_writes.contains(candidate)`) and is unchanged by this PR. `redefinition_reaches` itself is unchanged from #168's bounded, cycle-safe full-chain walk; only the conformance call site's equality and hop reach changed.

## Tests

- `r09_operation_redefinition_effect_axis_reaches_through_a_two_hop_field_redefinition_chain` (`tests/model_conformance.rs`): a two-hop `model.A <- model.B <- model.C` field-redefinition chain with an operation redefinition whose write only reaches the declared grant through both hops. Confirmed red without the fix, green with it.
- `r10_operation_redefinition_effect_axis_refuses_a_write_at_a_revision_the_grant_does_not_name` (`tests/model_conformance.rs`) and `enforce_frame_refuses_a_field_write_at_a_revision_the_declared_grant_does_not_name` (`tests/model_population.rs`): pin the equality choice at both call sites — a write to `A.x` at revision "2" against a grant for `A.x` at revision "1" must refuse, not admit by identity alone.
- `r11_operation_redefinition_effect_axis_refuses_a_chain_that_never_reaches_the_grant` (`tests/model_conformance.rs`): a one-hop chain (`C.x -> B.x`, no `B.x -> A.x`) that never reaches the declared grant refuses with exactly `[EffectEscape{C.x}]`.
- `r12_operation_redefinition_effect_axis_refuses_and_terminates_on_a_redefinition_cycle` (`tests/model_conformance.rs`): a redefinition cycle (`C.x -> B.x -> C.x`) terminates and refuses with the same outcome, rather than hanging.

Every new test asserts the whole outcome with `assert_eq!`. All were hand-mutation-probed against the pre-fix/pre-pin source (temporary edits, reverted before commit, never `git stash`): r09 red without the fix; r10 red under an identity-only closure at each call site; r12 hung under an unbounded `loop` (killed by an external `timeout`) instead of failing fast.

## Gates

- `cargo fmt --check` -- clean
- `cargo clippy --all-targets --all-features -- -D warnings` -- clean
- `make ci` (`ci-default-features`, `ci-all-features`, `ci-clean-build`) -- all green
- `make revendor-check` -- `native-v1: 69 matched, 0 drifted, 0 stray`; `complete-value: 64 matched, 0 drifted, 0 stray`

Closes #171
Part of #120


kreneskyp:
## Review verdict for PR #177 at `f2c6136`: FIX

This review covers code, Rust idioms and gaps. It is scoped to the diff against `main` (`2744e7b`).

The fix is correct. `redefinition_reaches` walks upward from the write, through `redefining -> redefined`, until it finds a grant. The walk is capped at `records.len() + 1` steps, so a cycle ends and returns `false`. I confirmed three cases with temporary probe tests that are not committed:

- a broken chain (`C.x -> B.x` with no `B.x -> A.x`) refuses with exactly `[EffectEscape{C.x}]`;
- a cycle (`C.x -> B.x -> C.x`) ends and refuses with exactly `[EffectEscape{C.x}]`;
- a write to `A.x` at revision 2, checked against a grant for `A.x` at revision 1, refuses with `EffectEscape`.

Full `ProducerKey` equality is the right choice. `normalize` already refuses any `fieldWrites` entry that does not fully match a declared field member (`index.field_member_keys.contains`), and `check_operation_redefinition` already looks up operations by full key. The problem is that the PR's tests do not pin two of these behaviours.

### Findings

1. **Medium. The equality change is not pinned at either call site.** Files: `src/model/conformance.rs:677`, `src/model/population.rs:1078`.
   - Before this PR, the conformance effect axis compared **identity only** (`w.identity == write.identity`). Now it compares the whole key. That is a real behaviour change: two writes that share an identity but differ in revision used to be admitted and are now refused.
   - `population.rs`'s `field_write_covered` already used full `PartialEq` before this PR, so the population frame's behaviour is unchanged. The PR body's "both call sites now use the same equality" reads as if population changed too. No `model_population` test changes meaning.
   - **Mutation (a):** I changed the conformance closure to `redefined.effect.field_writes.iter().any(|w| w.identity == candidate.identity)`. Every PR test and every existing test stays green.
   - **Mutation (b):** I made the same change in `population.rs`'s closure. `cargo test --all-features --tests` stays fully green.
   - The doc comment cites `f2_field_members_sharing_an_identity_but_differing_in_revision_both_survive` as the pin, but that test only checks `ConformanceIndex.fields`. It never runs the walk.
   - **Fix:** add one test per call site where a write to `A.x@rev2` meets a grant for `A.x@rev1` and must refuse. At conformance, assert `EffectEscape{A.x@rev2}`. At population, assert the frame violation.

2. **Medium. The new caller has no test for a chain that does not reach the grant, or for a cycle.** File: `tests/model_conformance.rs`, next to r09.
   - No test anywhere in the repo, in either `model_population` or `model_conformance`, covers a redefinition cycle for `redefinition_reaches`.
   - r03 covers only a write with no redefinition record at all, which is zero hops. It does not cover a walk that takes one or more hops and then fails to reach the grant.
   - **Failure scenario:** replace `for _ in 0..=bound` with `loop`. The full suite stays green, but a bundle with a cycle then hangs `check_operation_redefinition`.
   - **Fix:** add two tests, each asserting the whole outcome: `C.x -> B.x` with no `B.x -> A.x` must refuse with `Completed(Refused([EffectEscape{C.x}]))`, and the cycle `C.x -> B.x -> C.x` must refuse with the same outcome.

3. **Low (style). r09 matches and panics instead of using `assert_eq!` on the whole outcome.** File: `tests/model_conformance.rs:1164`. The match targets one exact unit variant, so the test is not vacuous. But `ConformanceCheckOutcome` derives `PartialEq`, so `assert_eq!(check_operation_redefinition(..), ConformanceCheckOutcome::Completed(ConformanceOutcome::Compatible))` is the direct form.

### Checked and clean

- The trace tag `#[trace("TC-196", "FR-151-AC-4")]` cites the right AC: effect variance is checked independently. FR-151's rule `quire.model.conformance.effect/v1` says "reaches one through redefinition records", and the fix matches it.
- No scope creep. #173, competing operation redefinitions, is untouched.
- Code comments describe the code. The walk's direction, bound and cycle handling match its doc comment.
- The only closing keyword in the commit and PR body is `Closes #171`.
- r09 fails without the fix: I restored `main`'s `conformance.rs` by hand and r09 refused with `EffectEscape`. It passes with the fix.
- Process: the author revert-probed with bare `git stash`, which is not allowed in this repo. `git stash list` in `qsl-171-redefinition-chain` is empty, so no stash was left behind.

### Gates at `f2c6136`, fresh worktree with its own target dir

- `make ci`, which runs fmt, clippy with default features and with `--all-features` under `-D warnings`, the tests in both modes, and the clean build: passed, rc=0.
- `make revendor-check`: `native-v1: 69 matched, 0 drifted, 0 stray`; `complete-value: 64 matched, 0 drifted, 0 stray`.

Findings 1 and 2 are the tests needed to pin the behaviour this PR introduces. Please fix them in this PR.

kreneskyp:
## Delta review verdict for PR #177: `f2c6136..3da837d` — MERGE

This covers only the delta since my last review. I checked each claim myself in a fresh worktree, using hand edits that I reverted by hand (no `git stash`).

| Finding from last round | Status | How I verified it |
| --- | --- | --- |
| 1. Equality not pinned at conformance | Fixed | Made the conformance closure compare identity only → only `r10` fails |
| 1. Equality not pinned at population | Fixed | Made the `field_write_covered` closure compare identity only → only `enforce_frame_refuses_a_field_write_at_a_revision_the_declared_grant_does_not_name` fails |
| 2a. No test for a chain that misses the grant | Fixed | Changed the no-further-record exit to `return current != *field` → only `r11` fails |
| 2b. No test for a cycle | Fixed | Replaced `for _ in 0..=bound` with `loop` → `r12` hangs until `timeout 30` kills it. Changed the post-bound `false` to `true` → only `r12` fails |
| 3. r09 not using `assert_eq!` | Fixed | r09 now uses `assert_eq!` |

All four new tests use `assert_eq!` on the whole outcome, including the full `AxisFailure` and `ModelRefusal`. The trace tags match their neighbours: `TC-196`/`FR-151-AC-4` for conformance and `FR-046-AC-3` for the frame tests. The only closing keyword is still `Closes #171`.

Gates at `3da837d`: `cargo fmt --check` is clean. `cargo clippy --workspace --all-targets --all-features -D warnings` is clean. `model_conformance` has 20 tests and `model_population` has 34, and all pass.

**Non-blocking nit (Low), `src/model/population.rs:1084`:** the doc comment says `field_write_covered`'s walk went from "one hop to unbounded, both fixed here together". That is not what happened: `main` already had the bounded full-chain walk from #168 (`for _ in 0..=bound`). This PR changed only the conformance call site. The same sentence appears in the PR body ("only its walk's reach (one hop to unbounded) changed"). It is a one-line doc fix and does not change behaviour.

PR #178, opened 2026-09-19
Charge phase-4 redefine facts as normalize.fact

### LINKED ISSUE #169: Charge phase-4 redefine facts as normalize.fact (value-accounting.md:453)
Found reviewing PR #167 (comment 5735840110). This predates that PR.

`value-accounting.md:453` charges every derived fact as `normalize.fact`. The redefine facts that phase 4 (dominance resolution in `src/model/normalize.rs`) produces are never charged, so a model's recorded work units come out below the priced amount.

Scope:
- Charge each phase-4 redefine fact as `normalize.fact` at the spec price.
- Add a test that pins the charge sequence on the TC-195 N06 resolved fixture.

Part of #120.

kreneskyp:
## Summary

Phase 4's `RULE_REDEFINE` derivation facts (built by `apply_redefinitions` — one on the redefining feature and one on the redefined feature per redefinition record reaching a type, `model-complete.md:231`) were constructed into the view but never charged. `value-accounting.md:453` prices every derivation fact as `normalize.fact` regardless of which phase formed it; `:455`/`:456` place these strictly after `normalize.redefinition-check` and before `normalize.conflict-check` in the charge sequence.

`Built::phase4_facts` collects these as `PendingFact` entries during `apply_redefinitions` (field redefinitions only — operation-member redefinitions never reach `Fact` construction, unchanged from #167's scope). `charge_all` replays them sorted by `sort_facts`, continuing the same `fact_count`/`derivation_facts` sequence phase 2/3 already run, between the redefinition-check loop and the conflict-check loop. The view and every identity it produces are unchanged — this only adds the missing charge.

## Changed test numbers (explained)

Three existing N06 fixture totals shift because they now include phase-4 fact charges that were previously silently free:

- `n06_conflict_check_charges_exactly_sigma_c_minus_1_times_f_o`: 100 → 110 work units (+10 facts: 2 at B, 2 at C, 6 at D). The `work_units=54` Incomplete boundary moves to 64; `next_charge`/`charge_point` unchanged (conflict-check, ordinal 18).
- `n06_a_single_redefiner_admits_no_conflict_check_charge`: 37 → 39 work units (+2 facts, one redefinition record). The `work_units=19` boundary is unaffected — it fires during `normalize.redefinition-check`, before any phase-4 fact charge.
- `conflict_check_charges_interleave_field_and_operation_groups_by_target_key`: 89 → 95 work units (+6 facts, field groups only: 2 at B, 4 at C). The `work_units=58` Incomplete boundary moves to 64; `next_charge`/`charge_point` unchanged (conflict-check, ordinal 6).

New test `n06_redefine_facts_are_charged_as_normalize_fact_between_the_two_phase4_checks` asserts the full N06 total (30 derivation facts, 110 work units) and an exact-bound `Incomplete` at the first phase-4 `normalize.fact` charge under a `work_units=54` budget.

`operation_redefinition_is_charged_like_a_field_redefinition`, `operation_redefinition_group_with_two_or_more_redefiners_is_charged_a_conflict_check`, and `n06_conflict_under_a_tight_fact_budget_is_incomplete_not_a_phase4_refusal` are unaffected (no field redefinitions, or denied before phase 4 for other reasons).

## Scope

`src/model/normalize.rs` and `tests/model_normalization.rs` only. Does not touch `conformance.rs` or dispatch/expression files (other #120 slices).

## Gates

- `cargo fmt --all -- --check` — pass
- `cargo clippy --locked --workspace --all-targets --all-features -- -D warnings` — pass
- `make ci` — pass (fmt/clippy/test default features, all-features, clean build + fixture-audit self-test + parse example)
- `make revendor-check` — pass (native-v1: 69 matched, 0 drifted, 0 stray; complete-value: 64 matched, 0 drifted, 0 stray)

Closes #169
Part of #120


kreneskyp:
## Review verdict: FIX

This covers code review, Rust review and gap analysis of the diff at `30cf4b1` against `main` (`2744e7b`).

**Gates, all run in a fresh worktree:** `cargo fmt --check` passes. `cargo clippy --locked --workspace --all-targets --all-features -D warnings` passes. `make ci` passes. `make revendor-check` passes (native-v1 69/0/0, complete-value 64/0/0).

**Revert probe:** I removed the new phase-4 `normalize.fact` loop by hand. Four tests fail: the new test and the three retotalled tests (at 1634, 1560, 2149 and 1702). The tests do gate the charge.

### What checks out
- **Placement is right.** `value-accounting.md:455` puts every `normalize.redefinition-check` "before its first `normalize.fact`". `:456` puts every `normalize.conflict-check` "after its last `normalize.fact`". `:453` orders facts phase 2, then 3, then 4, each ascending by (owner, declaration, inputs). So all redefinition checks run first, then all phase-4 facts, then all conflict checks. Nothing is interleaved per record. The amount is `derivation_facts=k` plus one work unit. `sort_facts` uses the key `:453` states.
- **The new totals are correct.** I recounted them from the fixtures:
  - N06 resolved: (T, record reaching T) pairs are B:1, C:1, D:3, so 5 pairs × 2 = 10 facts. That gives 100→110, the 54→64 boundary, and 20→30 derivation facts (8 qualify + 12 inherit + 10).
  - Single redefiner: 1 pair gives 37→39. The 19 boundary is still at the redefinition check (8 records + 5 qualify + 6 phase-3).
  - Interleave: B:1 and C:2 pairs give 6 facts. That gives 89→95 and 58→64.
- **Replay order is deterministic.** When phase-2/3 `derivation_facts` runs out, the result is still `Incomplete`, and the existing test passes. A phase-4 refusal still runs no normalize charges. The meter reads 0/0 on `fixture_n06_conflict`.
- The only closing keyword is `Closes #169`.

### Findings

1. **High.** `src/model/normalize.rs:1163-1177` together with `normalize_with_meter` at `:2015-2022`.
   - **Problem:** the new phase-4 fact charges are never reached for any bundle that phase 4 refuses. `build()` returns the `derivation-conflict` or `redefinition-target` refusal before `charge_all` replays anything.
   - **Scenario:** run `fixture_n06_conflict` with `derivation_facts = 19` (exactly its phase-2/3 fact count). It returns `Refused(DerivationConflict)`. The spec requires `Incomplete { DerivationFacts, limit 19, consumed 19, next_charge 20, NormalizeFact }`. At `value-accounting.md:480-485`, the first unavailable charge ends checking and "reports no refusal found in the unfinished stage". The conflict is found at conflict-check, after the phase-4 facts. Before this PR, 19 was enough and `Refused` was correct. This PR's own charge makes it wrong.
   - **Related existing bug, same cause:** a `work_units` sweep on `fixture_n06_conflict` returns `Refused` for every w in 16..=56. By the spec, everything below 57 is `Incomplete`: 40 through phase 3, +5 for redefinition checks, +8 for phase-4 facts, +4 for the conflict check. The 45..=52 range is new with this PR. The 16..=39 range is on `main` already and breaks #167's own invariant: `fact_budget_exceeded` compares the fact count with `work_units` and ignores the work units spent on records and cycle checks.
   - **Fix:** keep the phase-4 refusal inside `Built` instead of returning `Err`. Have `charge_all` charge through the conflict check, and return the refusal only once every phase-4 charge is admitted. This fixes both ranges. Add a test asserting the full `Incomplete` for `derivation_facts = 19` on the conflict fixture, plus a `work_units` case inside 45..=52.
2. **Low.** `tests/model_normalization.rs:1627`.
   - **Problem:** the new test is tagged `TC-196`/`FR-151-AC-2`. That criterion covers inheritance cycles and variance refusals, not normalization charges. The right tags are `TC-195` and `FR-150-AC-8`.
   - **Scenario:** a trace audit reports FR-151-AC-2 as backed by a test that never exercises it.
   - **Also:** FR-150-AC-8 has the shape "exact bound completes, one-less is incomplete". The test only uses `matches!(Completed(_))` under `UNLIMITED`. It should assert that `work_units = 110` completes and `109` is `Incomplete` at the last charge.
3. **Low.** `src/model/normalize.rs:1766-1768`.
   - **Problem:** `phase4_facts` is cloned and sorted only so the facts can be counted. Each charge carries no per-fact data, so the order cannot be observed. The same function's own comment (#141 F11, `:1720`) calls this pattern dead work. Each `PendingFact` also duplicates a clone of the path inputs.
   - **Scenario:** on a wide bundle this spends an extra O(n log n) and doubles the memory held for these facts, to no effect. Charging `built.phase4_facts.len()` times, or storing only a count, gives the same result.
4. **Low.** The comments narrate review history instead of describing the code:
   - `src/model/normalize.rs:159-165`: "already built these facts into the view before this fix; the fix … only adds the missing charge".
   - `tests/model_normalization.rs:1557-1559` and `:2147-2148`: "#169: 100 before this fix".
   - The new test's doc: "Before this fix…" and "the pre-fix defect…".

   **Scenario:** a later reader learns what the code used to do, not what it does.

### Answers to the scope questions
- **Field-only scope:** operation-member redefinitions really never become `Fact`s. `operation_edges` only feeds the conflict-check charge. But `model-complete.md:231` derives two redefine facts for every redefinition record, operation records included. Operation members also get no phase-2 qualify facts and no phase-3 inherit facts (`:229-230`). So those charges are missing too. The field-only choice matches how the whole pass treats operations, so I'm listing it as a ticket, not a PR finding.

### Out-of-scope gaps (ticket suggestions)
- **A.** Operation members produce no qualify, inherit or redefine facts, and so no charges (`model-complete.md:229-231`). For example, `fixture_operation_redefinition` comes out at 18 work units, below the spec price. #173 covers only detecting competing redefinitions, not this.
- **B.** Phase-4 `quire.model.normalize.subset/v1` facts (`model-complete.md:230`: one fact per subsetting record reaching T) are neither derived nor charged. The module doc at `normalize.rs:11-17` says subsetting derives no normalization fact, which contradicts the vendored table. When these facts are added, they must sort into one phase-4 sequence together with the redefine facts.
- **C.** The work_units ranges 16..=44 and 53..=56 from finding 1 are already on `main`. The fix for finding 1 covers them. If it isn't done in this PR, it needs its own ticket.

kreneskyp:
## Delta review (30cf4b1..a2ac457): FIX

**Scoped gates.** `cargo fmt --check` and `clippy --all-targets --all-features -D warnings` both pass. The `--lib` tests pass, as do `model_normalization`, `model_population`, `model_systems`, `model_reference_queries`, `model_conformance`, `model_dispatch` and `native_population_emission`.

### Checked and confirmed

**Sweep of `fixture_n06_conflict`, every value, not samples:**
- `work_units = 0..=56` all return `Incomplete`. The charge points follow the spec order: record (0-12), then phase-2/3 facts and cycle checks (13-39), then redefinition-check (40-44), then phase-4 fact (45-52), then conflict-check with next=4 (53-56). `57` returns `Refused(DerivationConflict)`.
- `derivation_facts = 0..=26` all return `Incomplete` at `normalize.fact`, with next = d+1. `27` returns `Refused`. That matches 19 facts from phases 2-3 plus 8 from phase 4.
- The R07 same-owner fixture sweep (0..=33) also returns `Incomplete` up to 30. It returns `Refused(RedefinitionTarget)` from 31, which is its full charge total.

**Revert probe (hand edit, undone after).** I returned `phase4_refusal` at the top of `charge_all`. `n06_conflict_refusal_waits_for_every_phase4_charge_to_admit` fails, and 38 others pass.

**Earlier findings:**
- The clone and sort are gone: `phase4_fact_count: u64`.
- The trace tag is now `TC-195`/`FR-150-AC-8`.
- The 109/110 exact-bound pair asserts the whole `Incomplete` with `assert_eq!`.
- `object_universe` surfaces the deferred refusal.

### Findings

1. **Medium.** The deferral fix misses one refusal. At `src/model/normalize.rs:1604-1616` (target not an effective member) and `:1624-1636` (redefining member not effective), `apply_redefinitions` still `return Err`s out of `build()`.
   - **Why it matters.** `model-complete.md` refuses this case: "a redefinition whose target is not a member inherited by its owning type ... `redefinition-target`". `normalize.redefinition-check` exposes it (`value-accounting.md:455`), so it is a phase-4-stage refusal. It must wait for the stage's charges just like the dominance refusals.
   - **Repro.** Take `fixture_n06_conflict` and add type `E` (no generalization), field `E.y` of type `A`, and `redef.E` (`E.y` redefines `A.x`).
     - `work_units = 30` returns `Refused(RedefinitionUnreachable)` with 0 charges replayed. The spec requires `Incomplete` inside phase 3.
     - Under `UNLIMITED`, the conflict D found first is now dropped in favour of E's early `Err`.
   - **Scope.** This is the same defect class as round-1 finding 1, which named `redefinition-target`. The new module doc (`:165-178`) claims phase-4 refusals are never returned by `build`, which is only true for the dominance refusals.
   - **Fix.** Store the refusal in `accounting.refusal`, skip that group with `continue`, and keep the first one found.
   - **Separate ticket.** The code/cause is `dangling_reference`/`RedefinitionUnreachable`, where the spec says `invalid_model_binding`/`redefinition-target`. That mismatch was already on main, so it belongs in its own ticket.
2. **Low.** `tests/model_normalization.rs:2243-2269` and the in-test comments still narrate review history: "PR #178 review finding 1", "new with this fix", "already broken on `main` before this PR", "its pre-review shape". Round-1 finding 4 asked for comments that describe the code.
3. **Low.** `tests/model_normalization.rs` (the new test) asserts that `work_units = 57` gives `Refused`, but not the one-less `56`, which should be `Incomplete{WorkUnits, 56, 56, next_charge 4, NormalizeConflictCheck}`. That is the exact-bound shape FR-150-AC-8 cites. My sweep confirms it passes today, but no test pins it. The `57` arm also checks only `refusal.cause`. It should `assert_eq!` code and cause.

kreneskyp:
## Delta review (a2ac457..27a7fcf): FIX (one Medium finding)

The merge of main (`68bdacb`) is clean. It brings in #177's conformance/population changes, and the only files it touches are those. The PR's own delta is `normalize.rs` and `tests/model_normalization.rs`.

**Gates.** Everything scoped passes:
- `cargo fmt --check`
- `clippy --all-targets --all-features -D warnings`
- `--lib` tests
- test targets: `model_normalization`, `model_population`, `model_systems`, `model_reference_queries`, `model_conformance`, `model_dispatch`, `native_population_emission`

### Verified by probe

**`fixture_n06_conflict_with_unreachable_redefiner`, every value.** The run completes at 65 work units and 29 facts.
- `work_units = 0..=64` all return `Incomplete`, in spec order:
  - record: 0-15
  - phase-2/3 fact and cycle-check: 16-44
  - redefinition-check: 45-52, amounts 2, 3, 3; redef.E is `m=1`, `r=2`
  - phase-4 fact: 53-60
  - conflict-check: 61-64, `consumed=61`, `next=4`
- `65` returns `Refused`.
- `derivation_facts = 0..=28` all return `Incomplete` at `normalize.fact`. `29` returns `Refused`.

**Unreachable-only bundle** (F1 plus E, E.y and redef.E). `work_units = 0..=15` all return `Incomplete`; 15 lands at redefinition-check. `16` returns `Refused`. `derivation_facts = 0..=6` all return `Incomplete`, and `7` returns `Refused`.

**`consumed: 53` is correct.** `Incomplete.consumed` is "that counter's consumed value before the denied charge" (`accounting.rs:212`). The sweep shows the same thing, for example `W32 → consumed=31, next=2`. My round-3 note that said `consumed: 56` was wrong. The author is right.

**No early `Err` remains in `apply_redefinitions`.** Every refusal site (`:1507`, `:1616`, `:1645`) stores the refusal in `accounting.refusal` if it is `None`, then continues.

**Phase-3 over phase-4 precedence holds.** `phase3_specialization_cycle_refusal_wins_over_phase4_derivation_conflict` passes.

### Findings

1. **Medium: the kept refusal follows type order, not charge order.** `src/model/normalize.rs:1507/1616/1645`, pinned by the doc at `tests/model_normalization.rs:2424-2429` and the UNLIMITED arm of `n06_unreachable_redefinition_target_also_waits_for_every_phase4_charge`.
   - **The policy.** "First found" means type order: D, then E. So a completed run reports D's `derivation-conflict` and drops E's unreachable target.
   - **What the spec says.** `value-accounting.md` says "every refusal that work exposes is reported, in charge order". E's refusal is exposed by redef.E's `normalize.redefinition-check` ("before testing that its target is a member inherited by its owning type", `:455`). That check is charged at W50. D's conflict is exposed by the `normalize.conflict-check` after it (W61). So in charge order E comes first.
   - **Why it matters.** The new test locks in the reverse.
   - **Minimal fix.** Keep one refusal slot, but rank redefinition-check-exposed refusals (unreachable target or redefining member, ordered by record producer key) ahead of conflict-check-exposed refusals (ordered by effective member key). Then flip the test's UNLIMITED expectation.
   - **Later.** Reporting *every* refusal needs `Vec<ModelRefusal>` and belongs in a ticket.
2. **Nit: `apply_redefinitions` can no longer fail.** It still returns `Result<(), ModelRefusal>` (`normalize.rs:1343`) and the caller still uses `?` (`:1194`), but no path returns `Err`. It could return `()`.
3. **Nit: the redefining-member branch leaves the group half-processed.** `normalize.rs:1639-1660` runs `continue 'targets` partway through the edge loop. Earlier edges' facts stay counted and the target is not hidden. The output is still a refusal, so nothing observable changes. A comment saying the partial count is intentional would help.

### Ticket suggestions (out of scope, already on main)
- **Phase-3 cycle refusal is still returned early from `build()`.** It comes via `ancestor_paths(..)?`, `normalize.rs:974`. On `fixture_n06_conflict_with_unrelated_cycle`, `work_units = 20/25/30` return `Refused(specialization-cycle)`, and none of the phase-2/3 charges are replayed. This is the same defect class as round-1 finding 1, one phase earlier.
- **Report every phase-4 refusal in charge order.** This needs `Vec<ModelRefusal>`.
- **Code/cause mismatch.** `RedefinitionUnreachable`/`dangling_reference` should be `invalid_model_binding`/`redefinition-target`, per `model-complete.md`.

kreneskyp:
## Final narrow check (27a7fcf..b4fbe2a): FIX, one Medium finding

The merge of main (`ef38675`, #172) is clean. The PR's own delta is only `normalize.rs` and `tests/model_normalization.rs`.

**Gates:** fmt and clippy (`--all-targets --all-features -D warnings`) pass. The scoped tests pass: `--lib`, `model_normalization`, `model_population`, `model_systems`, `model_reference_queries`, `model_conformance`, `model_dispatch`, `native_population_emission`, `dispatch_calls`.

### Question 1: is each site's stage and key right?

- **`:1616` (target unreachable) is correct.** It is ranked at redefinition-check and keyed by the least `record_key` in the group. At a given type, a group's target is missing only when no owner in that group inherits it. So every record in the group fails its own check (`value-accounting.md:455`), and the least record key is the first of those checks to be charged.
- **`:1645` (redefining member unreachable) is correct.** It is ranked at redefinition-check and keyed by that edge's own record.
- **`:1507` (dominance) has the right stage, conflict-check (`:456`). The key is wrong: see finding 1.**

### Question 2: when two refusals come from the same stage, is the first-charged one reported?

- **Two refusals from redefinition-check work.** I added records `redef.E` (type E) and `redef.A0` (type F, which is processed after E). F's refusal is reported, which is correct because its record key sorts first.
- **Two refusals from conflict-check fail.** Details below.

### Findings

1. **Medium: the conflict-check rank omits the effective type.** `src/model/normalize.rs:1657` passes `&target_key` alone.
   - **Why it's wrong.** `:456` charges conflict-checks "ascending by effective member key". `model-complete.md:160` defines that key as `(owner effective type identity, original declaration producer key)`, so the type comes first. The code itself charges conflict-checks per type, in `type_keys` order, then by target.
   - **Probe.** I took `fixture_n06_conflict` (D conflicts on `A.x`) and added a second diamond: root `K`, redefiners `K1` and `K2`, bottom type `B9` conflicting on `K.w`.
     - The meter charges B9's conflict-check first (W118, amount 5), then D's (W123, amount 4).
     - Yet UNLIMITED and W127 both report D's `derivation-conflict`, because `A.x < K.w`.
     - So the reported refusal is not the first one charged.
   - **Fix.** Rank conflict-check refusals by `(type_key, target_key)`, the same order the charges are pushed in. Add this two-type case as a test.
2. **Low: a tie on the same record can name the wrong owner.** `normalize.rs:1302-1318` / `:1616`.
   - **What happens.** When a descendant of the record's owner sorts before the owner, both produce the same rank `(stage, record_key)`. Strict `<` keeps the first one processed, which is the descendant.
   - **Probe.** With `Da <= E`, the output reads "not an effective member of model.Da". The spec tests the record's *owning type*, which is E.

### Ticket suggestions (out of scope, already on main)
- **Conflict-check charge order uses the wrong key.** `conflict_check_work` is pushed in producer-key type order. The spec (`model-complete.md:160`) orders by owner **effective type identity**. The same may apply to `normalize.declaration`.
- **Carried from the last round, still open:**
  - The phase-3 cycle refusal is returned early from `build()`.
  - Only one phase-4 refusal is reported when there are several.
  - The `RedefinitionUnreachable` code and cause should be `redefinition-target`.

Issue #159, opened 2026-09-18
Re-vendor resources/complete-value past QSpec #86 (TC-196 R01/R07 stale)


kreneskyp:
resources/complete-value is pinned at QSpec 7d7943ab, before QSpec #86. Two vectors differ on origin/main:

- **R01:** the cycle listing names types, not records.
- **R07:** adds a second shape. `B/z` and `B/z2` both redefine `A/x`, and must produce the same `redefinition-target` refusal.

PR #157 implements the origin/main shapes. This ticket re-vendors the snapshot with `cargo xtask revendor` so the vendored text matches.

Found in the review of #157 (A04 slice 4).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

kreneskyp:
Scope addition: the operation frame vocabulary.

- QSL's `resources/complete-value` is pinned at QSpec 7d7943ab. There, FR-151 and `model-complete.md` spell the operation effect as a producer-supplied `{fieldWrites, creates, deletes}`. QSL's `src/model/bundle.rs` `OperationEffect { field_writes, creates, deletes }` follows that spelling.
- QSpec main (d227270) states it as the **declared frame `{modifies, creates, deletes}`**.
  - FR-151:78.
  - `model-complete.md` Frames row: each `modifies` entry names a field *or relationship*, and each `creates`/`deletes` entry names an object type *or a process* (FR-208).
  - FR-340 (CheckedPackage V2 wire) and Contract IR (ir #127, `FrameMember { Modifies, Creates, Deletes }`) use the same three keys, in the tie-break order `modifies` → `creates` → `deletes`.

The frame text came in with QSpec #99 (56c3e0b). The re-vendor target is QSpec main, currently d227270, so the delta is everything between 7d7943ab and that pin. This re-vendor must therefore also:
1. rename `OperationEffect.field_writes` to `modifies` everywhere in src/model (bundle, conformance effect axis, normalize, and PR #168's `admit_invocation` frame check). This is spec-driven, not a churn rename;
2. admit relationship entries in `modifies` and process entries in `creates`/`deletes`, with FR-272 `malformed-declaration` for any other meaning, per the Frames row;
3. check the `conformance.axis` effect-axis charge text in `value-accounting.md` at the new pin.

Raised by agent-e-claude against PR #168.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #179, opened 2026-09-19
Re-vendor complete-value past QSpec #86

kreneskyp:
## Summary

Re-vendors `resources/complete-value` from `7d7943ab1482e091f6d126401ada38957c4a1ccf` to `d227270fbeb28289df6abba7e94173118345c028` (`origin/main` of agent-ix/quire-specification at the time of this PR; QSpec PR #86 renamed TC-196 R01's cycle listing to name types rather than generalization records and added R07's second contending-redefiner shape, as #159 asks) via `cargo xtask revendor --tree complete-value --qspec-clone <clone>`, then brings QSL's code and tests in line with the re-vendored text.

`cargo xtask revendor` wrote all 34 changed `quire-specification` paths (24 unchanged, 6 external unicode files verified unchanged, 0 removed); `VENDOR.json`'s digests and `commit` were updated by the tool, not by hand.

## QSpec commit used and why

`d227270fbeb28289df6abba7e94173118345c028` — this is `origin/main` of agent-ix/quire-specification, and is the fallback commit #159 names ("or to d227270 if main has moved past something unrelated"). It contains QSpec PR #86 as an ancestor.

## QSL changes required

- **Operation-frame grant rename `fieldWrites` → `modifies`** (value-accounting.md renamed the grant). `src/model/bundle.rs:` `OperationEffect.field_writes` → `.modifies`; `src/model/population.rs` (all `.field_writes` accesses plus two runtime refusal messages), `src/model/refusal.rs` (doc comments on `ModelRefusalCause::FrameTypeChanged`/`FrameFieldWriteOutsideGrant` — variant names kept as-is, out of scope), `src/model/normalize.rs:765,1734`, `src/model/conformance.rs:671,674,681,948`, and `tests/{model_normalization,model_conformance,model_reference_queries,model_population}.rs` (struct-literal fields, helper parameter names, doc comments).
- **`src/value/package_identity.rs:38-47`**: added `"declaration"` to `NODE_OPTIONAL`. The revendored `checked-package-v2/fixtures/positive-nominal-identities.json` now carries a top-level `declaration:{qualified_name}` member on scalar_type nodes (part of the new `ApplicationNode` preimage variant in `node-identity-preimage.schema.json`), which QSL's strict member-checker rejected as `UnknownMember("declaration")`. The fixture's `nominal_identity_preimage.qualified_declaration` is unchanged, so export-name derivation still works unmodified; this member is accepted and ignored rather than interpreted, since QSL doesn't implement `ApplicationNode` (see Out of scope). Fixes `tests/library_resolution.rs::l01_export_node_keys_derive_from_a_checked_package_v2_identity_projection`.
- **`src/value/node.rs`**: `node-identity-preimage.schema.json`'s `ModelOwner` changed from `{kind, authority, identity, export}` to `{kind, identity, node}`. Renamed `ModelSubject{authority,identity,export}` → `{identity,node}`, and `NodeOwner::is_well_formed`/`canonical()`/`CanonicalOwner` to match (only `authority` is now optional per owner kind: Source/Definition carry it, Model carries `node` instead). No production code constructs a `Model`-kind owner today — the only call sites were this struct definition and its `mod.rs` re-export — so this is a contained shape rename. Updated the three `"kind":"model"` JSON literals in `tests/text_enum_identity.rs:587,589,604` to match. Fixes `tests/text_enum_identity.rs::enum_node_identity_vectors_reproduce_and_noncanonical_preimages_refuse`.
- **`tests/complete_value_lock.rs`**: recomputed all 21 `PINNED_NON_CATALOG` digests from the updated `VENDOR.json`, updated the doc comment's pinned commit, and updated the mutation test's hardcoded digest-fragment to the new `edition.md` digest.
- **`tests/exact_decimals.rs`**: `value-accounting.md` dropped the `normalize.unsupplied-item` charge point (producer-interface-1.2.0 vocabulary removed) — removed from `DEFERRED_POINTS`. It also added a "Collection sum schedule" section whose prose incidentally text-matches `sum.quantity` against the charge-point regex; that string is an FR-322 sum-operator application identity, not a `ChargePoint`, so it's excluded via a new `NON_CHARGE_POINT_MATCHES` const rather than folded into `DEFERRED_POINTS`/`DEFERRED_FAMILIES` (which would misrepresent it as tracked by another accounting system).
- **`src/value/definition.rs`, `resources/complete-value/README.md`, `docs/dependencies.md`**: updated hardcoded references to the old pinned commit.

TC-196 R01 and R07 were already implemented (from PR #157) — only the vendored bytes needed to catch up; no Rust changes were required for either.

## Out of scope

The re-vendor also brings in a much larger normative vocabulary overhaul that QSL's model layer does not implement and this PR does not attempt:

- The "domain package" rename family: `model-effective-declaration.schema.json`'s `ProducerKey`/`ModelSelection` → `DeclarationKey`/`DomainPackageRef`, `bundle` → `domain-package`, `generalization-record` → `supertypes`, `closedWorld` → `extent`.
- `node-identity-preimage.schema.json`'s new `ApplicationNode` preimage variant (FR-307 expression/application node identities) and its own `declaration` member — QSL currently accepts-and-ignores the outer node object's `declaration` member (see above) but implements none of the `ApplicationNode` semantics.
- `model-complete.md`'s D05 abstract-type rework, `Pre:`/`Post:` contract syntax, `Populations`/`Relationships`/`Business-model-meanings` sections, and FR-153's new AC-8 "Outputs clause."

These need their own ticket(s); they're not small and not required for any currently-failing test to pass.

The historical Plan-013 campaign records (`spec/functional/FR-055-govern-complete-v1-lane.md`, `spec/integration/IT-011-adopt-complete-v1-baseline.md`, `plan/Plan-013-complete-v1-delivery/plan.md`, `tests/complete_v1_plan.rs`) reference the original baseline commit with specific derived counts (83 capability identities, "18 complete, 36 partial and 29 missing") and are left untouched as intentional historical record; `tests/complete_v1_plan.rs`'s `PLAN.contains("7d7943ab...")` assertion still passes since `plan.md` is unchanged.

## Gates

- `cargo fmt --check`: clean
- `cargo clippy --workspace --all-targets --all-features -- -D warnings`: clean, no warnings
- `make ci` (`ci-default-features`, `ci-all-features`, `ci-clean-build`): all green, 0 failures
- `make revendor-check` (`cargo xtask revendor-check --tree all`): `native-v1: 69 matched, 0 drifted, 0 stray`; `complete-value: 64 matched, 0 drifted, 0 stray`
- `cargo test --workspace --all-features`: all green (run twice, once before each failure fix and once clean at the end)


kreneskyp:
## Review of #179 at 421bfd0: FIX

Scope: this PR's diff against `main`, plus the vendored spec text that the diff re-pins. I ran a code review, a Rust review and a gap analysis.

### What passes
- **The re-vendor is byte-exact.** All 58 files under `resources/complete-value/quire-specification/` byte-match `git show d227270:<path>` from agent-ix/quire-specification. Each file's SHA-256 equals its `VENDOR.json` entry. Nothing under `resources/native-v1/` changed. The only non-vendored edit under `resources/` is QSL's own `resources/complete-value/README.md`. d227270 is still QSpec `origin/main`.
- **The frame rename is complete.** No `fieldWrites` or `field_writes` is left in `src/` or `tests/`, including error strings. `OperationEffect` has no serde, so no alias accepts both spellings. `modifies` matches the vendored spelling (model-complete.md:162, FR-151:78, TC-196, TC-198 L07).
- **ModelOwner matches the schema.** It is `{kind, identity, node}`, required, with `additionalProperties: false` (node-identity-preimage.schema.json:20), and `ModelSubject` uses `deny_unknown_fields`. No struct was renamed; only fields changed. The canonical key order `authority?, identity, kind, node?` is valid JCS.
- **exact_decimals.**
  - `normalize.unsupplied-item` is absent from the vendored text.
  - `sum.quantity` at value-accounting.md:432 is shorthand for the operation identity `quire.op.collection.sum.quantity`, not a charge point. The exclusion is correct.
- **Pinned digests.** I recomputed all 21 `PINNED_NON_CATALOG` digests from the files and all match. The `4cf0b7ac51a3b941` mutation fragment is `edition.md`'s digest.
- **TC-196 R01 and R07.** The R01 test asserts the type listing `[model.A, model.B]`. The R07 two-redefiner shape is tested through `normalize`.
- **Closing keywords.** Only `Closes #159`.
- **Gates** (fresh worktree, own target dir): `make ci` passes (fmt, clippy default and `--all-features -D warnings`, tests, clean build). `make revendor-check` reports `native-v1: 69 matched, 0 drifted, 0 stray` and `complete-value: 64 matched, 0 drifted, 0 stray`.

### Findings

**1. High: `declaration` is accepted and ignored, but the spec gives it meaning.** At `src/value/package_identity.rs:38-48`, `declaration` was added to `NODE_OPTIONAL` and nothing reads it. The vendored text says what the member means:
- `proposals/checked-package-v2/README.md:208-220` (§Declarations): "The member is part of the node identity. On a nominal node it equals the nominal `qualified_declaration` (else `declaration-nominal-mismatch`, `invalid_package`), and two nodes declaring one name refuse as `ambiguous-name` (`ambiguous_declaration`). Library exports resolve only through this member."
- `FR-307:70-86`: `Name` "resolves only through that package's FR-322 `declaration` members". A library whose "declaration that disagrees with its nominal qualified declaration is already refused by the FR-322 reader and exports nothing."
- `node-identity-preimage.schema.json:17`: `Declaration` = `{qualified_name: QualifiedName}` with `additionalProperties: false`.
- `native-diagnostics.md:82,108-122`: the `declaration-nominal-mismatch` payload, and its place in the V2 reader's refusal order (schema, then graph, then stale key, then **declaration-nominal-mismatch**, then **ambiguous-name**).

I ran a probe through `resolve_libraries` on the vendored `positive-nominal-identities.json`, with `package_id` recomputed for each case. All four of these return `Ok`:
- the baseline fixture (expected to admit);
- the `Example::Status` node with `declaration.qualified_name = ["Example","Other"]` (the spec says `invalid_package`/`declaration-nominal-mismatch`);
- `declaration: 42` (the spec says a schema refusal);
- `declaration: {qualified_name:[...], surplus:0}` (the spec says a schema refusal).

Before this PR, the member was refused as `UnknownMember`. This PR turns that refusal into silent admission of packages the spec requires QSL to refuse, and exports keep resolving through `nominal_identity_preimage`, the path FR-307 now forbids.

The doc comment also names the wrong schema member. It calls this member the `ApplicationNode` preimage's `declaration`, but the member is the projection node's own `declaration` defined in README §Declarations. The ApplicationNode's field is a separate one.

Fix, in this reader:
- validate the `Declaration` shape;
- refuse `declaration-nominal-mismatch` when `declaration.qualified_name` differs from the nominal `qualified_declaration`;
- detect duplicate names on `declaration.qualified_name` as `ambiguous-name`;
- resolve exports from `declaration`;
- add traced negative tests for each.

**2. Low: stale pin reference.** The canonicity doc at `src/value/package_identity.rs:251` still says "every name QSpec 7d7943a gives a projection node". The pin is now d227270, and that pin adds the `declaration` member name.

**3. Low: vague docs and a missing old-shape test on the model owner.** The `ModelSubject` field docs (`src/value/node.rs:105,107`) say "Nonempty identity" and "Nonempty node". README:75 defines them as the domain package identity and the IR node identity. `tests/text_enum_identity.rs:587-589` also lacks a negative case showing that the old model-owner shape `{kind:"model",authority,identity,export}` is refused.

**4. Nit: pin history in the README.** `resources/complete-value/README.md:5-9` narrates pin history ("the pin previously stood at 7d7943ab…"). State the current pin; Git holds the history.

### Out of scope (tracked by #131/#132, not blocking here)
- **TC-195..198 are re-specified around domain packages.** New fixtures, digests and row meanings arrived; for example, TC-195 N01 is now `3b79bb92`/`2cd5d1e7` rather than `f1cc59cd`/`9ae1232a`, and N08 now means something else. About 109 QSL tests trace those TCs and still assert the old values, e.g. `tests/model_normalization.rs` N01/N08 and `tests/model_dispatch.rs`.
- **`normalize.unsupplied-item` is still in the code.** The spec dropped it, but `src/model/accounting.rs:192` still defines it and `src/model/normalize.rs` still charges it. This PR removed it only from the `exact_decimals` bookkeeping.
- **New checked-package-v2 content has no QSL reader.** `ApplicationNode` and the `operation_vectors`, `operation_mutations` and `frame_mutations` in `node-identity-vectors.json` are unconsumed. So are the `Pre:`/`Post:` contracts, FR-153 AC-8, and the FR-145 `sum<N>` charge schedule.

### Conflicts with open PRs (information, not findings)
- **#177: textual conflict.** It conflicts in `src/model/conformance.rs` and `src/model/population.rs` (2 hunks). It also adds new `redefined.effect.field_writes` uses and `fieldWrites` doc text, which will not compile after #179 lands.
- **#172: merges cleanly but breaks the build.** It touches `tests/exact_decimals.rs` and `tests/model_reference_queries.rs` without textual conflict. However, #172 and #179 each change `DEFERRED_POINTS: [&str; 22]` to `[&str; 21]` by removing a different entry (`dispatch.select` vs `normalize.unsupplied-item`). The merged array will have 20 entries under a declared length of 21, so it fails to compile. The second PR to land must change the length to 20.
- **#178: clean** (`normalize.rs`, `tests/model_normalization.rs`).

kreneskyp:
## Delta review of #179, 421bfd0..a4d7660: FIX

### What checks out
- **`5cee46c` merge of #177.** `src/model/conformance.rs`, `src/model/population.rs` and `tests/model_conformance.rs` are byte-identical to `origin/main` (68bdacb) once `field_writes`/`fieldWrites` is rewritten to `modifies`. `tests/model_population.rs` differs only by one rustfmt reflow of a `panic!`. #177's `redefinition_reaches` walk and the r09–r12 and frame tests are preserved unchanged.
- **`DuplicateDeclaration` → `AmbiguousDeclaration` changes the refusal code, and the change is correct.** A duplicate name now refuses as `ambiguous_declaration`/`ambiguous-name` instead of `invalid_package`/`invalid-value`. That is what vendored `native-diagnostics.md:81` pins for a `quire.checked-package/v2` reader: equal `declaration.qualified_name` values give `ambiguous-name`, retaining the name and both node keys in ascending digest order. The keys come out ascending because node order is enforced first. The only test that pinned the old code is updated, and nothing else in src or tests references it.
- **Shape checks.** `declaration` must be exactly `{qualified_name}` (schema line 17), and exports derive from `declaration` (FR-307:70-86). The `composite_values` fixture fix and the old-owner-shape test are both right.
- **Gates.** These scoped tests pass: `library_resolution`, `text_enum_identity`, `composite_values`, `model_conformance`, `model_population`, `model_normalization`, `model_reference_queries`, `exact_decimals`, `complete_value_lock`. `make revendor-check` reports `native-v1: 69 matched, 0 drifted, 0 stray` and `complete-value: 64 matched, 0 drifted, 0 stray`.

### Findings

**1. High: a mismatch is emitted with the wrong cause.** At `src/value/library.rs:262,279`, `NodeDefect::DeclarationNominalMismatch` falls into the generic `InvalidPreimage` arm and is emitted as `invalid_package` with cause **`invalid-value`**. The spec pins a distinct cause: `invalid_package`/`declaration-nominal-mismatch` (checked-package-v2 README:218; native-diagnostics.md:82, where it is in the cause list). native-diagnostics.md:82 also says it "retains the node key, its `declaration.qualified_name` and its nominal `qualified_declaration`". The defect variant (`package_identity.rs:117`) carries none of those three values.

A probe using the vendored fixture with `Example::Status`'s `declaration.qualified_name = ["Example","Other"]` returns `code=invalid_package cause=invalid-value`. `tests/library_resolution.rs:731,737` pin that wrong cause for the two new mismatch cases.

Fix:
- add `LibraryCause::DeclarationNominalMismatch` (`declaration-nominal-mismatch`);
- give the defect the node key and both names;
- map the defect to the new cause;
- have the tests `assert_eq!` the whole refusal.

**2. Medium: refusals come out in the wrong order.** The loop at `src/value/package_identity.rs:326-339` checks each node completely (schema, then mismatch, then ambiguity) before moving to the next node. native-diagnostics.md:108-122 fixes a global order: all schema refusals first, then `declaration-nominal-mismatch` in ascending node-id order, then `ambiguous-name`.

Probe: the fixture with a mismatch on node 1 and an unknown member `surplus` on node 3 returns `Node { index: 1, DeclarationNominalMismatch }`. The spec requires the schema refusal on node 3. For the same reason, an ambiguity found at a lower index is reported ahead of a mismatch at a higher index.

Fix: run the schema pass over every node, then the mismatch pass, then the ambiguity pass. Add a test for each crossing.

**3. Low: two comments still describe the old behaviour.**
- The `UndeclaredExport` doc at `src/value/library.rs:138` says "matches no nominal declaration". Exports now resolve through `declaration`.
- The `ProjectedDeclarations` doc at `src/value/package_identity.rs:120` says "by nominal qualified declaration". The map is keyed by `declaration.qualified_name`.

**4. Low: a comment names a cause the code doesn't emit.** The module doc at `src/value/library.rs:12-15` and the `NodeDefect` doc at `src/value/package_identity.rs:112-116` say the package "is refused as `declaration-nominal-mismatch`". The code emits cause `invalid-value` (see finding 1). This resolves once finding 1 is fixed.

kreneskyp:
## Review round 3 (final): MERGE

Head 893c092. Range a4d7660..893c092, which includes the merge of main ef38675 (#172).

### Round 2 findings are fixed
- **High, wrong cause: fixed.** `PreimageDefect::DeclarationNominalMismatch { node, declared: Option<String>, nominal }` carries exactly what checked-package-v2 README:218 and native-diagnostics.md:82 require:
  - the node key;
  - `declaration.qualified_name` (`None` when it is absent);
  - the nominal `qualified_declaration`.
- **Code: correct.** The code is `InvalidPackage` and the cause is `declaration-nominal-mismatch`. native-diagnostics.md:82 lists this cause under `invalid_package`.
- **Medium, ordering: fixed.** `project_declarations` now runs three global passes, which match native-diagnostics.md:108-122 steps 1, 4 and 5:
  1. schema and node shape over all nodes;
  2. mismatch, by ascending node id;
  3. ambiguity.
- **Tests: both cases pinned.** `l08_b` and `l08_c` `assert_eq!` whole refusals for both cross-node orderings.

### DEFERRED_POINTS `[&str; 20]`: correct
- `normalize.unsupplied-item` is gone from the vendored value-accounting table, so it is right to drop it.
- `dispatch.select` is #172's `IMPLEMENTED_IN_DEFERRED_FAMILY` carve-out.
- `every_vendored_charge_point_is_named_in_table_order` passes, so every other deferred point is still in the table.

### Steps 2, 3 and 6: not introduced by this PR
The reader accepts the `expression`/`function` node tags but never inspects `body` or recomputes node keys. That means:
- It cannot raise graph-shape (step 2) or stale-node-key (step 3) for application nodes.
- It consumes no operation vectors, so step 6 cannot arise.

Neither `NODE_TAGS` nor body handling changed in this range. This is the ApplicationNode gap already listed as out of scope in round 1. It belongs in a follow-up ticket, not this PR.

### Gates
- Scoped tests pass: library_resolution 13, exact_decimals 25, dispatch_calls 20, composite_values 17, text_enum_identity 19, model_conformance 20, model_population 34, model_normalization 37, model_reference_queries 36, complete_value_lock 3.
- `make revendor-check`: native-v1 69/0/0, complete-value 64/0/0.

No blocking findings. No new Low or Nit findings.

PR #196, opened 2026-09-19
#131 slice 1: align model vocabulary and population shape with QSpec

kreneskyp:
## Summary

Slice 1 of #131 ("Implement model intake from domain packages"): aligns QSL's model vocabulary and the population declaration shape with the QSpec text vendored at `resources/complete-value` (QSpec commit d227270, PR #179). No intake wiring in this slice — this only renames existing types/fields/strings in place and reshapes the population declaration; `crate::model::domain_package::DomainPackage` still consumes the same producer-interface-1.3.0 caller-constructed input it always has.

## What the code does now

- `DomainPackageRecord::Population(PopulationRecord)` is a record variant. `PopulationRecord` is `{key, member_types, extent}`: it lists the population's declared member types and carries one population-level `Extent` (`Closed`/`Open`). A member type carries no multiplicity of its own; the declared maximum `N` stays on the binding (`admit_binding`'s own `declared_maximum` parameter, unchanged). `normalize`'s phase-1 reference check refuses `missing_declaration`/`missing-name` when a population names a member type that is not a declared object type (`model-complete.md`'s "Populations" row).
- `admit_binding` (`crate::model::population`) now takes a `population_key: &DeclarationKey` parameter, not a caller-supplied `&PopulationRecord`. It resolves the record from `domain_package.records` itself and refuses `foreign_reference`/`foreign-model-selection` when no `Population` record there carries that key (FR-153: "Its declaration key must belong to the binding's ModelSelection"). Object closure holds exactly when the resolved record's `extent == Extent::Closed`; otherwise it returns `AdmissionOutcome::UnknownClosure` with `Code::IncompletePopulation`/`ModelRefusalCause::IncompleteScope`, same as before. `PopulationDocument` no longer carries a `closed_world: bool` field — that runtime document only ever described members now; extent is a static property of the domain package's own population declaration, not the runtime document, and never a value the caller states independently of it.
- `InvocationContext` gained a `population: &DeclarationKey` field (the population's own declaration key, resolved the same way), threaded through both of `admit_invocation`'s `admit_binding` calls.
- Every exhaustive match over `DomainPackageRecord` (`normalize.rs`, `conformance.rs`, `dispatch.rs`, `systems.rs`) has a `Population` arm.
[...]

PR #197, opened 2026-09-19
#131 slice 2: reshape DeclarationKey and flatten DomainPackageRef

kreneskyp:
## Summary

Part of #131.

- `DeclarationKey` reshaped: removed `authority`/`revision`/`digest`, added `package`/`node`/`digest_domain`. `to_json()` emits exactly the schema's preimage.
- `DomainPackageRef` adopts the flat FR-321 shape `{identity, version, digest_domain, digest}`.
- Every hardcoded digest retargeted to vendored vectors (`model-effective-declaration-vectors.json`) where a vector exists, asserted against the vector file rather than a pasted literal.
- Merged `origin/main` (#196, now at `f8bba18`) into this branch: resolved conflicts from #196's `declaration_records` rename, `admit_binding`/`InvocationContext` resolving `PopulationRecord` by key from `domain_package.records`, and the new `UnknownPopulationMemberType` refusal cause.

## Findings to report (not fixed in this PR — flagged per the original ruling)

- **Revision-disambiguation tests deleted as structurally incompatible** with the reshape (a `DeclarationKey` no longer carries `revision`, so these tests no longer have a subject). 9 test functions + 1 now-dead helper, all only on the origin/main (#196) side at merge time:
  - `tests/model_conformance.rs` (pre-merge blob `f8bba18`) lines 36, 108
  - `tests/model_systems.rs` line 45
  - `tests/model_dispatch.rs` line 124
  - `tests/model_normalization.rs` lines 702, 743, 784
  - `tests/model_population.rs` lines 22 (helper), 238, 511
- `src/model/normalize.rs`'s `RULE_INHERIT` fact-input construction does not match the vendored FR-150 vectors for inherited declarations (pre-existing finding, documented in `n01_normalizes_f1_to_the_exact_ground_truth_identities`'s own doc comment in `tests/model_normalization.rs`); several assertions in that file are pinned to the implementation's current output rather than true vendored-vector ground truth for that reason.

## Test plan

- [x] `cargo test --all-features` (full workspace): all pass
- [x] `cargo fmt --all` applied, clean
- [x] `cargo clippy --workspace --all-targets --all-features -- -D warnings`: clean
- [x] `make ci`: clean
- [x] `make revendor-check`: clean


kreneskyp:
## Review of #197 (head f293f9a): FIX

Scope: `git diff origin/main...f293f9a` (main f8bba18). Code review, Rust review (`rust-review` checklist, repo conventions first), and gap analysis against the four ACs.

**Gates on f293f9a** (own worktree, own `CARGO_TARGET_DIR`): `make ci` passed (exit 0). `make revendor-check` passed: native-v1 69/0/0, complete-value 64/0/0.

**What is right**
- AC1: `DeclarationKey` is `{package, node}`, and `to_json` emits `{package, node, digest_domain: "sha256-jcs"}`. That is exactly `$defs.DeclarationKey` in the vendored schema (line 16). `n01-type-A` and `n01-member-A-x` reproduce the vendored digests, which proves the preimage.
- AC2: `DomainPackageRef` is the flat FR-321 shape `{identity, version, digest_domain, digest}` (model-complete.md:50). `n01-universe` reproduces `91320bde`, which proves the selection preimage. TC-195 N01's type `A` is `3b79bb92`.
- The merge with #196 is intact. `admit_binding` still resolves the population by its full key (population.rs:646-667). `UnknownPopulationMemberType` is still in place (normalize.rs:937) and tested (model_normalization.rs:772).
- No change to Cargo.toml or Cargo.lock, so no quire-rs pin change. No CI workflow change.
- The `l05` "different bundle revision" test was retargeted to a version-only difference, not deleted. Correct.
- The deletion of `n05` and `unsupported_interface_version` is correct. FR-321 has no `contract_version`. The digest domain is a constant that the typed `DomainPackageRef` cannot misstate. The byte-level N04/N05 (FR-154 admission table, TC-195 N04/N05 "admit package bytes") belongs to the #131 intake half.

### Findings

**1. High: `RULE_INHERIT` and `RULE_REDEFINE` fact inputs cite record keys, not path types.** Files: `src/model/normalize.rs:589`, `:1007-1008`, `:1065-1066`, `:1750-1752`. Tests pinned to the wrong output: `tests/model_normalization.rs:406-421` and `:424-555`, plus `:961-1063`, `:1244-1258`, `:1306-1336`, `:2408-2420`, `:2601-2615`.

The vendored rule text says:
- model-complete.md:269: *"For each type `T` and each supertype path `T -> P1 -> ... -> A`: one fact on `T` with inputs `[P1, ..., A]`; then, for each member `m` whose owner is `A`, one fact on `(T, m)` with inputs `[P1, ..., A, m]`."*
- model-complete.md:271 and :274: redefine inputs are `[path types..., r, t]`, where "`path types` is the list of supertypes from `T` to the owner, excluding `T`."

`ancestor_paths` pushes the traversed `SupertypeRecord.key` (`new_path.push(record.key.clone())`, line 589), not `record.general`. So:

| Fact | This PR emits | Vendored vectors and TC-195 require |
| --- | --- | --- |
| Type fact for `B` | `[model.gen.B-A, A]` | `[A]` (`n01-type-B`) |
| Member fact for `(B, A/x)` | `[model.gen.B-A, A/x]` | `[A, A/x]` (`n01-member-B-x`) |
| Redefine facts | `[model.gen.D-B, model.redef.B, A/x]` | `[B, B/x2, A/x]` (TC-195 N06) |
| `D`'s own redefine fact | `[model.redef.D, A/x]` | `[D/x4, A/x]` (TC-195 N06) |

Line 1751 pushes `edge.record_key` where it should push `edge.redefining`.

The byte arithmetic confirms the diagnosis. `B` is 641 bytes against the vector's 563: the difference is 78, one extra key. `(B, A/x)` is 578 against 583: the difference is -5, because `model.gen.B-A` is 5 bytes shorter than `ix://test/orders/A`.

Failure scenario: every declaration reached through inheritance or redefinition gets a non-spec identity. That covers `B`, `C`, `D`, every inherited or redefined member, every view, and every universe except `A`. The PR's own n01 and n02 tests then pin those wrong identities (`c111a68c`, `5e63ca48`, `e50ac471`, `3e8f2930`, 5457 and 13709 hashed bytes). As written they can only go red when the code is fixed.

This is not "pre-existing and independent". This PR is the vector-alignment slice, and it wrote the assertions.

Verified fix. I changed line 589 to push `record.general` and dropped the extra `ancestor_key` push at 1008. The result:
- F1 reproduces all four vendored digests: `3b79bb92`, `750ec6cf`, `9c1ce45c`, `e8e6044d`.
- F1 reproduces view `2cd5d1e7` and 5311 hashed bytes, exactly TC-195 N01.
- F2 reproduces view `f9fbcf16` and 12853 hashed bytes, exactly TC-195 N02.

The redefine side also needs line 1751 changed to `edge.redefining`, with `edge.path` carrying types. The same change fixes the conflict-refusal path listing, which TC-195 N06 gives as `[B, B/x2, A/x]` and `[C, C/x3, A/x]`. Then re-point every pinned assertion at the vectors (finding 2) and delete the "Known gap" doc blocks.

**2. High: no test reads `model-effective-declaration-vectors.json` (AC3).** Files: `tests/model_normalization.rs:424-555`, `:558-561`, `:1244-1291`.

Every digest is a pasted literal. The only mention of the vectors file is a comment. The repo already has the pattern for reading vectors (`tests/text_enum_identity.rs:379`, `tests/quantities.rs:59`).

Failure scenario: a revendor that changes a vector, or a code change that happens to match the pasted literal, keeps these tests green. This happened here: the n01 and n02 tests pass against wrong values.

Fix:
- Load the vendored file and assert each of the 18 named vectors (`n01-*`, `n02-*`, `n01v2-*`) by name.
- Also assert each vector's `preimage` equals the computed preimage's JSON, not only the digest.
- The two `n01v2` vectors (view `e6f8a8cd`, universe `3526be8a`, TC-195 N09 third clause) are untested today. They are the direct test of the behaviour this PR introduces: a version-only change keeps declarations and changes the view and universe (FR-154-AC-4).

`DomainPackageRef::fixture` (domain_package.rs:425) documents "F1 as version `2` of `n01v2`" but cannot produce it, because version is hard-wired to `"1"`. It needs a version parameter. The same applies to `fixture_f1_with_version` (model_population.rs:196), which keeps the `n01` digest.

**3. Medium: deleted `f2` "both survive" tests (normalization ×2, conformance, systems, dispatch). The behaviour still exists as FR-154's `conflicting-binding`, and the code now drops records silently.**

With `revision` gone, the old two-revisions-of-one-name fixture becomes two records with one `DeclarationKey`. QSpec still has a rule for that. FR-154 Declaration refusals: *"Two nodes share one identity … | `invalid_model_binding`/`conflicting-binding`"*. model-complete.md: *"Two different original declarations never collapse"*.

`normalize` already implements other rows of that same FR-154 table in `validate_references` (the `missing-name`-class refusals and `UnknownPopulationMemberType`), so this check belongs there too.

Verified on f293f9a. `[A, FieldMember A/x, FieldMember A/x (dup), A (dup)]` returns `Completed` with 2 declarations after charging 4 `normalize.record` and 3 facts. That is exactly the "silent drop while the meter charged both" defect the deleted `f2` tests guarded (PR #140 F2).

Fix: refuse a duplicate `DeclarationKey` across records with `InvalidModelBinding`/`conflicting-binding`. Retarget the five deleted tests to assert that whole refusal, one per index: normalize, `ConformanceIndex`, `SystemsClassification`, `DispatchIndex`.

**4. Medium: deleted `r10` (model_conformance) and `enforce_frame_refuses_a_field_write_at_a_revision…` (model_population). Full-key frame equality is still required and is now untested.**

`DeclarationKey` still has two components, and keys "compare by `package`, then `node`" (model-complete.md). FR-321-AC-3 says a key is fixed by package and node. A write to `{package: other/orders, node: model.A.x}` is not covered by a grant for `{test/orders, model.A.x}`.

Verified. I mutated both call sites to compare `.node` only: `population.rs:1169` (`field_write_covered`) and `conformance.rs:681` (effect axis). The full `cargo test --all-features` suite stays green. The comment at `population.rs:1121-1128` still says this is "pinned by" the two deleted tests.

Fix: restore both tests with the write key's `package` changed instead of `revision`. Assert the same whole `FrameFieldWriteOutsideGrant` / `EffectEscape` outcomes. Update that comment.

**5. Medium: deleted `n04_absent_revision_refuses_wrong_model_selection`. Its new form is empty `package`, `node`, `identity` or `version`.**

The vendored schema types `DeclarationKey.package`, `.node`, `DomainPackageRef.identity` and `.version` as `Nonempty`. model-complete.md says every identity is the hash of a *schema-valid* preimage. FR-321 says: *"Missing required properties … out-of-domain values … refuse before consumption."*

Failure scenario: `DeclarationKey { package: "", node: "" }` or a selection with `version: ""` now normalizes and hashes a schema-invalid preimage into an identity.

Fix: restore the phase-1 check as an empty-component refusal, and retarget n04 to it. The byte-level N04 (`wrong-model-selection` over package bytes) stays with the #131 intake, and the deletion comment should say so rather than read "Real N04 …".

**6. Medium: the F1 fixture still charges a separate `Supertype` record, so the N01 charge totals cannot match TC-195.** Files: `tests/model_normalization.rs:567-622`, `:1260-1291`.

TC-195 N01 specifies `declaration_records: 3` and 19 work units ("three `normalize.record` (`A`, `A/x`, `B`)"). `B.supertypes = [A]` is a property of the node, not an IR node. The fixture's `model.gen.B-A` record is charged, giving 4 records and 20 work units, and the tests pin 4 and 20.

After the finding 1 fix, hashed bytes match (5311), but work units stay 20. Either stop charging generalization and redefinition records as declaration records in this PR, or open a named follow-up under #131 and state the exact TC-195 N01 divergence in the test instead of calling it "ground truth".

**7. Low: comments describe code that no longer exists or history.**
- `src/model/population.rs:1121-1128` names the fields `authority, identity, revision, digest` and two deleted tests.
- `src/model/population.rs:415` and `:531` say "export identity".
- `src/model/normalize.rs:400-403` and `:724-726` say "differ in revision or digest".
- `src/model/mod.rs:9` and `src/model/checked_dispatch.rs:22` say "producer interface `1.3.0`". The constant is deleted and QSpec has no such version.
- `Cargo.toml:55` names `ProducerKey::fixture`, `ModelSelection::fixture` and `ProducerDigest::of_identity`, none of which exist.
- The `normalize` doc (normalize.rs, above `pub fn normalize`) says the FR-154 byte checks "run before a caller builds one". No such intake exists in the crate, so that is false today.
- History narration in tests: `tests/model_dispatch.rs:373`, `tests/model_reference_queries.rs:622` and `:1458`, `tests/model_population.rs:336`, and `tests/model_normalization.rs:1254-1258` and `:1311-1313` ("#131's reshape changed…", "down from PR #140's original 27969"). These should state the current rule, or cite the vector.

**8. Nit: `DomainPackageRef` derives `Debug` over `digest: [u8; 32]`.** File: `src/model/domain_package.rs:414-428`. Refusal payloads (`ForeignModelSelection { expected }`) now print 32 decimal bytes. The deleted `ProducerDigest` had a hex `Debug`, which made failure output readable. Restore a hex `Debug`.

**9. Low: add a digest-only case to the l05 correspondence test.** File: `tests/model_population.rs:1088`. FR-321 says: *"Two selections with equal identity and version and different digests are unequal."* The version-only case is covered; the digest-only case is not.

**Out of this PR (for the record).** QSpec 80d8e55 bumped the model rules to `1-draft.2` and re-issued the vectors. For example, `n01-view` becomes `7281ed1f`, yet TC-195 at 80d8e55 still quotes `2cd5d1e7`. This PR correctly targets the vendored d227270 set. The revendor is separate work.

Verdict: **FIX** (findings 1–7).

kreneskyp:
## Round-2 delta review: QSL #197 at 590e721 (delta f293f9a..590e721)

**Verdict: FIX.** Two blocking findings, F1 and F2, both in code this round added or changed.

Round-1 findings: 1, 2, 4, 8 and 9 are closed. 3, 5 and 7 are mostly closed. 6 is open.

### Round-1 findings

| # | Status |
|---|---|
| 1 RULE_INHERIT / RULE_REDEFINE fact inputs | **Closed.** `ancestor_paths` pushes `record.general`. Type inherit inputs are `ancestor.path`. Redefine inputs push `edge.redefining`. This matches model-complete.md:269/271/274 and my independently verified values: A 3b79bb92, B 9c1ce45c, A/x 750ec6cf, (B, A/x) e8e6044d, n01 view 2cd5d1e7 / universe 91320bde, n02 f9fbcf16 / 749e472d, n01v2 e6f8a8cd / 3526be8a. N06 inputs are now `[ix://test/orders/B, model.B.x2, A/x]` and `[model.D.x4, A/x]`. |
| 2 vector tests | **Closed.** `vectors()` reads the vendored JSON via `CARGO_MANIFEST_DIR`. `assert_declaration_matches_vector` checks the digest **and** the full preimage against an independent reconstruction. n01, n02, n07 (n02-view) and n01v2 all read the file, and `fixture_with_version` exists. Nit: n01v2's "declarations reused" assert compares n01-view against n01v2-view inside the vendored file only, so it tests the fixture, not the code. The computed view and universe asserts are the real check. Not blocking. |
| 3 silent drop of duplicate keys | **Behaviour closed; order wrong (F2).** `ConflictingBinding` refuses. The deleted f2 tests are retargeted in normalization, conformance, dispatch and systems, and each asserts code and cause. |
| 4 full-key frame equality | **Closed.** The package-differing retargets (`enforce_frame_refuses_a_field_write_at_a_package_…` and `r10_operation_redefinition_effect_axis_refuses_a_write_at_a_package_…`) assert whole outcomes and document their mutations. The population.rs:1118-1128 comment is updated. |
| 5 empty-component refusal | **Mostly closed.** Reusing `MalformedDeclaration` is right: it is FR-154 row 1, malformed id. The remaining test gap is F2(c). |
| 6 N01 charge count | **Open. See F1.** |
| 7 stale comments | **Mostly closed.** mod.rs, checked_dispatch.rs, Cargo.toml and "revision or digest" are fixed. Left, non-blocking: population.rs:415 and :531 still say "export identity". The new test comments still narrate history ("#131's reshape and the #197 … fix each changed…") instead of citing the vector. |
| 8 `DomainPackageRef` Debug | **Closed.** Manual `Debug` with a hex digest. |
| 9 l05 digest-only case | **Closed.** `l05_view_from_a_domain_package_with_a_different_digest_refuses_at_admission`, which also asserts zero admitted charges. |

### F1 (blocking): `declaration_record_count` fixes the N01 count by filtering, not by fixing the record shape, and it still overcharges

`charge_all` now charges one `normalize.record` for every record except `Supertype` and `Redefinition`.

What QSpec says:
- **value-accounting.md:489.** One `normalize.record` per "declaration record that intake builds from a domain package IR **node** (object type, field, relationship, operation, clause, population, part, port, interface, connection or allocation)".
- **model-complete.md:155/159/160 and :270/271.** Supertypes, subsets and redefines are **properties of a node** (`B.supertypes = [A]`, `B/x2.redefines = A/x`), not nodes. TC-195 F1 is written in that shape.

Answers to the two questions:
- **Are supertypes inline on the object type?** Yes. They are not separate declaration records. The real problem was the fixture shape: `supertype("model.gen.B-A", …)` is its own record with its own key.
- **Are redefinitions charged as records?** No. A redefinition is a property of the redefining member.

So excluding these records gives the right *number* for N01. But it makes the count fit TC-195 without fixing the shape. Two defects remain:

1. **`Subsetting` is also a property (model-complete.md:159/270), and it is still charged.** Any package with a subsetting record overcharges `normalize.record` and `declaration-records` against value-accounting.md:489, by one per subset. No test covers this.
2. **`Supertype`, `Redefinition` and `Subsetting` records keep their own non-node `DeclarationKey`s** (`model.gen.B-A`, `model.redef.B.op-A.op`). Those keys go through the new empty-component and conflicting-binding checks and the refusal ranking. Intake can therefore refuse on an identity for which QSpec has no node.

**Correct fix:** move `supertypes` onto `ObjectTypeRecord`, and move `redefines` and `subsets` onto the field and operation member records, with no keys of their own. Then `normalize.record` equals `records.len()` and the filter goes away.

If the reshape is split out to a named follow-up under #131, the minimum for this PR is:
- add `Subsetting` to the exclusion;
- add a test that charges a package with one subsetting record: 3 nodes plus 1 subset gives 3 `normalize.record` charges.

### F2 (blocking): `ConflictingBinding` runs in the wrong order

`validate_references` checks for duplicate keys across the whole package before any missing-name check.

native-diagnostics.md:108-122 is the refusal order for a checked-package/v2 **reader**, not for intake, so it does not govern here. **FR-154** does:
- "Intake reports every declaration refusal in node order … Within one node, every check … in table order."
- Table order: malformed id, kind/construct, invalid node, **missing-name (row 4)**, declaration-form, **conflicting-binding (row 6)**, multiplicity.

Failure scenarios:
- **Across nodes.** `[field A.y whose owner does not resolve, type B, type B]`: QSpec reports missing-name on A.y, which comes first in node order. The PR reports conflicting-binding on B.
- **Within one node.** A duplicated member whose owner is dangling: QSpec reports missing-name (row 4) first. The PR reports conflicting-binding (row 6).

Fixes:
- (a) Fold the duplicate check into the per-node pass, so refusals come out in node order and, within a node, in table order.
- (b) Add one ordering test for each scenario above.
- (c) Add empty-`version` and empty-`package` tests. There are none today, and the two existing empty-component tests assert code and cause only, not the whole outcome.

Pre-existing, not blocking this PR: `build` runs `validate_references` before `charge_all`. Every declaration refusal, including the new ones, therefore carries zero `normalize.record` charges. TC-195 N08 has refusals "after three `normalize.record` charges". This is worth its own ticket.

### Gates (own worktree, explicit `CARGO_TARGET_DIR`)
- `make ci`: exit 0.
- `make revendor-check`: exit 0. native-v1: 69 matched, 0 drifted, 0 stray. complete-value: 64 matched, 0 drifted, 0 stray.

Green gates do not cover F1 or F2. No test runs a subsetting record under charge limits, and no test checks refusal order.

PR #253, opened 2026-09-20
QSL-57 items 1+3: wire TypeEnvironment's conformance graph into equality upcast and lookup check-time typing

kreneskyp:
## Summary

#164 (QSL-57) was filed off PR #158 (2026-09-18), whose comment said `TypeEnvironment` carries no generalization graph, deferring three FR-153/FR-149 obligations to evaluation. PR #204 (2026-09-19) added `TypeEnvironment::ancestors`/`conforms` the very next day, for dispatch's own reference-upcast case — neither #164 nor the stale comments were updated to reflect it. This PR lands two of #164's three items using that already-existing graph; item 3 below required a second, independent fix beyond the type-level one.

## Changes

- **`lookup<T>(p, r)` check-time conformance** (`src/value/expression/check.rs`): refuses a non-conforming `S` at check time via `self.scope.types.conforms`, the same source `check_dispatch_argument`'s upcast case already uses (#204). Evaluation's `type_conforms` backstop is unchanged, for a `NodeKind::Lookup` built without going through this checker. The two aren't fully equivalent — `conforms`'s closure is unbounded, `type_conforms` walks at most 128 steps — a documented, resource-bound divergence (checker admits what evaluation `ResourceExhausted`-refuses on a >128-deep chain), not a soundness gap.

- **Equality upcast** (`src/value/equality.rs`): `check_equality` now admits a `Reference<A>`/`Reference<B>` pair when one conforms to the other, not only when equal. **This type-level fix alone was not sufficient** — my own new test caught it before I trusted it. `operand_value` independently re-validates each operand's *runtime* value against its *declared* type via `ValueType::admits`, which does exact object-type-key matching. A `lookup<A>(...)`'s real result value carries the looked-up object's true most-specific type, not `A` (documented at `model_query.rs:233-242`), so even with the checker fix, evaluating a genuine upcast pair would still refuse `CheckedInvariant`. Fixed by skipping that structural check for `Reference` operands specifically — the identical soundness argument `OptionValue::from_admitted` already uses for `lookup`'s empty-mode path (by the time `operand_value` runs, `check_equality` has already proven the types compatible). The skip is written as an exhaustive match naming every other operand kind explicitly as structural/exact, not a wildcard: object types are the only nominal type in this crate with a declared supertype relation today, so this is a checked claim, not an assumption — if a future variant gains one, this stops compiling here instead of silently falling through the wrong arm.

- **Stale docs** (`src/value/model_query.rs`): rewrote the "TypeEnvironment island" module doc, which repeated the same stale claim. States current reality: the graph exists (#204), the checker uses it directly for both fixes above, production still feeds an empty `TypeEnvironment` (`checked_dispatch.rs`'s sole `PackageDeclarations` builder leaves `types` at `Default`, no intake bridge exists yet), and the 128-step divergence.

- **Fixture bug found and fixed** (`tests/model_reference_queries.rs`): the shared `types(scenario)` fixture built checker-level `M::A`/`M::B` with no supertype edge, even though `fixture_f1`'s `DomainPackage` (used by the same `scenario()`) already declares `model.B -> model.A`. This is a live instance of the exact "TypeEnvironment island" bug class #164 describes — 4 existing tests failed the moment check-time conformance was turned on, because the checker-side fixture disagreed with the runtime-side one it's tested against. Fixed by mirroring the edge onto the fixture. Worth stating plainly rather than burying as a fixture touch-up: this is the check working, not a test being wrong.

## Not in this PR

Item 2 (`deref(r).f` inherited-field resolution) stays open on #164. It's a storage-model gap, not a lookup gap: `ObjectEnvironment`'s slots are sized to an object's own declared type's `attributes()` alone, so a field only an ancestor declares has no storage slot on a subtype's instance at all today — nowhere to look it up, not just nowhere currently looked. It needs a ruling (flatten inherited attributes at declaration time vs. per-ancestor storage) before implementation; that discussion and a real spec-backed complication (field redefinition, FR-151) are recorded as comments on QSL-57/#164.

Remaining work: #164 (item 2)

## Test plan

- [x] `cargo fmt --all -- --check`
- [x] `cargo clippy --locked --workspace --all-targets -- -D warnings`
- [x] New tests: `e16a_references_admit_a_conforming_upcast_either_direction`, `e16b_references_refuse_an_unrelated_object_type` (`tests/equality_matrix.rs`); `lookup_expression_admits_a_conforming_reference_type_at_check_time`, `lookup_expression_refuses_an_unrelated_reference_type_at_check_time` (`tests/model_reference_queries.rs`)
- [x] All 4 touched test files green (142 tests)
- [x] `cargo test --features test-support` (full workspace)
- [x] `make ci` (clippy + tests, default and all-features; clean build; fixture-audit self-test; parse example)


kreneskyp:
Closing this unmerged. The work is still wanted; this branch cannot carry it.

**It is superseded by the M-5 split, not merely stale.** PR #282 (QSL-139, ADR-011 M-5) renames the primary file this PR edits:

```
R098    src/value/expression/check.rs -> src/check/check.rs
```

A 98%-similarity rename. This branch's changes to `src/value/expression/check.rs` target a path that no longer exists after M-5, so there is no clean merge available — the diff would have to be re-authored against `src/check/check.rs` regardless of what we do here. It also touches `src/value/equality.rs`, which the same split reorganises.

**Measured state:** 1 commit ahead of `origin/main`, **31 behind**. GitHub reports `mergeable: UNKNOWN`. The branch predates #261 (`836dc12`) and #263 (`59d885da`), both of which landed on the code it modifies, and predates tonight's history rewrite.

**Sequencing.** Issue 164's own position is that it was deliberately not started because it collides with #131's wiring and X-1. That collision is now concrete and visible rather than anticipated: M-5 is relocating exactly these files, and M-2 (QSL-7) moves `model::checked_dispatch` and `model::conformance::check_field_refinement_obligation` into `check` after it. Re-authoring `TypeEnvironment`'s conformance-graph wiring before those land means writing it twice.

**What survives:** nothing is lost by closing — the branch stays in the repository and the diff is readable from this PR. Issue 164 remains **open** with its scope intact. The work should be redone against `src/check/` once M-5 and M-2 have landed, at which point the right home for it is unambiguous.

Generated with [Claude Code](https://claude.com/claude-code)

Issue #213, opened 2026-09-19
[QSL#213] [ARCH-20] Implement checked identity, typestate, Capability/outcome primitives, provenance, and bounds


kreneskyp:
Parent: [#205](<https://linear.app/agent-ix/issue/AGE-374/qsl205-epicarch-establish-the-qsl-v1-semantic-and-execution>). Layer: 2 — foundational implementation. Depends on the accepted Layer 1 gate [#212](<https://linear.app/agent-ix/issue/AGE-381/qsl212-arch-g1-pass-the-whole-system-architecture-change-scenario-gate>) and the boundedness decisions in [#222](<https://linear.app/agent-ix/issue/AGE-391/qsl222-arch-42-accept-temporal-trace-and-boundedness-architecture>). Consumes the capability specification from [#229](<https://linear.app/agent-ix/issue/AGE-397/qsl229-arch-13-align-the-qsl-capability-specification-with-fr-290>). It implements the canonical Rust `Capability` value type and shared outcome primitives; [#185](<https://linear.app/agent-ix/issue/AGE-362/v1-a08-lowering-target-registry-and-routing-over-the-canonical>) depends on this ticket and alone owns backend registration/routing.

## Goal

Implement the minimum shared Rust types required by the accepted architecture for stable semantic identity, checked typestate, structured outcomes/refusals, provenance, proof/execution modes, and bounds.

## Scope

* Consolidate identities and provenance selected by [#211](<https://linear.app/agent-ix/issue/AGE-380/qsl211-arch-12-assign-canonical-type-package-and-conversion-ownership>); eliminate public display-name or collection-position identity.
* Make unchecked, checked, linked, and packaged values distinct where [#209](<https://linear.app/agent-ix/issue/AGE-378/qsl209-arch-10-accept-the-qsl-stage-dag-and-cratemodule-dependency>) requires stage separation.
* Provide structured success, violation, refusal, unsupported, timeout, cancellation, incomplete, bound-exhausted, and internal-failure outcomes without collapsing them into strings or booleans.
* Implement typed finite/bounded/unbounded request and resource-bound representations selected by [#222](<https://linear.app/agent-ix/issue/AGE-391/qsl222-arch-42-accept-temporal-trace-and-boundedness-architecture>).
[...]

PR #271, opened 2026-09-20
ADR-011: correct M-2's order — it depends on M-5, not the reverse (QSL-7)

kreneskyp:
## Summary

QSL-7 (ADR-011 T-7, M-2: move `model` below `check`, create `semantic_value`) was reported startable. Reading the `use` lines and consumer graphs before moving anything found that M-2 cannot land in the order ADR-011 §7.3 states, for reasons the ADR's own tables don't show. This PR is spec-only: it corrects ADR-011, and does not touch any Rust source. QSL-7 stays blocked after this lands.

**Real order: X-1 → QSL-131 → M-5 (QSL-139) → M-2 (QSL-7).** M-3a already landed ahead of M-2 in the stated sequence, correctly and without conflict — its scope is disjoint from M-2's. **QSL-131 is the head of this critical path and is unstaffed.**

## What changed in ADR-011

1. **§7.3's order preamble.** Was: X-1, M-2, M-3a, M-5 land "one at a time" because they "all edit `value` and `model`". That reason is wrong — M-3a's scope is disjoint from M-2's `value` submodules, so the edit surface never collided. Corrected to state the real constraint (a dependency, not a collision) and the real order.
2. **§7.3's M-2 row.** Order cell: `after X-1` → `after M-5 (QSL-139)`. Scope split into its three independently-blocked parts, naming QSL-165 for `semantic_value`.
3. **§7.3's M-5 row.** Order cell dropped "after M-2" (now backwards/circular given #2) — left as "after M-3a and QSL-131".
4. **§6.2's `model::checked_dispatch` row.** Owning change `M-2` → `M-5` (covers both `model::checked_dispatch` and `model::conformance::check_field_refinement_obligation`, which the row already lists together).
5. **§6.1's "K is a leaf" bullet.** Added that X-1 (#213 S-1, PR #254) did not make the cuts this bullet requires — records that the requirement was not met, per PR #254's own summary ("No file under `src/value/` or `src/model/` is touched here") — and names QSL-131 as owner of the outstanding cuts.

### M-2 part (a): `model::checked_dispatch` needs M-5

`src/model/checked_dispatch.rs:109-112`:
```rust
use crate::value::{
    BinaryOperator, DeclaredClauseKind, DispatchCandidate, DispatchOperation, DispatchTable,
    Expression, FieldInitializer, FunctionDeclaration, NodeKey, PackageDeclarations, ValueType,
};
```
`BinaryOperator`/`DeclaredClauseKind`/`Expression`/`FieldInitializer`/`FunctionDeclaration` resolve to `crate::forms` (layer 2, moved there by M-3a) — fine, layer 3 may depend on layer 2. But `DispatchTable`/`DispatchCandidate` are defined in `src/value/expression/ir.rs`, and `DispatchOperation`/`PackageDeclarations` in `src/value/expression/check.rs` — both still layer 5 today. ADR-011 §6.2 assigns their owning submodules (`value::expression::{check,facts,termination,ir}`) to layer-3 `check`, but that move is **M-5's** scope. Moving `checked_dispatch` into `check` now creates `check → value::expression` (layer 3 → 5), forbidden by §6.1's exhaustive allow-list ("2, F, K").

### M-2 part (b): `check_field_refinement_obligation` needs M-5, through a different route

Its exclusive helper chain (nothing in this chain is called by any function that stays in `model`) is `clause_location` → `self_field_node`/`field_domain_type` → `presence_condition`/`comparison_condition` → `established_facts` → `check_field_refinement_obligation`, plus `format_interval` (`src/model/conformance.rs:759-1085`). It is built directly out of `value::expression` primitives:

- `Location`, `Origin` (line 759) — `value::expression::refusal`
- `Node`, `NodeKind`, `Connective`, `OrderedKind` — `value::expression` (`ir.rs`), re-exported `pub(crate)` at `src/value/mod.rs:131-133`
- `established_field_fact` (line 891) — `value::expression::facts`, the FR-146 fact-derivation primitive the function exists to call
- `ProvedInterval` (line 896) — `value::expression::refusal`

Same four M-5-owned submodules as part (a) (`check`, `facts`, `ir`, `refusal`), same forbidden edge if moved now.

Separately: this function shares private machinery (`ConformanceIndex`, `AxisFailure`, `ConformanceOutcome`, `missing_member`) with functions that stay in `model` (`ConformanceIndex::build` is also called at `conformance.rs:390,469,549,1106`, not just at `917`). Once M-5 lands and this part is unblocked, the move needs those four widened to `pub(crate)` individually verified (revert one at a time, name the compiler error and call site each one is for), not a single cascading revert — moving the function is not a pure cut-and-paste.

### M-2 part (c): `semantic_value` needs QSL-131 (tracked as QSL-165)

All six non-kernel `value` submodules except `unit` are imported back by K-designated `value` siblings that stay put pending QSL-131:

| K-designated module (stays in `value`, §6.2 row 712) | imports (line) |
|---|---|
| `division.rs:8` | `definition::AdmittedIntegerDivision` |
| `ieee.rs:24` | `definition::{...}` |
| `equality.rs:23,25` | `enumeration::compare_enum`, `quantity::{...}` |
| `composite.rs:22,26,28` | `enumeration::EnumValue`, `quantity::{Quantity, QuantityUnit}`, `reference::ObjectReference` |
| `collection.rs:24` | `key::compare_keys` |
| `outcome.rs:12` | `reference::ObjectReference` |

`unit` is clean (its one sibling importer, `quantity`, moves with it). §6.1's "K is a leaf" bullet already names this exact edge set: "`collection`, `equality`, `division` and `ieee` import `key`, `enumeration`, `quantity` and `definition`. X-1 must cut every one of these edges... K depends on nothing." X-1 did not cut them (PR #254: "No file under `src/value/`... is touched here").

This is not the same shape as `semantic_value` depending on not-yet-relocated K submodules (forward direction, self-heals when QSL-131 lands, since the direction is already correct). This is K reaching *up* into layer 3, forbidden in every version of the target — and the eventual fix is a **different** edge, not this one relocated: `quire-exact`'s own `composite`/`collection`/`equality`/`division`/`ieee`/`outcome` will use `quire-exact`'s own already-built narrow `quantity.rs`/`reference.rs`/`key.rs` (present today under `quire-exact/src/`), never `semantic_value`'s full declaration types.

## Not in scope here

No Rust source changed. `spec/reviews/` untouched (dated snapshots). `semantic_value`, the `checked_dispatch`/`check_field_refinement_obligation` moves, and QSL-131's edge cuts are all separate, blocked tickets (QSL-165, M-2b, QSL-131) — this PR only corrects the ADR that governs their order.

## Gates run

- `quire validate spec/decisions/ADR-011-stage-dag-and-dependency-architecture.md` — exit 0 (pre-existing module-registration warnings unrelated to this edit; same warnings independent of content).
- `cargo fmt --check` — exit 0 (no Rust files changed).


kreneskyp:
Pushed a fix-up commit (\`c69037c\`) addressing the REQUEST CHANGES review. Six findings, three of them errors in the coordinator's own prior instructions rather than in the reading behind them:

1. **Ownership reverted, M-2 not M-5.** \`model::checked_dispatch\`/\`check_field_refinement_obligation\` are owned by M-2 (moving \`model\` code is M-2's job), blocked by M-5 — not owned by M-5. This was wrong in three places (§6.1's bullet, §6.2's module-map row, §7.3's M-2 row); all three now say the same thing.
2. **Fixed a real cycle.** M-3a's row said "after M-2"; combined with M-2's new "after M-5", that closes M-3a → M-2 → M-5 → M-3a. Changed to "after X-1", M-3a's actual dependency.
3. **Dropped an unsupported edge.** "QSL-131" was on M-5's Order cell on the assumption M-5 needs \`semantic_value\` to exist. Checked \`value/expression/{check,facts,ir,refusal,termination}.rs\`'s imports: they pull only K-designated \`value\` siblings, plus, in \`check.rs\`'s case (\`check.rs:15,20\`), \`enumeration\`/\`quantity\` — in the *allowed* direction (layer-3 \`check\` depending on layer-3 \`semantic_value\`, later depending on earlier in the layer's order). Nothing forces \`semantic_value\` to exist first. M-5's row now reads "QSL-139, after M-3a", full stop.
4. **\`semantic_value\` split into its own row (QSL-165)**, matching how QSL-146 already gets its own row next to X-1, instead of being part (c) of a three-part M-2 row.
5. **Fixed the X-1 row's tense** — it asserted X-1 makes the kernel leaf cuts; §6.1's bullet already says it didn't.
6. **Dropped the QSL-163 citation** (different failure — test-traceability gap, not the uncut K edges) and trimmed the sequencing-ruling history out of the ADR text; that narrative lives in this PR's description and the Linear ticket.

Net effect on the stated order: **X-1 → M-3a → M-5 (QSL-139) → M-2 (QSL-7)**, and independently, **QSL-131 → QSL-165**. No remaining cycle among M-2/M-3a/M-5/X-1 — checked every mention of all four in the document (74 occurrences, 48 lines) against "after"/"before"/"blocked"/"depend".

\`quire validate\` and \`cargo fmt --check\` both still exit 0 (same pre-existing module-registration warnings, unrelated to content).

PR #284, opened 2026-09-21
Stop vendoring quire-specification resources into QSL (PLAT-887)

kreneskyp:
**DRAFT — not ready for review. Full gates have NOT completed on this branch.** Opened to make the work durable and its state legible, not to request review.

Removes the vendored spec trees from QSL. PLAT-887.

## Removed, and what replaced it

| removed | replaced by |
|---|---|
| `resources/complete-value/` — 58 files + Unicode 17.0.0 corpus, ~7.1 MB, from private `quire-specification` | `src/value/definition.rs`: `DefinitionLock` as Rust-native consts — 21-entry `CATALOG`, `ALWAYS_ROLES`/`CONDITIONAL_ROLES`/`EXACTLY_ONE_ROLES`. Recognition is identity+revision. |
| `resources/native-v1/` | linking/protocol_artifact layers resolve by identity, not frozen bytes |
| `xtask/` crate (`revendor`, `revendor-check`) | nothing — its only purpose was the deleted vendoring machinery |
| `NFR-011`, TC-149/150/151/152 | nothing — the requirement was "vendor resources from exact pins" |
| `.cargo/config.toml`'s `xtask` alias | nothing — it dangled at a deleted package |
| `tests/complete_value_lock.rs`'s 19 selection vectors | inlined as Rust const test data |

The per-role digest field and its equality checks (`resolve_division`, `value/ieee.rs::admit_ieee_profile`) are **dropped, not reconstituted**, and `PackageCause::ByteDigestMismatch` is removed. Those digests compared a local file against a local constant — both operands in this repo — so they could only ever detect someone editing the fixture, never upstream movement.

## Still vendored in this tree

`tests/fixtures/architecture/` (112K) and `tests/fixtures/modules/` (360K), each still carrying `VENDOR.json`, both from `agent-ix/filament-core-data`.

Reason: `lift_document` needs real files at a real filesystem path, and there is no mechanism to resolve a git dependency's fixture directory from this crate. That is an upstream gap, not a justification — the copies still go, on a follow-up ticket. Recorded here so the survivors are visible rather than quietly retained.

## Gate status — nothing rounded up

| gate | status |
|---|---|
| `cargo check --workspace --all-targets --all-features` | completed, passed except the 2 known compile failures below |
| `cargo clippy ... -D warnings` | completed, passed — zero warnings outside those 2 |
| `cargo fmt` | **targeted only**, on the 8 touched files. Never run as full-workspace `--all --check`; the workspace has pre-existing drift (e.g. `src/protocol_artifact/intake.rs`) left alone deliberately |
| `cargo test --workspace --all-features` | completed once, but **predates the last commit** — not a green run against `504c4da` |
| `make ci` | **never run on this branch** |
| second (pre-merge) full gate run | **has not happened** |

## Known red

Items 1-8 are the accepted Part B surface — 9 compilation units, not the 7 originally scoped, since two `#[cfg(test)]` modules inside `src/` were found later.

1. lib unit tests — 10 failed: `diagnostic::foundation_tests::*` (2), `value::member::tests::*` (8)
2. `tests/exact_decimals.rs` — **fails to compile** (`include_str!` of a deleted file)
3. `tests/library_resolution.rs` — **fails to compile** (same)
4. `tests/ieee_profiles.rs` — 1 failed (TC-193 fixture)
5. `tests/model_normalization.rs` — 8 failed
6. `tests/quantities.rs` — 4 failed
7. `tests/text_enum_identity.rs` — 2 failed
8. `tests/unicode_17_text.rs` — 3 failed
9. `tests/compiled_protocol_v2.rs` — 2 failed, rooted in `examples/protocol-handoff/producer.rs`, which belongs to a separate in-flight change. Not addressed here.

## The one open verification

`tests/composed_definitions.rs` had a failure caused by a stale test left from the prior session's digest-check removal in `src/linking/composed/definitions.rs`. A fix is committed in `504c4da` (`missing_or_self_inconsistent_selected_rules_refuse_only_their_closures`), but **it has never been confirmed by a passing run** — both attempts were inconclusive or interrupted.

It is not claimed green. It needs one `cargo test --test composed_definitions`. The only evidence for it today is that its logic matches an already-verified self-consistency pattern used elsewhere in the same file, which is an argument, not a result.

## Before this leaves draft

- [ ] `cargo test --test composed_definitions` passes against `504c4da`
- [ ] full gates run twice, with the red list above reconciled against what is expected
- [ ] the two `filament-core-data` fixture trees have a follow-up ticket


kreneskyp:
Do not merge this as it stands. Its own commit message ships the repo red:

> Known accepted-red (Part B, deferred, not part of this change): 9 compilation units still read the deleted resources/complete-value path at runtime or compile time

Removing a vendored tree and leaving the build broken is not a completed de-vendoring. Green is the completion test: whoever removes the copy fixes what it breaks, in the same change.

Two further reasons this branch cannot land as-is:

- It deletes `xtask/` wholesale. That was true when it was written; it is not now. `make ci` runs `cargo xtask seam-probe`, and `xtask/src/import_graph.rs` plus `definition_scan.rs` back merged tests from agent-ix/quire-spec-language#282. Only the vendoring machinery goes.
- The branch is 252 files in conflict with current `main`, having been overtaken by agent-ix/quire-spec-language#282 and agent-ix/quire-spec-language#287.

The work is being redone on current `main` on branch `plat-887-devendor`, citing `504c4da` as prior art for the decisions -- particularly its re-platforming of `DefinitionLock` off the compile-time `include_bytes!` of the vendored lock file. Scope is widened to all four vendored trees rather than one: `resources/complete-value`, `resources/native-v1`, `tests/fixtures/architecture`, `tests/fixtures/modules` (227 files, 7.4 MB). The deferral of the filament-core-data fixtures is not carried over -- if they genuinely cannot be de-vendored that is a blocker to raise, not a reason to keep the copy.

This PR will be closed as superseded once the replacement lands.

kreneskyp:
Superseded by agent-ix/quire-spec-language#288, merged as d5d49df.

That PR does all four vendored trees rather than one, and without this branch's `Known accepted-red (Part B, deferred)` — all nine compilation units that read the deleted path are re-platformed in the same change. The filament-core-data fixtures are not deferred either: `tests/model_intake.rs` resolves them through `cargo metadata`'s `manifest_path`, so the "needs a real filesystem path this crate cannot currently resolve from a git dependency" blocker did not hold.

`xtask/` is gutted rather than deleted, keeping `seam_probe`, `import_graph`, `definition_scan` and `string_edge`, which this branch predated.

Your decisions carried over and are credited as prior art in the merge commit — particularly moving `DefinitionLock` off the compile-time `include_bytes!` of the lock file, and removing `PackageCause::ByteDigestMismatch`. Thanks for those; they were the hard part.

PR #361, opened 2026-09-23
QSL-191: S6a family kind and seam dispatch (FR-090-AC-4, TC-385)

kreneskyp:
Linear: QSL-191 (FR-090-AC-4, TC-385).

## FR-090 variant-count contradiction, resolved by lead ruling (b)

FR-090 lines 43-45 define the S6a family kind as one variant per family that implements `ReferenceEvaluation`. Lines 164-165 and 473, TC-385 step 1 and ADR-012 §5.1 S1 fixed five variants. Only `Value` implements `ReferenceEvaluation`, and the spec defines no environment, `Observed` type or `FamilyContract` for the other four. Ruling (b): the definition wins. The S6a kind is the `ReferenceEvaluation` families, which today is `{Value}`. It never has a `Relation` variant, and the seam has one arm per variant. No hook is stubbed.

## Code
- `S6aFamilyKind { Value }` in the check core (`src/family/mod.rs`), with a `#[cfg(seam_probe)]` variant.
- `evaluate_declaration` (`src/value/expression/mod.rs`): the S6a seam, one arm per variant and no `_` arm, with `wildcard_enum_match_arm` and `match_wildcard_for_single_variants` denied. It builds the `Evaluation` from the hook result and the env. `CheckedPackage::call` now evaluates through it. `From<EvalOutcome<T>> for FamilyOutcome<T>` is the pass-through.
- `FamilyOutcome` gets a `#[cfg(seam_probe)]` variant. The seam-probe list adds `evaluate_declaration` and `ValueFunctionFamily::evaluate`, the one lib `match` over `FamilyOutcome`.

## Spec
FR-090 (Behavior, Inputs, Dependencies, Status), TC-385 (step 1, Status), ADR-012 §5.1 S1, `spec/spec.md` and `spec/tests.md` now state the rule.

## Tests (TC-385)
- `s6a_family_kind_admits_no_relation_and_family_outcome_has_two_arms` (unit): exhaustive matches over `S6aFamilyKind` and `FamilyOutcome<Value>`, each with no `_` arm. Each S6a kind goes through the seam against an empty package and returns `Err(InternalFault)` with stage `S6a` and invariant `checked-identity-not-resolved-by-package`. A declared identity through the seam returns `Evaluated(Completed(3))`.
- `both_family_outcome_arms_reach_a_caller_through_the_s6a_seam` (it): `CheckedPackage::call` returns `FamilyEvaluated(Refused)` with code `cardinality_out_of_bound`/`above-maximum` for a 2-member binding declared `[0,1]`, and `Evaluated(Completed(1))` for a 1-member binding.

## Mutation proofs
- A `Relation` variant added to `S6aFamilyKind`, with a seam arm so the lib builds: the lib test build fails with E0004 at the test's `s6a_family_name` match.
- A third `FamilyOutcome` variant, with an arm in `ValueFunctionFamily::evaluate` so the lib builds: the lib test build fails with E0004 at the test's `outcome_arm` match.
- `cargo xtask seam-probe` run before the list update reported both new seams as unexpected-but-present. After the update it confirms 6 locations.

`make ci`: exit 0.


kreneskyp:
Review findings on 2f768e13 fixed in d3167121. origin/main is merged in: #359 (qsl-forms) and #360 (K4, merged in 7678b6a9).

- **M1:** FR-063's seam-probe list and AC-6 now name the S6a seam `evaluate_declaration`. The deleted stage-participation table and its `Relation` evaluation arm are gone from both.
- **M2:** `S6aFamilyKind::family` maps each S6a kind to its `FamilyKind` with an exhaustive match. A compile-time check rejects any mapping to `Relation`, and any two kinds that map to the same family.
  - Mutation: add a `Relation` variant, and write `Value | Relation => FamilyKind::Value` in `family()`, the seam and the test. The build fails with E0080 "two S6a family kinds map to one FamilyKind".
  - Mutation: map `Relation => FamilyKind::Relation`. The build fails with E0080 "an S6a family kind maps to FamilyKind::Relation".
- **M3:** the migration recipe now lists the S6a variant, the `family()` arm, the seam arm and the TC-385 arm. It says the evaluator is reached through `evaluate_declaration`.
- **L1:** `s6a_family_kinds!` declares the enum and `S6aFamilyKind::ALL` from one list, and TC-385 iterates `ALL`.
- **L2:** the seam doc now says `identity` and `env` are `Value`'s, and that the next migrating family reshapes them.
- **L3:** the xtask test is now `checked_in_locations_name_functions_that_hold_a_match`. It checks with syn that each location names a function that holds a `match`. Renaming a list entry makes it fail. `S6aFamilyKind::family` is added to the list, which now has 7 entries.
- **L4:** added an ADR-011 §6.2 row for `family`: layer 3, `check` core.

`make ci` on 7678b6a9: exit 0.

